PR #271, opened 2014-03-20
Try to detect DMX value changes and output a minimal changeset first, as...

leonerd:
... it will be quicker

I have a fairly cheap clone of the uDMX hardware, and it turns out that it's quite slow to output a full DMX universe of 512 channels. I can only get about 3 or 4 frames per second out of it. This is too slow for smooth fades.

I find that if I make this change then my fades become a lot smoother when there's only a few fixtures affected, as the output range is smaller, so it outputs quicker.

I doubt this code is particularly well implemented so far, but it serves my purpose, and hopefully someone will be able to comment on how it can be improved.


mcallegari:
More than one year an no activity.
Closing this now.

PR #415, opened 2014-12-16
function start and stop status

plugz:
A function will always switch to the states it has been given
A function will always end up in the last state is has been given

https://sourceforge.net/p/qlcplus/discussion/development/thread/d344d9a9/


plugz:
Will redo with another method.

PR #432, opened 2015-01-07
Add robe pointe

Moehritz:
Based on this specs: http://www.robe.cz/fileadmin/robe/downloads/dmx_charts/Robin_Pointe_DMX_charts.pdf


janosvitok:
Thanks, I will fix one omission in a moment.

janosvitok:
I've fixed some details. Would you please fill in gobos where possible (from the built-in gobos folder)?

PR #440, opened 2015-01-20
Adding new tutorial about Mixxx integration with QLCplus

markusb:
Tutorial, as discussed here: https://sourceforge.net/p/qlcplus/discussion/general/thread/be3a87e4/

Had a conflict when pushing as the changelog was modified elsewhere. I removed my change, merged, then added my change again. This may not be elegant, but it worked :-).

Markus


janosvitok:
Great! Thanks a lot!

mcallegari:
Too late comment: this doesn't sound like a good idea to me.
Most likely 1 user on 1000 will be interested in Mixxx and QLC+, but this change delivers this information also to the other 999 "not interested" users.

I'd prefer if this tutorial was placed in the GitHub wiki. I was happily willing to reference it from the QLC+ website with credits to Markus.

I believe what users want are basic tutorials (basically what I tried to do with the videos, but in HTML) to move the first steps with direct examples.
Very specific tutorials like this one risk to be a complain why we didn't write the basic ones first...

PR #442, opened 2015-01-22
VCFrames: allow to enable page scroll loop

plugz:
2015-01-22-183117_4480x1440_scrot

With this checkbox checked, going further that the last page will take back to the first page.
And going before the first page will take to the last page.


janosvitok:
Thanks! You were fast!

BTW. I'm not native English speaker, but another word that I've heard in this situation is "wraps" or "wraps around". I'll leave it up to you to judge :)

plugz:
You're right :)

What about "circular page scrolling" ?

Issue #435, opened 2015-01-16
Build issue on openSUSE tumbleweed: nanodmx.cpp


nettings:
Hi Massimo, hi everybody,

I'm trying to re-sync to the latest qlcplus git after updating my laptop, and I've been running into a number of build issues.
System is openSUSE tumbleweed of 2015-01-15, i.e. pretty recent.

g++ -c -pipe -Werror -Wno-unused-local-typedefs -g -Wall -W -D_REENTRANT -fPIC -DQT_WEBKIT -DLIBFTDI1 -DQT_PLUGIN -DQT_GUI_LIB -DQT_CORE_LIB -DQT_SHARED -I/usr/share/qt4/mkspecs/default -I. -I/usr/include/QtCore -I/usr/include/QtGui -I/usr/include -I../../interfaces -I../../midi/common -I/usr/include/libftdi1 -I/usr/include/libusb-1.0 -I. -o nanodmx.o nanodmx.cpp
nanodmx.cpp: In member function ‘QString NanoDMX::getDeviceName()’:
nanodmx.cpp:84:52: error: invalid use of incomplete type ‘libusb_device {aka struct libusb_device}’
         if (dir.startsWith(QString::number(m_device->bus->location)) &&
                                                    ^
In file included from qlcftdi.h:36:0,
                 from dmxusbwidget.h:24,
                 from nanodmx.h:25,
                 from nanodmx.cpp:20:
/usr/include/libftdi1/ftdi.h:345:12: error: forward declaration of ‘libusb_device {aka struct libusb_device}’
     struct libusb_device *dev;
            ^
Makefile:312: recipe for target 'nanodmx.o' failed

I'm not using a nanodmx, so I'm currently working around the issue by commenting out the corresponding lines in the Makefile.

Best,

Jörn

kripton:
Hello, thanks for the report. There are two API-incompatible version of libFTDI currently out there. You're using the 1.0-branch of libFTDI. I once fixed it for the plugins that existed at that time (#287) but was unable to fix build for the plugins introduced afterwards (tried several times though).

nettings:
kripton, thanks for the feedback. I'm confused, though. This is what I have installed (from openSUSE tumbleweed repository):
i | libftdi1       | package | 1.2-1.1 | x86_64 | repo-oss  name: libftdi1
i | libftdi1-devel | package | 1.2-1.1 | x86_64 | repo-oss   name: libftdi1-devel
[...]

PR #443, opened 2015-01-25
fix build errors on system uses libftdi1

axelvvd:
This is a patch to resolve the build issue for qlcplus on system wich use libftdi1 #435


mcallegari:
Merged, thanks

PR #448, opened 2015-02-16
updated and cleaned playbackwing

offtools:
*removed the code of the older patches
*pageup/pagedown buttons now works inside the frame widget pages (only)
*go/back buttons now work

### DIFF
--- a/plugins/enttecwing/src/enttecwing.cpp
+++ b/plugins/enttecwing/src/enttecwing.cpp
@@ -241,8 +241,6 @@ void EnttecWing::addDevice(Wing* device)
     connect(device, SIGNAL(valueChanged(quint32,uchar)),
             this, SLOT(slotValueChanged(quint32,uchar)));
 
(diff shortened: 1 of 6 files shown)


offtools:
Hi Massimo,

two things.

1. it seems to be something wrong, it should be only the last commit, not my upstream merge commit, strange. Anyway can you use it?

2. Syncing sliders with the wing on page changes is still an issue. Do you think its possible to use sendFeedback and check/solve syncs inside the plugin? Have no other idea at the moment.

mcallegari:
Merged. Seems all OK, thanks !

About feedbacks, they are already sent when a page is changed.
If I remember correctly though, a frame understands a page change only if the signal is a "next page" or "previous page" as defined in an Input profile.
I see there are no such definitions in the enttec playback/shortcut profiles. Maybe you can give it a try and see if feedback gets OK after that ?
In case, an update to the existing input profiles would be welcome :)

PR #450, opened 2015-02-20
Added initial support for StopActions

joepadmiraal:
I added the option to select a stop action for a scene in the SpeedDialWindow.
This allows me to switch between scenes/chases without the pan/tilt channels to reset.
This is very useful when controlling QLC+ live.
See the discussion on the forum:
https://sourceforge.net/p/qlcplus/discussion/general/thread/5717245b

Regards,
Joep Admiraal

### DIFF
--- a/engine/src/scene.cpp
+++ b/engine/src/scene.cpp
@@ -37,6 +37,10 @@
 
 #include "sceneuistate.h"
 
+const QString KResetToZeroString       (       "ResetToZero" );
+const QString KResetToStartString      (      "ResetToStart" );
+const QString KKeepCurrentValuesString ( "KeepCurrentValues" );
+
 /*****************************************************************************
  * Initialization
  *****************************************************************************/
@@ -322,6 +326,9 @@ bool Scene::saveXML(QDomDocument* doc, QDomElement* wksp_root)
     /* Speed */
     saveXMLSpeed(doc, &root);
 
+    /* Stop action */
+    saveXMLStopAction(doc, &root);
+
     /* Channel groups */
     if (m_channelGroups.count() > 0)
     {
@@ -411,6 +418,10 @@ bool Scene::loadXML(const QDomElement& root)
         else if (tag.tagName() == KXMLQLCFunctionSpeed)
         {
             loadXMLSpeed(tag);
+        }
+        else if (tag.tagName() == KXMLQLCSceneStopAction)
+        {
+            loadXMLStopAction(tag);
         }
         else if (tag.tagName() == KXMLQLCSceneChannelGroups)
         {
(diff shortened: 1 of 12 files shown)


mcallegari:
Hi Joep, thanks for investigating this.
However, I won't merge it as this is a workaround for the specific case you found and it's not a generic solution. Plus, it's very invasive.

I've investigated a similar behaviour in the past too, and I concluded that the real problem is that transactions between QLC+ functions don't happen in the same Mastertimer tick.

So basically what happens is:
- function running (several ticks)
- function stop: in this tick (if fade out is set to 0) QLC+ will commit to reset all the values to 0
- new function start: this happens the next tick after stop, but it's too late since values are considered to be 0.

This involves also HTP channels and I noticed it a while back on some sequences, where I had lights going black for 1 tick. Quite annoying.

Now many changes have been done on QLC+ since then and I just noticed that when a function is stopped, LTP channels are reset to 0 as well. In the original QLC code, they were staying at the last value. Only HTP channels were reset to 0.
I believe this is the main problem and I think we should fix that.

The fix must be done at a very low level. Any "high level" change, is my opinion, is a workaround and doesn't solve the real issue.

joepadmiraal:
Hi Massimo,

Thanks voor your detailed comment.
Fixing this issue low level would solve my problem with the pan/tilt functions. However I think it would also change the way the system currently works. I think a lot of people are now relying on this behaviour. Therefore I think we still need to provide an option to choose between resetting to 0 and keeping the last value. That being said, fixing the issue low level seems like a good plan. I can take a look at it but it will take some (a lot) of time. Since I almost have no free time at the moment :)

mcallegari:
A reminder for me:

[MasterTimer] *********** tick: 2022 **********
[MasterTimer] ticking DMX source 0
[MasterTimer] ticking fader (channels: 0 )
Function start(). Name: "New Chaser 3" ID:  3 , startTime: 0
Function stop(). Name: "New Chaser 2" ID:  2
[MasterTimer] *********** tick: 2023 **********
void ChaserRunner::postRun(MasterTimer*, QList<Universe*>)
Function stop(). Name: "middle" ID:  0
Function postRun. Name: "New Chaser 2" ID:  2
void Function::resetElapsed()
Function postRun. Name: "middle" ID:  0
void Function::resetElapsed()
Function preRun. Name: "New Chaser 3" ID:  3
Function start(). Name: "up" ID:  1 , startTime: 0
[MasterTimer] ticking DMX source 0
[MasterTimer] ticking fader (channels: 2 )
[MasterTimer] *********** tick: 2024 **********
Scene preRun. ID:  1
Function preRun. Name: "up" ID:  1
[MasterTimer] ticking DMX source 0
[MasterTimer] ticking fader (channels: 0 )

Joep, this is worse than I thought. The switch between two chasers happens in 3 ticks...not 2 :(
I don't know if the debug messages I've added caused that, but we definitely need the switch to happen in a single tick.

mcallegari:
Hey Joep, I think I have good news !
This change: https://github.com/mcallegari/qlcplus/commit/829bb9384399b685a890e10ec6c5b052eda71452
Seems to fix your zero test project as well.

Can you please confirm and eventually report if I broke anything else that I'm not aware of ATM ?

Thanks

joepadmiraal:
Hi Massimo,

Great news!
I'll check it out and do some tests today and tomorrow.

joepadmiraal:
I tested it and did not see any problems with the commit.

mcallegari:
Great, thanks for the feedback !

PR #455, opened 2015-03-02
Enttec Playbackwing using Feedback to sync widget values and playbackwing sliders 

offtools:
Finally implemented syncing between widgets in the virtual console and sliders of the playbackwing by using Feedback.

If wing sliders and widget values in the virtual console differ, in case of page changes or changing stuff with the mouse, the wing slider have to fetch the current values in the virtual console, before sending new input. This prevents jumping of the output values.

With this commit also pages/page changes are fully supported for the playbackwing, when using the frame widget. Page changes work in booth directions (qlcplus -> wing, wing -> qlcplus). 

I only had to make two little changes outside the plugin (vcframe.cpp, vcwidget.cpp). Hope this is ok, I added some comments there for review.

offtools

### DIFF
--- a/plugins/enttecwing/src/enttecwing.cpp
+++ b/plugins/enttecwing/src/enttecwing.cpp
@@ -53,7 +53,7 @@ QString EnttecWing::name()
 
 int EnttecWing::capabilities() const
 {
-    return QLCIOPlugin::Input;
+    return QLCIOPlugin::Output | QLCIOPlugin::Input | QLCIOPlugin::Feedback;
 }
 
 bool EnttecWing::reBindSocket()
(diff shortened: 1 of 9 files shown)


mcallegari:
Hi @offtools , please see my inline comments.
Also, the test unit for the wing fails. Can you please fix it as well ?

Thanks !

offtools:
thanks for the pointers, will fix it :)

offtools:
Hi @mcallegari 

I changed the things you demanded.
I also changed or removed the bcd call. I can send the page number without conversion to the wing and it is shown correct on the lcd panel. Maybe it is since an firmware update but I cant remember. The API still says, that the format for the wing is bcd. There are no further information about API changes on the enttec site. Anyway on my wing it is working, dont know if anyone else of the qlc+ users using the wing.

offtools

mcallegari:
Merged ! :+1: 
Thank you very much ! Gonna update the changelog soon

PR #458, opened 2015-03-10
Collection: pass down intensity

plugz:
As of now, the intensity of a collection is not passed down to it's member functions.


plugz:
I don't know which is the desired behavior.

mcallegari:
It seems the expected behaviour to me but it's incomplete.
This is OK to set the startup intensity of the children functions, but the interesting case is when a Collection is a Chaser step, with fade in and fade out.
I guess that's what users are asking since quite a while now

plugz:
Alright, I'll complete this later :)

PR #459, opened 2015-03-12
Build Travis with Qt4 and Qt5

peternewman:


mcallegari:
Wonderful ! Thank you very much @peternewman !

mcallegari:
Hey @peternewman , here's a curious thing.
There is one important QLC+ class that tests a timer. It relies on the fact that the test is running at a "production" speed, so in normal conditions.
The test fails on a virtual machine (I've seen it before)
And here it is in #461 where the qt5 target goes fine and the qt4 doesn't.
Probably the qt4 build happened in a moment of processing spike of the server, so the test didn't run at the correct speed and failed.
Now, I can soften the test, but when a patch would cause the current test to fail, we won't spot it in the future.

I am wondering if there is even a workaround to this...

peternewman:
I'd personally, and for OLA, we loosen the tests so they pass for everyone. OLA already gets run on VMs for some people. If QLC+ doesn't work on VMs, then I guess that's less relevant.

You should have the ability to rerun the build here, which would confirm it:
https://travis-ci.org/mcallegari/qlcplus/jobs/54237553

Personally, I'd be inclined to add a build time config option for VMs that allows you to loosen the timing so the tests pass. You won't spot the issue on Travis, as it won't fail there, but that's probably more useful than continually false positives for people opening PRs. It can still be caught by people running on real hardware when they do, without disrupting the contribution process too much. Buildbot would obviously help this, although you'd still only find out after the code was merged.

mcallegari:
Agree. Is there any environment variable that I can count on to loosen (or exclude at this point) the test ?

Maybe we can add BUILDSYS='travis' to the 'env' statement ?

mcallegari:
Please continue the discussion on #463 
(found it awkward to discuss on closed PRs :smile: )
Thanks

PR #460, opened 2015-03-12
Function start stop 6

plugz:
Kinda like https://github.com/mcallegari/qlcplus/pull/415

This needs Documentation, both code comments and some diagrams to help understand what's happening in the engine :)

The testfile add the end of https://sourceforge.net/p/qlcplus/discussion/development/thread/d344d9a9/ seems to work ok with this change.


plugz:
PS  - Unit Tests will not pass with this PR, they need to be changed too.

PR #461, opened 2015-03-13
Eurolite USB DMX512 Pro plugin

karrika:
The plugin has been tested on Ubuntu 14.04 amd64 only. Reading the serial number with ENTTEC protocol did not work so I omitted that part.

Cheers,
Karri Kaksonen

### DIFF
--- a/plugins/dmxusb/src/dmxusb.cpp
+++ b/plugins/dmxusb/src/dmxusb.cpp
@@ -26,6 +26,7 @@
 #include "dmxusbwidget.h"
 #include "enttecdmxusbpro.h"
 #include "enttecdmxusbopen.h"
+#include "euroliteusbdmxpro.h"
 #include "dmxusb.h"
 
 /****************************************************************************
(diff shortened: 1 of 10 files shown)


mcallegari:
Hi Karri, thanks.
Please see my inline comments. Easy fixes. Can you please do that and commit ?

Does this device support DMX input ?

Maybe if you subclassed EnttecDMXUSBPro and just rewrite the detection part, you could have had all the functionalities of the Enttec Pro adapters out of the box. (including input)
Probably 50 lines of code for everything.
Worth a try ?

karrika:
I actually tried the input but did not get it to work. When time permits 
I will try it again.

--
Karri

karrika:
The Eurolite has only one output connector. DMX input is not possible.

I was only trying to read the serial number.

--
Karri

mcallegari:
Hi Karri, one last thing. The new added type must be handled also here
https://github.com/mcallegari/qlcplus/blob/master/plugins/dmxusb/src/qlcftdi-libftdi.cpp#L186
and here
https://github.com/mcallegari/qlcplus/blob/master/plugins/dmxusb/src/qlcftdi-ftd2xx.cpp#L201

One last effort and then I'll merge your PR 
Thanks ! :+1:

mcallegari:
Thing is that the Eurolite adapter doesn't use a FTDI chip !
So using the pre-compiled FTDI library in Windows wouldn't detect it anyway.

The FTDI library in Linux, instead, can be used a generic USB->serial interface even for non-FTDI chips

mcallegari:
Oh, I merged the PR.
Thanks !

karrika:
Thanks. I just realized that the file src.pro in the dmxusb/src directory did not exclude eurolite from windows compilations...

PR #464, opened 2015-03-14
Add Eurolite-LED-Flood-Light-252-RGB-19

karrika:
I could not find this fixture so I created it with the fixture creator.


kripton:
Hi, thanks for your submission. This fixture looks a lot (channel mapping wise) like the existing "Eurolite-LED-PAR64-RGB.qxf". Massimo, do we have a policy about fixtures from the same manufacturer, different model name but same behavior? Can we define aliases?

mcallegari:
@kripton  unfortunately we don't.
I think there are already some kind of "duplicates" in the library.
The alias idea is not bad at all actually

mcallegari:
Thanks @karrika :+1:

PR #466, opened 2015-03-22
Use the globally defined TRAVIS=true env var rather than setting our own

peternewman:


mcallegari:
Ah, nice, thank you !

RickBrock:
Guys,
I really do think you are fantastic folks.

I've been following these threads for a while and have not got a clue what
they mean.... But I can tell there is a LOT of work goes into all this.

Thanks.
Rick

Issue #132, opened 2013-04-13
Build issue with Qt


jeromelebleu:
I can't build latest release with current Qt4 on ArchLinux (I didn't tried on other distributions) because of all warnings being treated as errors.

```
In file included from /usr/include/qt4/QtCore/QtCore:53:0,
                 from /usr/include/qt4/QtXml/QtXml:3,
                 from qlccapability.cpp:26:
/usr/include/qt4/QtCore/qtconcurrentfilter.h: In function 'QtConcurrent::ThreadEngineStarter<void> QtConcurrent::filterInternal(Sequence&, KeepFunctor, ReduceFunctor)':
/usr/include/qt4/QtCore/qtconcurrentfilter.h:108:47: error: typedef 'Iterator' locally defined but not used [-Werror=unused-local-typedefs]
     typedef typename Sequence::const_iterator Iterator;
                                               ^
cc1plus: all warnings being treated as errors
make[2]: *** [qlccapability.o] Error 1
make[2]: Leaving directory `/home/light/Documents/Logiciels/QLC+/package/src/qlcplus-4.3.2/engine/src'
make[1]: *** [sub-src-make_default-ordered] Error 2
make[1]: Leaving directory `/home/light/Documents/Logiciels/QLC+/package/src/qlcplus-4.3.2/engine'
make: *** [sub-engine-make_default-ordered] Error 2
```

It's the same issue than this one https://github.com/Mezomish/juffed/issues/52, fixed by removing **-Werror** from compilation flags : https://github.com/Mezomish/juffed/commit/054bf0c0adf19a0bf5be58cf54049379f50b9211
I tried to remove it from `variables.pri` and it fixes this issue too.

PR #467, opened 2015-03-23
InputSources: wrap in QSharedPointer to avoid double frees.

plugz:
(PR because quite big number of changed lines)

This fixes at least 2 crashes:
- A VCMatrix containing a Control connected QLCInputSource
    -> delete the Control, its QLCInputSource gets deleted
    -> delete the VCMatrix -> double free of the QLCInputSource

- Any widget in the virtual console connected to a QLCInputSource
    -> Go to the widget's properties, change the input
        (this deletes the current InputSource)
    -> Cancel the properties window
    -> the widget's InputSource is now pointing to freed memory

A simple QSharedPointer wrapping will avoid double frees and leaks.

### DIFF
--- a/engine/src/channelsgroup.cpp
+++ b/engine/src/channelsgroup.cpp
@@ -39,7 +39,6 @@ ChannelsGroup::ChannelsGroup(Doc* doc)
     : QObject(doc)
     , m_id(ChannelsGroup::invalidId())
     , m_masterValue(0)
-    , m_input(NULL)
 {
     setName(tr("New Group"));
     m_doc = doc;
@@ -60,9 +59,6 @@ ChannelsGroup::ChannelsGroup(Doc* doc, const ChannelsGroup* chg)
 ChannelsGroup::~ChannelsGroup()
 {
     m_channels.clear();
-    if (m_input != NULL && m_input->isValid() == true)
-        disconnect(m_doc->inputOutputMap(), SIGNAL(inputValueChanged(quint32,quint32,uchar)),
-                   this, SLOT(slotInputValueChanged(quint32,quint32,uchar)));
 }
 
 /****************************************************************************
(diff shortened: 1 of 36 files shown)


plugz:
About the -Wno-unused-typedefs

With gcc-4.8, I have trouble compiling with the flag -Wno-unused-local-typedefs
But no issue whatsoever without it...

mcallegari:
OMG I want to cry :cry: 

Can't we just improve the qDeleteAll(m_inputs); call in vcwidget.cpp ?
In any case this change doesn't pass the CI, so please fix first.
As for GCC 4.8, I normally use it on Mint 17.1 and got no issues. What distro do you use ?

plugz:
qt4 build failed... Do we still need to support qt4 ? If so I'll fix this.
Also, which version of Qt4 is used on the travis builds ?


> Can't we just improve the qDeleteAll(m_inputs); call in vcwidget.cpp ?
No, it's more complex than this. There is no "owner" to a QLCInputSource, it's sometimes deleted in the widget using the inputsource, sometimes in the widgets' properties. Sometimes never...
This PR changes a lot of LOCs, but it's goal is very simple: wrap QLCInputSources in a shared pointer, so it can't be ninja-deleted.


As for gcc: I was wrong, there is no issue with gcc 4.8.3 (fedora 20), but removing the line does not cause any error to appear. On another computer with gcc 4.6.x, I have to remove the line to allow qlc+ to compile.
My question: which configuration makes this line mandatory ?

mcallegari:
> qt4 build failed... Do we still need to support qt4 ? If so I'll fix this.
Yes we do. At least until QLC+ 4 exists. Keep in mind that Linux DEBs are still qt4-based. Windows and OSX are already on Qt 5

> Also, which version of Qt4 is used on the travis builds ?
Expand line 169 of the log. It's Qt 4.8.1

>  wrap QLCInputSources in a shared pointer, so it can't be ninja-deleted.
Makes sense. I just saw a lot of changes and got scared :fearful: 

> My question: which configuration makes this line mandatory ?
I added it for gcc 4.8. Discussed in #132

plugz:
Alright, I'm working on the qt4 build right now :)

mcallegari:
:+1: I think it's just a missing include <QSharedPointer> in inputoutputmap.h

plugz:
Should be good now.

> I added it for gcc 4.8. Discussed in #132

OK thanks.
Well I just tried building with the line commented, gcc 4.8.3, qt 4.8.6, and the build suceeded. Maybe the issue has been fixed in a more recent qt4 version ?
I say this because removing the line could allow building with clang on linux :)

mcallegari:
> I say this because removing the line could allow building with clang on linux :)

Well, the compiler can always be detected in a qmake enviroment. I think you should use $QMAKE_CXX:
http://doc.qt.digia.com/4.6/qmake-variable-reference.html#qmake-cxx

plugz:
I haven't found a configuration where the line is mandatory. Just sayin', maybe it's not useful any more.

Anyway, this line has nothing to do with the PR :)

Travis looks happy now, so if you have no other remark, I'll merge the change.

mcallegari:
I think the gcc flag need depends on the Qt version. Let's leave it there for now.

No remarks, please proceed. Yours the responsibility though if something explodes :sweat_smile:

plugz:
> Yours the responsibility though if something explodes :sweat_smile:

Sure thing :)

I'm trying to achieve the exact opposite here :p

PR #468, opened 2015-03-31
Fixing elided fixture names to ... in sceneeditor

offtools:
when having a huge amount of fixtures in a scenes the fixture names are elided to (...), which makes it hard to find a certain fixture tab in the scene editor. 
Setting the ElideMode to Qt::ElideNone forces the tabs to write the full fixture names.


mcallegari:
Hi, thanks but this change will make a bunch of other users not happy.
The elide was introduced, in fact, to see the beginning and the end of a fixture name, when there are lot of fixtures like in your case.
Basically the beginning helps to recognize the make/model, while the end to recognize the progressive number of a fixture batch.

I'm sorry but I'm not gonna merge this. I guess you need to live with a custom build of QLC+.

P.S. The change can be done in SceneEditor.ui --> m_tab (QTabWidget)

offtools:
Hi Massimo,

the problem is whith longer names (e.g. default naming scheme of Dimmer #XY) and around 40 Fixtures in a scene elide doesnt work like expected, it starts to cut also start and end, only dots are shown.

mcallegari:
Well, then I think there is no proper solution for this, unless we implement multi-row tabs.

I evaluated it a long time ago but it is not a standard feature of Qt. Probably the QTabWidget item should be subclassed and properly customized.
Kind of a hard work to do.

See this topic: https://forum.qt.io/topic/29899/how-to-create-multiple-rows-tab-with-qtabwidget-in-qt

I'll keep this in mind on QLC+ 5 :wink:

PR #473, opened 2015-04-09
Function startstop 08 f

plugz:
This fixes the issue "scene in a very quick chaser may stay ON after the end of the chaser"

Also, if several "sources" start a function (a chaser, a collection, whatever), all the sources have to stop the function for the function to actually stop.

This allows having a function in several collections at the same time.
If several collections contain the function, they will have to all be stopped for the function to stop.
-> The function will not be stopped when stopping only one collection.


plugz:
See #478

PR #474, opened 2015-04-09
fix build on fedora 21

plugz:
add missing ifdef in euroliteusbdmxpro.cpp

reorder package detection in dmxusb plugin

Will merge right after travis' green light.


mcallegari:
@plugz in Travis we trust :smile:

PR #480, opened 2015-05-05
Function startstop 09

plugz:
This fixes the issue "scene in a very quick chaser may stay ON after the end of the chaser"

### DIFF
--- a/engine/src/chaserrunner.cpp
+++ b/engine/src/chaserrunner.cpp
@@ -78,7 +78,6 @@ ChaserRunner::~ChaserRunner()
 {
     clearRunningList();
     delete m_roundTime;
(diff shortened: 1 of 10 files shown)


mcallegari:
@plugz  I like this one ! :+1: 
Let's see what Travis old fart says :construction_worker: 

Sooner or later we should add a test unit for Sequences, Right now they are not covered and every change to Scene/Chaser/Function/MasterTimer can affect them.
I will think of something that makes sense, unless you're faster than me :smiley:

plugz:
:)

mcallegari:
Should we merge this then ?

plugz:
Yes

PR #481, opened 2015-05-15
Add Nodle U1 support for HID plugin

StefanKrupop:
- Added VID/PID of DMXControl Projects e.V. Nodle U1
- Fixed wrong HID device beeing returned for output when HID device with input only is available


mcallegari:
Thanks for this.
From the pure code organization, placing "DMXControl Projects Nodle U1" into a class dedicated to the "Digital Enlightenment FX5" doesn't make much sense.
In fact the VID/PID defines start with "FX5_"

Basically if they share most of the code, I would rather prefer to have a base class called HIDDMXDevice and then subclass HIDFX5Device and HIDNodleU1Device. Mostly like what we did in the DMXUSB plugin.

I can do that, but I have none of the HID DMX devices to test if everything works ok.
Do you have also a FX5 ?

StefanKrupop:
I personally do not have an FX-5, but the "original" Digital Enlightenment interface. The only differences are the VID/PID and the use of SMD components.
But we as the "DMXControl Projects e.V." club have some FX5s available so that we can organize a test.

janosvitok:
Massimo, I'd say that if onle needed change is to add VID/PID, there's no need to create new class.

As I understand it, the rest of the changes are fixes for all HID devices:

1. fix out-of-range condition: when the offset is outside of DMX range (offset  more than 512-packet size)
2. indexing output devices should skip devices that don't have output. I didn't check how the selection in GUI works, but:
 - I'd say that the indexes should be related to the devices regardless of their capability (i.e. input and output of a device should have the same index). I'm not 100% sure about this design goal, it just looks natural.
- since the code works for Stefan, I suspect that it is not the case now. So Stefan's fix improves the situation.
- the ultimate fix is to rewrite selection to fulfill the above design goal (unless we decide it's a wrong goal). I'm afraid there's nobody to write it now...

I have no more time now to investigate more, if I find some I'll comment more.


Stefan, thanks for the improvements!

mcallegari:
@janosvitok there's no problem for me to merge this as it is, moreover cause I don't have the devices to test anything.
My point is that I've already done 2 major rewrites of the dmxusb plugin code, because the original structure didn't handle multiple outputs/input (nor midi)
So basically, what happens if tomorrow a Nodle U2 comes out with 2 DMX outputs ?
In general I like to have generic and flexible structures, to avoid major rewrites in the future.

But again, for now this is OK, so let's merge it

mcallegari:
OK, I couldn't resist to rename a few things :stuck_out_tongue_closed_eyes: 
See this a55f0b9e1e84a99843771f98d5bb8b1f55fe3cb6

UI strings were just wrong anyway. They were all referring to the FX5. I made them more generic.
@StefanKrupop please check if everything is still OK after my changes.
If you can, a test on the FX5 would be much appreciated too !

Thanks again !

StefanKrupop:
Hi!

I checked with all three types of interfaces that are (now) supported and could not find any issues. I tested DMX-Out and DMX-In, each with some channels at the beginning and at the end of the universe.

mcallegari:
That's great news !
Thank you very very much !

PR #485, opened 2015-05-17
Euro lite ledslsqcl

kripton:
Add definition file Eurolite-LED-SLS-X-QCL.qxf (via manual) that matches the SLS-5 QCL and SLS-12 QCL. Physical data taken from SLS-12 QCL.


mcallegari:
Thanks @kripton :+1:

PR #487, opened 2015-05-24
Artnet: setting for output mode (Full 512 channels or minimal)

janosvitok:
Proof of concept. I didn't get to test it with real device.  Please comment. I'll do docs etc. later.

If the idea is OK, it may be worth doing the same for usbdmx and e131...


mcallegari:
Ouch, please standby.
At the moment I am travelling through hell to implement plugin parameters.
I will include this change as well.

mcallegari:
Closing. I've integrated this into the new code.

PR #494, opened 2015-06-06
Update catalan translation ArtNet, E1.31 and OSC plugins

sbenejam:
Update catalan translation ArtNet, E1.31 and OSC plugins


mcallegari:
Thanks Santiago :+1:

Issue #457, opened 2015-03-08
(Channel Modifier Dialog) ChannelsSelection::getSameChannels returns empty list for generic Dimmers


offtools:
Hi,

In the Channel Modifier Dialog  "Apply changes to fixtures of same type" does not work for generic Dimmers.

ChannelsSelection::getSameChannels::210

Mainly caused because generic Dimmers do not have a fixture Definition. The recently added Generic RGB Fixtures have a fixture definition, so it works there.

Workaround for that is to create an own Fixture Definition for Simple Dimmers  with the Fixture Editor.

mcallegari:
You got the point. 
Dimmers are a special case in QLC+ and we definitely need to get rid of that.
It's in my TODO list, but I have no idea when I will be able to treat this thing.

The idea is to add a special tag in the XML fixture definition to allow an arbitrary number of channels.
At the moment that's the only thing (AFAIK) that prevents a dimmer to have a fixture definition

janosvitok:
@offtools: Do you want to apply the change in this case to ALL channels of ALL dimmer fixtures or just to the Nth channel of every dimmer fixture? For example: 3 dimmer packs of 12 channels each; if you invert channel 5 of the first dimmer pack, do you want to also invert all channels in all dimmer packs (3x12 channels total) or just channel number 5 in each dimmer pack (3x1 channels total).

mcallegari:
@offtools ping ?

mcallegari:
After 2999f40690dc195c5c581b49917062522c33aa17 the functionality now works as it should.
I'm closing this, cause we haven't had any comment in 4 months

PR #496, opened 2015-06-11
Simple fix for "Apply changes to fixtures of same type" & dimmers. Fi…

janosvitok:
…xes #457.


mcallegari:
Uhmm...I'm not sure if this is what theater guys want.
If I read the code right, you are selecting every channel of every dimmer fixture.

I believe what they want is: I have 10 fixtures with 6 channels each. If I select channel 2, every channel 2 of the other 9 fixtures should be selected.

mcallegari:
@janosvitok  After 2999f40690dc195c5c581b49917062522c33aa17 can we abandon this ?

janosvitok:
Let's forget this :)

PR #497, opened 2015-06-15
Switch to gcc-5 on Travis

peternewman:


peternewman:
Can you merge this if you're happy @mcallegari ? My next change is to get it building OLA, from our PPA once 0.9.7 is ought, which may help to catch issues like https://github.com/OpenLightingProject/ola/issues/826 in future.

mcallegari:
hi @peternewman , sorry for the big delay.
I thought about merging this PR a few times, but I believe it's too soon.
If the purpose of Travis is to spot build errors, then we should build as much as possible with currently used versions of GCC.
Now I have no idea which Linux distros have already switched to GCC 5, but I can tell you that:
- on Windows I'm using GCC 4.9.2
- on Linux GCC 4.8 (maybe 4.6 on ubuntu 12.04 32bit, but I need to double check)
- on OSX clang
- on the raspberry pi GCC 4.8.3

So for now I would stay on GCC 4.8, unless we refine this patch to build another target like "qt5-gcc5"

peternewman:
Hi @mcallegari no worries about the delay.

Yes, I'd agree the purpose of Travis is to spot build errors, and the best way to spot those errors is to run the bleeding edge versions, as then you'll catch any errors people may experience. Firstly because some people are running the latest versions ( e.g. https://github.com/OpenLightingProject/ola/pull/765 ) and secondly because the latest versions of the compiler have the latest and greatest/tightest error detection in, so will catch borderline issues, which makes the code more stable overall ( the same example, but still valid https://github.com/OpenLightingProject/ola/pull/765 ).

12.04's standard is indeed still 4.6.

So in summary I'd say we should run Travis at the bleeding edge, so it catches the issues before your users do. If you're really worried about compatibility, a quick Google only found one bug where running the latest version only wouldn't spot it ( https://bugs.gentoo.org/show_bug.cgi?id=476910 ), which was due to adding a feature only in the later version, then the thing to do would be to decide the minimum version you want to support and run that and the latest, e.g. 4.6 and 5, to cover the whole range.

mcallegari:
OK, let's go for this. Thanks

peternewman:
Great, thanks @mcallegari

PR #498, opened 2015-06-19
make use of hotplugmonitor's informations

plugz:
This allows connecting USB devices while QLC+ is running, so no need to kill QLC+ and restart it when you forgot connecting your controller before running QLC+ :)

This also allows to disconnect/reconnect a DMXUSB output device while running: no need to panic and run to the I/O tab to uncheck and recheck the output box.

Affected plugins:
* dmxusb
* hid
* midi
* peperoni

No tested:
* I have not tested this on windows
* I have not tested this on OSX
* I have not tested this with a peperoni device


Tolmino:
Dear plugz, in peperonidevice.h ispeperonidevice is defined two times.
Do it works again in that way?

plugz:
Dear Tolmino, yes. I think this http://www.cplusplus.com/doc/tutorial/functions2/ explains why it works.

mcallegari:
Of course it works ! It's one of the magic tricks of C++ :smiley: 
This is another example:
https://github.com/mcallegari/qlcplus/blob/master/ui/src/addfixture.h#L123
findAddress is defined 3 times with different arguments. At build time, the compiler treats them like different functions depending on their arguments

mcallegari:
@plugz just tested on Linux. Awesome stuff :+1: 
I noticed though that if you go to the IO tab, the plugins list of InputOutputPatchEditor is not refreshed on plugged/unplugged. You need to select another universe and then back.
The universe item instead is correctly updated.
Would be nice to have that feature too.
Gonna test on Windows and OSX now

mcallegari:
@plugz I just finished some preliminary tests on Windows and OSX.
I'm afraid I don't have good news.

On Windows it doesn't work at all. I don't see the HotPlugMonitor emitDeviceAdded/emitDeviceRemoved debug messages. It must be the platform specific implementation code to be fixed/finished.

On OSX the HPM signals are emitted but it doesn't work as expected.
- with a HID device (so input only) I got the unplug working but when I plugged back the device doesn't receive any input anymore. Plus I experienced a few crashes of the application. Probably an issue in the HIDDevice destructor specific to the OSX implementation.
The last messages before segfault are:
`void HIDPlugin::slotDeviceRemoved(uint, uint) "c12" "5"`
`virtual HIDDevice::~HIDDevice()`
`QThread: Destroyed while thread is still running`

- with a DMX USB (enttec open output only) unplug works as well, but not re-plug.

In general, I think this won't make it into 4.9.0. It's a very nice feature to have, but it needs a lot of cross-platform tests and I suppose I am the only one who can make them.
In any case, let's continue to work on this PR, until it reaches a good state.

mcallegari:
Ah, now I see the problem.
Apparently you removed the input thread stop from HIDDevice destructor:
https://github.com/plugz/qlcplus/commit/5ab8da3e1db3e0c0c105d051cbbd5d8fec416152#diff-35a0041da705a6bd1dd016934b725dd7L36

This produces a crash on OSX. I'm wondering why Linux requires this change ?

plugz:
@mcallegari 

IO Patch Editor not updated :
Yes, this would be better :)

Windows not working :
Nice, should not be hard to fix. I'll try a windows build later to fix this.

OSX issues:

> Apparently you removed the input thread stop from HIDDevice destructor:
> plugz@5ab8da3#diff-35a0041da705a6bd1dd016934b725dd7L36

> This produces a crash on OSX. I'm wondering why Linux requires this change ?

Without this change, the destruction of an HIDDevice instance causes a crash on linux ("pure virtual method called").


> [EDIT] restoring the thread stop in the destructor makes HID hotplug working on OSX
OK
What is your HID device detected as ?
hiddmxdevice, hidjsdevice, hidjoystick ?



> In general, I think this won't make it into 4.9.0. It's a very nice feature to have, but it needs a lot of cross-platform tests and I suppose I am the only one who can make them.

Yes, a whole lot of OSes and USB devices to test...

mcallegari:
> What is your HID device detected as ?
> hiddmxdevice, hidjsdevice, hidjoystick ?
On OSX I tested a HID gamepad, so hidosxjoystick

mcallegari:
I'm picking this back up.
I have cherry picked the code I needed from this PR and updated the outdated one in DMX USB.
Plus, I've fixed the IO manager instant update and kept the error popup in case a mapped device cannot be opened.
I have an idea how to fix it on Windows. Gonna commit and test again on OSX and then fix Windows.
Linux works like a charm :wink:

mcallegari:
the code should be completed now.
I've opened #628 to monitor this change.
Closing this then

PR #500, opened 2015-07-12
Two proposed changes to OLA plugin

rawbengal:
I'm not really sure how to do the pull request thing but I thought I'd risk looking like a fool...  Probably should have submitted these as two pull requests but couldn't figure out how to do that.

Change 1:  fix a bug in which OLA write buffer can have uninitialized data
Change 2:  correct the number of the first OLA universe to 0

### DIFF
--- a/plugins/ola/olaio.cpp
+++ b/plugins/ola/olaio.cpp
@@ -54,7 +54,8 @@ void OlaIO::init()
     m_thread = NULL;
     ola::InitLogging(ola::OLA_LOG_WARN, new ola::QLCLogDestination());
     // TODO: load this from a savefile at some point
-    for (unsigned int i = 1; i <= UNIVERSE_COUNT; ++i)
+    // For now, start counting at OLA universe #0
+    for (unsigned int i = 0; i < UNIVERSE_COUNT; ++i)
         m_outputs.append(i);
 
     bool es = false;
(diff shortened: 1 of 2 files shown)


mcallegari:
Hi, patches are submitted correctly.

BUT:
Change 2 breaks the backward compatibility, so users currently using OLA will have to change their projects
Change 1 can be shortened a lot. For example:

m_data.universe = universe;
memset(m_data.data, 0, K_UNIVERSE_SIZE);
memcpy(m_data.data, data.data(), data.size());
if (m_pipe)
        m_pipe->Send((uint8_t*) &m_data, sizeof(m_data));

rawbengal:
Thanks for the feedback.

Change two breaking backward compatibility:  I had assumed that usage of this plugin was low given that the bug fixed in change one would have been reported earlier.  Either way, I will post a new proposed fix that corrects the user display side and introduces an (optional) preprocessor define which specifies the first OLA universe to use.  The default will be the current behavior of OLA Universe 1.  The longer term fix is to allow user configuration of this value.  If I get some free time, maybe I will do that as well.

Change one:  Yes, you can fill the buffer twice (once with zeros and then again with the data) but the trade off will be processor efficiency when the supplied data is many channels.  In other words, the discussion isn't really about code correctness but style (fewer lines of code vs fewer processor cycles).  "m_data" is not used anywhere else in the code, so there is no reason to fill it if "m_pipe" is NULL.  In fact, technically, it doesn't even need to be a class member... it could just be static to the function.  Let me know how you'd like this change addressed and I will submit it again.

rawbengal:
I'm going to close this and see if I can figure out how to submit two separate pull requests. So they can be reviewed and accepted as independent thoughts.

PR #501, opened 2015-07-13
RGB Matrix: fix Waves script "In" direction behavior.

nedmech:
- Separated logic for determining pixel step index and pixel fill state.
- Added missing logic for pixel step index on "In" past center.


mcallegari:
Thanks for the update !

PR #503, opened 2015-07-17
Allow compile-time override of OLA plugin starting universe

rawbengal:
Same as last pull request for changing the starting OLA plugin universe, but now leave it as the same previous behavior (universe 1 is start).  Allow for a compile-time override using QLC_OLA_FIRST_UNIVERSE preprocessor symbol.  TBD is to allow for a user config for this value.


rawbengal:
Withdrawing this... The commit didn't turn out clean the way I wanted.

PR #504, opened 2015-07-17
added compile option to change OLA plugin starting universe

rawbengal:
Added a compiler preprocessing macro, QLC_OLA_FIRST_UNIVERSE, which allows the OLA plugin to be built with a different starting universe number.  Also, updates code for user display of universe number.  This commit does not represent any user-facing changes unless the module is compiled with a different starting universe number.


mcallegari:
Merged, thanks

PR #505, opened 2015-07-17
OLA:  Initialize output buffer with zeros

rawbengal:
Correct a bug in which output shorter than a full 512 channels sends uninitialized data to the OLA server.  Now fills buffer with zero bytes before copy.  Uses Massimo's suggested version of the change.

### DIFF
--- a/plugins/ola/olaoutthread.cpp
+++ b/plugins/ola/olaoutthread.cpp
@@ -104,10 +104,16 @@ void OlaOutThread::run()
  */
 int OlaOutThread::write_dmx(unsigned int universe, const QByteArray& data)
 {
-    m_data.universe = universe;
-    memcpy(m_data.data, data.data(), data.size());
     if (m_pipe)
+    {
+        Q_ASSERT(data.size() <= (int)sizeof(m_data.data));
+
+        m_data.universe = universe;
+        memset(m_data.data, 0, sizeof(m_data.data));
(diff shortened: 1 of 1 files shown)


mcallegari:
OK, sorry, ignore my previous comment.
I always forget that Q_ASSERT behaves on an inverse logic !
Wondering why the Qt guys wrote it like that...

mcallegari:
Merged, thanks :+1:

janosvitok:
Note to why is Q_ASSERT implemented that way: Q_ASSERT(_condition_) works the same way as standard assert() from <cassert> - it complains if the condition is false. Read "Q_ASSERT(_condition_)" or "assert(_condition_)" as "assert/assume/verify here, that _condition_ holds true". If _condition_ is not valid here i.e. the assertion is not valid, report to the user (dialog box/log message/breakpoint).

rawbengal:
This is how I think about asserts... They are there to explicitly specify what your program expects in order to operate properly.  Asserts are meant to say "this thing must be true otherwise something is HORRIBLY wrong with the logic of my program."  Their purpose is partially to test that your program works as intended but also to document your code.  They should not be used for normal error checking within your program.  They can be compiled out of your code with preprocessor macro QT_NO_DEBUG.

Thanks for merging my commits!   I might have more interesting changes to OLA in the near future.

PR #510, opened 2015-07-31
New modes for the EFX editor

grebecchi:
I have implemented an improved version of EFX generator to support more parameters other than classic Pan & Tilt.

This version has the following improvements:
- support RGB, Dimmer and Pan & Tilt for each fixture; 
- support parameters mix inside the same EFX;
- has a better EFX preview with RGB feedback and better preview on efx test;

### DIFF
--- a/engine/src/efx.cpp
+++ b/engine/src/efx.cpp
@@ -621,18 +621,22 @@ bool EFX::addFixture(EFXFixture* ef)
 {
     Q_ASSERT(ef != NULL);
 
-    /* Search for an existing fixture with the same ID to prevent multiple
-       entries of the same fixture. */
-    QListIterator <EFXFixture*> it(m_fixtures);
-    while (it.hasNext() == true)
+    /* Search for an existing fixture with the same ID and append at last but do
+     * not prevent multiple entries because a fixture can have multiple efx. */
+    //! @todo Prevent multiple entries using head & mode
+    int i;
+    for(i = 0; i < m_fixtures.size (); i++)
     {
-        /* Found the same fixture. Don't add the new one. */
-        if (it.next()->head() == ef->head())
-            return false;
+        if (m_fixtures[i]->head() == ef->head())
+        {
+            m_fixtures.insert(i, ef);
+            break;
+        }
     }
 
-    /* Put the EFXFixture object into our list */
-    m_fixtures.append(ef);
+    /* If not inserted, put the EFXFixture object into our list */
+    if(i >= m_fixtures.size())
+        m_fixtures.append(ef);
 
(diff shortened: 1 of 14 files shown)


mcallegari:
Hi, thanks for this, I've started to add some comments.

Not a big deal, but the title of this PR is kinda wrong. This doesn't add new shapes, but instead it should be "New modes for the EFX editor"

Since you copied my RGB Click&Go code into the engine, it would be nice to have a single place for that code, even though my color picker has also some preset boxes. I might take care of that once this PR is merged.

I would rename the "Gradient" class into some more meaningful name like "RGBColorPicker". In the end it provided 256 gradients :wink: 

I haven't tested the code yet, but it looks good to me. Well done !
@janosvitok is the maximum expert of the EFX area, so maybe he can review your code as well ?

Last thing. Since this makes the EFX editor no more related to pan&tilt only, we'll need to review the documentation accordingly

grebecchi:
Hi Massimo, I have fixed all suggestion of your comments.
I think the Gradient class as a generic gradient helper. For now it implement only RGB gradient but in the future it can implement also CMYK and white-black gradient (usefull for dimmer feedback).

mcallegari:
Thanks, I'll try it and eventually merge it as soon as possible

grebecchi:
I have added two missing modes in my previous implementation: the animation of single fixture channels (on X and Y axis).

channelx-channely

mcallegari:
I'm not too convinced about this last change, as the whole thing is getting confusing.
First of all, the names "ChannelX" and "ChannelY" are misleading. It's probably better to call them "X-Axis" and "Y-Axis" even though they're still confusing because they might refer to the XY axis of the fixture and not the XY axis of the EFX.

Also, when no channel can be selected, you should not even display a combo box, because from the screenshot above it seems that the RGB mode controls a PAN channel.

In general I think you should revert this last change and create a new, separate, pull request.
I'll merge the original one and then we can discuss on this new addition.

Last but not least, this whole thing must be documented well, otherwise the EFX function will get difficult to use. I can write it once the original PR gets merged.

grebecchi:
Ok, reverted.

I try to make a less confusing version of 'ChannelX' and 'ChannelY' modes (and I will create a new PR) because I believe that these modes are very powerfull.

With only one EFX you can create complex light effects saving a lot of programming time than using traditional scenes; with 5-10 EFXs you can create a cool dance or concert show, specially if you have complex fixtures (I have tested this modes with 10 ClayPaky Sharpy using gobos, frost, zoom and position animation).

mcallegari:
Merged ! :+1:

PR #521, opened 2015-08-23
Fixed Chauvet Intimidator Spot LED 150 Tilt Fine group

brunchboy:
Both tilt channels had the same byte value.


mcallegari:
Thanks :+1: 
(I renamed the patch title for history records :wink: )

brunchboy:
You’re welcome! And thanks for fixing the commit message, I apologize for not reading up on your standards first.

mcallegari:
That's fine, there are no standard rules here really. Just good sense for when someone needs to go through the project history :wink:

PR #524, opened 2015-08-30
Make Linux desktop files Freedesktop compliant

dguglielmi:
- Remove carriage returns
- Use standard category names from
  http://standards.freedesktop.org/menu-spec/menu-spec-1.0.html


mcallegari:
Hi, thanks.
The "Art" category still exists, but I guess nobody uses it

PR #526, opened 2015-09-01
SpeedDial improvements

grebecchi:
Hi developpers, in this push commit I propose some improvements to the speed dial.
These are the improvements:

1) MUL / DIV BUTTONS
I have added mult / div buttons on the right of the plus/minus buttons. These buttons multiply or divide for 2 the current timer values and are a very usefull shortcut to fast increase / decrease current time value staing in sync.
PS. Icons used for mult/div are the turned plus/minus icons, maybe can be replaced with better icons.

2) APPLY BUTTON
This button can be used to apply current time value to scenes, chaser, etc. 
So you can use multiple timer associated to a scene and change time on the fly only pressing a button.
You can also hide all other fields and leave only the apply button and create multiple buttons of different timer associated to the same scenes.
This button is only show on virtual console speed dial.

3) SAVE & RESTORE
Time value of the virtual console speed dial is saved on project. This is usefull for the apply button to store timers when you reopen a project.

qlc_duration


kripton:
Hello Giorgio,
thanks for your contribution that makes the Speed Dial even more useful.
Just two comments:
- Indeed, the MUL and DIV icons should be replaced since they look a bit confusing right now
- Just an idea: How about adding small (optional/hideable) buttons (4 to 6 of them) at the bottom of the widget to store time values instead of using multiple speed dials for one chaser? Think of it as the "memory"-function of a calculator. You modify the time, it doesn't apply until Apply is pushed. As an alternative, you click one of the memory-buttons with Shift or Ctrl pressed and the time is stored at that memory. If you click one of the memory-buttons without Shift or Ctrl , the time stored there is loaded and automatically applied to the function.

mcallegari:
Hi, these are nice and useful changes indeed.
My only fear is that this widget is becoming too complicated to use.
The more stuff we add, the more difficulties users will have to understand how it works.

I would say every QLC+ widget should have a "basic" mode (default when adding a new widget) and then several options in the configuration window, to turn the widget into an "advanced" mode.
Think for example about the Click & Go button on Sliders (disabled by default) or crossfade sliders in the Cue List widget (hidden by default).

I agree to improve it, but guys, let's keep it simple by default !

P.S. In QLC+ 5 I will completely rewrite the speed dials. They created too many problems in the past and I want them to be intuitive and easy to use. I still haven't reached a final design though...

grebecchi:
Hi Massimo, I can easily change the default visibility mask to hide by default the mult/div/apply buttons.

When you think that QLC+5 will be available? In other words do you think that working on SpeedDial has sense or it is a wast of time considering that this widget will be completely rewrite with the new QLC version?

What is the icon set used by QLC for other icons? Icons of mult/div should be replaced.

@kripton your idea is very cool but more complex to be implemented than the current Apply button. Maybe is better to wait the new SpeedDial rewrite of QLC5.

mcallegari:
Giorgio, there is no release date set for QLC+ 5 but QLC+ 4 will be surely be supported for the whole next year. So working on it at the moment is not a waste of time.

In QLC+ 5 I'm currently working on the Virtual Console. To implement Cue Lists, I will need to implement Chasers first and to implement Chasers, I will need Speed Dials (or whatever name they will be called).
Even though they will be rewritten, I will "port" the functionalities of QLC+ 4.

As for icons, I guess you won't find ready-made icons for x2 and /2.
The only suggestion I have is to simplify the "multiplier block" with 3 elements instead of 2:
- an arrow up button (to multiply by 2)
- a text box indicating the current multiplier: 1x, 2x, 4x or 1/2x, 1/4x, 1/8x, etc
- an arrow down button (to divide by 2)

In this way the user has also track of the current multiplier and doesn't have to remember it.

grebecchi:
In this push commits:

1) Mult / Div icons switched to  arrow up/down.
2) SpeedDial default visibility mask is now applied to all SpeedDials, before it was ignored. 
3) Changed the default visibility mask (mult/div and apply buttons are hidden by defaults)

If it is not enought clear for user I can add an additional text box for the current multiplier.

speeddial

Tolmino:
I think that a moltiplier text box maybe usefull

mcallegari:
Hi Giorgio, yes, please add a text box with the current multiplier.

I'm thinking about introducing a tab view or a button to show/hide "advanced" options.
In this way the widget will be basic by default and advanced if the user wants it to be so.

grebecchi:
Widgets positions should be aligned.

speeddial

mcallegari:
Just place the up/down arrows and the text box into a QVBoxLayout.

I'm not sure if it is a good idea to use a QSpinBox to display the multiplier.
Qt spin boxes work on contiguous integer or double values, but here we want discrete values like 1,2,4,8 and 1/2, 1/4, 1/8.
Maybe it's better to use just a QLabel

grebecchi:
Using up/down buttons we don't need discrete values but only integer values from 2 to ...
ie. 1/4 is equal to pressing down arrow with 4x factor.
Otherwise I can use a combo box with discrete values (1,2, 1/2, ...) and a single mult button.

This is the layout using a QVBoxLayout
speeddial

mcallegari:
Please commit what you have and then I'll fix it.

mcallegari:
But please explain exactly what the "Apply" button does, cause I don't understand it.
At the moment the speed dial adjustments are applied immediately, so unless the "Apply" button can be checked, I don't understand how you implemented it.

grebecchi:
The apply button is usefull if you need to fast change a scene fade/hold between a set of pre-defined values. 
You can use multiple dials associated to the same scene and set these dials with different times. Now you only need to press the apply button of one speed dial to set the required time without need to manually set time.

mcallegari:
So it is useful only the first time the user goes in Operate mode ?
As I said, all the changes to a speed dial, are immediately applied to the target function, so the apply button is useful only one time at the very beginning of a show.

I would prefer if it was a checkable button to enable/disable immediate changes on the target function.
By default it will be checked, to preserve the current behaviour and then the user can uncheck it, change the speed dial time, and check the button to apply the new time.
Sounds OK to you ?

grebecchi:
Not only the first time.
For example if I need to change duration of chaser steps on a live dance show generally I want change fast between some predefined times (1 beat, 4 beats, etc.). 
I can set 2, 3 or more dials to the same scene and set each dial to a different time value (1 beat, 4 beats, etc.) during the show I only need to press the apply button to immediately change the scene time without need to manually change the dial time.
You can fast tap between apply buttons of different speed dials (associated to the same scenes) to create interesting light effects on the fly.

mcallegari:
OK, got it now.
Basically this is a usage when you have many speed dials with fixed timings that you won't change during the show.
I was referring to the usage with a single speed dial.

Then maybe @kripton idea would be good concentrate both usages on a single dial, where you have a list of preset timings (or timings that you can store during a show) each with a button to call them.

Anyway, one step at the time :smiley:

lcslouis:
Having a speed dial with preset usage would be awesome

grebecchi:
I like the @kripton idea.
Probably the Apply button could be substituted with Save button. When user click on save a new preset button is created under the time edit boxes.

mcallegari:
After I slept on it, I came up with a solution that should make everyone happy :smile: 

We can do something similar to what I've done on the Animation widget. It requires a bit of development but I guess it would be worth it.
Basically we add a new tab to the speed dial configuration dialog to add custom presets. Each preset can have a key binding or an external controller binding.
Presets are shown as buttons at the bottom of the speed dial widget, in a FlowLayout.
Button labels can be the time itself or a mnemonic label entered by the user.
I would say the "infinite option" can be removed and treated as a custom preset, if needed.
There could be by default a "add" or "store" button in the FlowLayout to add a (temporary?) custom preset button while running a show.

In this way we should cover all the cases, and most likely one single speed dial will replace the usage of many SD in a VC, thus saving space.

What do you guys think ?
I am pretty busy coding QLC+ 5 but I can dedicate some time to this, cause in any case it will be ported to QLC+ 5. Otherwise someone can volounteer for this and I'll review the code. @plugz ? @grebecchi ?

plugz:
I think that the Animation widget should not be copied as-is.

It needs some more things: the ability to reorder controls, move things around, resize them...

The Speed Dial controls layout is kinda nice currently, and applying the same techniques as in the Animation widget would destroy this :/

grebecchi:
If @plugz is willing, I can continue to work to EFX to add the possibility to animate each fixture channel.

mcallegari:
@grebecchi please finalize this PR with what you have, then we take it from there.
Thanks

grebecchi:
This PR is updated.
It include:
- virtual console SpeedDial save & restore 
- apply button
- mult / div buttons + factor edit box

mcallegari:
Merged

PR #527, opened 2015-09-01
EFX gradient size should be 256x256

grebecchi:
Gradient size should be 255x255 because it is accessed by EFX when RGB mode is used.
After my initial commit, gradient x size was changed to 252 to reflect the ClickAndGoWidget size.

Now I have correct the Gradient and the  ClickAndGoWidget x size to 255.

### DIFF
--- a/engine/src/efxfixture.cpp
+++ b/engine/src/efxfixture.cpp
@@ -37,6 +37,7 @@
 /*****************************************************************************
  * Initialization
  *****************************************************************************/
+QImage EFXFixture::m_rgbGradient = QImage();
 
 EFXFixture::EFXFixture(const EFX* parent)
     : m_parent(parent)
@@ -56,6 +57,9 @@ EFXFixture::EFXFixture(const EFX* parent)
     , m_intensity(1.0)
 {
     Q_ASSERT(parent != NULL);
+
+    if(m_rgbGradient.isNull ())
+        m_rgbGradient = Gradient::getRGBGradient (256, 256);
 }
 
 void EFXFixture::copyFrom(const EFXFixture* ef)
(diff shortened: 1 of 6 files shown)


mcallegari:
This is wrong anyway.
Also, I renamed the title of this PR.

The calculation for 6 base colors is:
256 / 6 = 42.6666
Approximated to 42
42 * 6 = 252

Your change doesn't include a fix for the approximation, so resizing the pixmap to 256 pixels will produce a strip of 4 black (or random) pixels on the right.

A proper fix is to add a method to rescale the generated gradient pixmap to the desired size and use it in the EFX class. Something like this:

void  Gradient::resizePixmap(int width, int height)
{
    m_rgb = m_rgb.scaled(width, height);
}

I haven't tested it though

grebecchi:
Hi Massimo, I have reverted ClickAndGoWidget and Gradient size.
I have fixed this problem (EFX use x coordinates until 255 while Gradient has an image of 0-251 pixels) making a check on Gradient::getRGBColor. Input x and y parameters of Gradient::getRGBColor are now checked that contain a valid value.

mcallegari:
I'm afraid this is still not right.
Basically a Gradient is a lookup table when used by the EFX class. If the QImage is 252 pixels wide, you will still have 4 pixels handled not properly.
If you test your change with a Square EFX in RGB mode, you will notice that the last 4 horizontal pixels will return the same color. Not a big deal, but I guess this is not what we want.

I think the only way is to rescale the QImage to the needed size, as I suggested.
You just need to find a proper place in the EFX class to call the rescale method.

grebecchi:
Yes I know, this is a work around. 
X pixels are truncated to 252, the difference between 252 and 255 is a little difference, imperceptible in real RGB animations.

In the next commit I can change and store a static rescaled QImage in EFXFixture (so in it shared for all EFXFixture).

mcallegari:
Merged, thanks :+1:

PR #529, opened 2015-09-09
Cue List: Can change next/prev buttons behavior

plugz:
This allows to change the behavior of the next/prev buttons
while the chaser is not running.

http://qlcplus.org/forum/viewtopic.php?f=5&t=8694

2015-09-09-121422_4480x1440_scrot
2015-09-09-121429_4480x1440_scrot


mcallegari:
Yay ! More stuff to port to QLC+ 5 :smile: 
Just kidding :stuck_out_tongue_winking_eye: 

Just one thing: please don't cut words as they will be weird to translate. Use "previous".

plugz:
Sure thing :)

PR #530, opened 2015-09-09
RGB Scripts: add Plasma (Colors) script

nedmech:


mcallegari:
Thank you very much Nathan ! :+1: 

Just out of curiosity, where did you get the list of colors ?
I ask because in QLC+ 5 the color picker will allow selecting colors from preset lists.
Unfortunately Pantone and Lee Filters didn't agree to include their colors, but maybe makers like Caly Paky or Robe will.

nedmech:
Most of them were (I think) standard HTML color names, or at least close derivatives even if the RGB values don't quite match.  Some of the others (Cream, Seafoam, Electric Blue, etc.) were my own interpretation.  I tired to get 5-6 relatively even gradient colors around 5 basic hue groups: Red, Yellow, Green, Blue, Violet.  It was mostly trial-and-error until I came up with a set of color presets I liked and felt gave me sufficient flexibility for creating color-matched wall washes to go along with my video screen backgrounds.

janosvitok:
I don't understand why the color names are listed three times in the file. I'd expect some searching through the array... though I don't know much about javascript. Is there any reason?

Something like:

algo.setColor = function(_index, _preset)
{
    var i = colorPaletteList.split(',').indexOf(_preset);
    if (i == -1) i = 31;
    algo.colorIndex[_index] = i;
    return i;
}

algo.getColor = function(_index)
{
    var i = algo.colorIndex[_index];
    if (i > 31) i = 31;
    return colorPaletteList.split(',')[i];
}

Not tested.

colorPaletteList.split(',') may be cached (stored in another member) to improve performance.
Alternatively, the code can be rewritten to make colorPaletteList an array and use join() to produce the string.

nedmech:
Thank you, janosvitok!  That would be a great improvement.  I originally just brute-forced this based on the original plasma.js script.  What you're suggesting would greatly clean up the implementation and make it much more flexible for future palette expansion of changes.  I'll look at it today and see what I can make happen!

PR #538, opened 2015-09-13
New fixture - Eurolite LED FE-1500

boxy321:
Created 28/07/15


mcallegari:
Seems good. Just one little thing. Modes shouldn't contain the word "mode" because it's implicit.
Basically this should have just "8-channel"
Anyway, doesn't matter. I can fix it later.
Thanks

PR #539, opened 2015-09-13
New fixture - Lightronics FXLD618C2I

boxy321:
The part number ends in I, not 1 (one).


mcallegari:
Same as before. Modes shouldn't have the word "mode".

Thanks :+1:

mcallegari:
Ah, also I'm gonna change the white channels to Intensity->White, instead of generic.

boxy321:
I included 'mode' only where it is mentioned in the manual's channel name or to distinguish between names of other similar channels. I agree that it is redundant so sure.

PR #541, opened 2015-09-13
New fixture - American DJ Fog Fury Jett

boxy321:
Created on 01/09/15.

Modifed to add RGBA colour macros with debatable mapping to RGB values.


mcallegari:
Thanks, I'll fix "mode"

PR #542, opened 2015-09-14
New fixture - Eurolight LED KLS-401

boxy321:
Created on 04/09/15.

Red and green on head 1 are multipurpose so the channels are effect, not intensity. The effect colours are correct however.


mcallegari:
Thanks :+1:

PR #545, opened 2015-09-15
New fixture - XStatic X-240Bar RGB IRC

boxy321:
Created on 22/08/15. Corrected absence of capabilities for RGB modes.

Manual that matches the fixture is found at:

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8829

### DIFF
--- a/resources/fixtures/XStatic-X-240Bar-RGB-IRC.qxf
+++ b/resources/fixtures/XStatic-X-240Bar-RGB-IRC.qxf
@@ -1,0 +1,383 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE FixtureDefinition>
+<FixtureDefinition xmlns="http://qlcplus.sourceforge.net/FixtureDefinition">
+ <Creator>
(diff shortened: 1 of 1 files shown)


mcallegari:
Uhm...is this the same of what's already in the library ?
https://github.com/mcallegari/qlcplus/blob/master/resources/fixtures/XStatic-X-240Bar-RGB.qxf

Or the IRC version is different ?

boxy321:
It was on Git but not yet on the distro I had. I have added capabilities to the definition anyway. It seems the IRC version is the same😠

mcallegari:
When working with GIT, you should be always updated to the latest version.

In case, can you update the existing one if it's missing something ?

mcallegari:
@boxy321 What about this one ? Shall I close it ?

boxy321:
You could update the existing. I think I improved it a bit. Just doing an Ingress Op.

mcallegari:
Done !
Now there's basically your copy on GIT.
Thanks :+1:

PR #551, opened 2015-09-15
New fixture - Robe ClubWash 600 CT

boxy321:
Created on 06/09/15

Nearly every capability did not match the datasheet in some way so all had to be changed from the original post:

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8884
http://www.robe.cz/it/support/list-files/15/file-detail/4509/file-cat/robe-downloads/


mcallegari:
And this is fixture number 600 ! Yay ! :smiley:

boxy321:
Wow! Sorry about the earlier descrepancies. Some definitions are awful and it's like starting from scratch.

mcallegari:
Ouch, I've deleted my comment cause we're not at 600 yet, sorry.
Just one more to go !

mcallegari:
> Wow! Sorry about the earlier descrepancies. Some definitions are awful and it's like starting from scratch.

Yeah, I know. Sometimes people just share crap that seems fine to them :(

PR #555, opened 2015-09-15
VCSpeedDial Presets

plugz:
VCSpeedDialProperties: Add a tab to edit presets
When clicking on a preset, the speed dial value is changed
Like the VCAnimation, it's possible to assign an external input to each preset button

TODO:
- Add a button on the SpeedDial to "Save" current value in a new preset in live
- rework the multiplier functionnality ?

### DIFF
--- a/ui/src/speeddial.cpp
+++ b/ui/src/speeddial.cpp
@@ -241,6 +241,9 @@ void SpeedDial::setValue(int ms, bool emitValue)
     else
         m_infiniteCheck->setChecked(false);
 
+    // Time has changed, stop blinking
+    stopTimers(false, true);
+
     m_preventSignals = false;
 }
 
--- a/ui/src/src.pro
+++ b/ui/src/src.pro
@@ -146,6 +146,7 @@ HEADERS += virtualconsole/addvcbuttonmatrix.h \
            virtualconsole/vcsoloframeproperties.h \
            virtualconsole/vcspeeddial.h \
            virtualconsole/vcspeeddialfunction.h \
+           virtualconsole/vcspeeddialpreset.h \
            virtualconsole/vcspeeddialproperties.h \
            virtualconsole/vcwidget.h \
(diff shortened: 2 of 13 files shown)


plugz:
2015-09-15-180405_4480x1440_scrot

mcallegari:
Awesome stuff !
As I mentioned before, now the "Infinite" option can be a preset button as well.
For backward compatibility, those having the infinite option enabled will find a preset button, otherwise no action should be taken.
Then we could mark the option as "legacy"

> TODO: rework the multiplier functionality ?
Anything in particular ?

plugz:
> As I mentioned before, now the "Infinite" option can be a preset button as well.
> For backward compatibility, those having the infinite option enabled will find a preset button, otherwise no action should be taken.

I'm working on it, almost done

> >TODO: rework the multiplier functionality ?

> Anything in particular ?

> [EDIT] Concerning the multiplier feature, there is something affecting the ms field (again :cry: )
Enable the buttons. Set the dial to 1s.42. Click on divide a few times until you reach 1/16x.
The ms field shows ".8"

Well, lots of things.. This, and.. lets say you have a speeddial @ 1.00s, you multiply to x4 you're @ 4.00s, then move the big dial button to get @ 200.00s, then crank up the multiplier to x8 -> you get back to 8.00s.

Also, changing the values from the input fields seem not to affect the "original value" from which the multiplier get its new values.

I was thinking of getting the multiplier out of the speeddial, and put it as part of the VCSpeedDial, so it's simply applied on top of the current speeddial value. No more complicated back-and-forth between original value and multiplied value.


> [EDIT2] Watching your screenshot I was thinking...what about if the speed dial widget was embedded in the configuration page ? (basically instead of the clock icon button)
Dunno if it fits, but personally I find those floating windows frustrating.

I've tried this, and it does not fit very well, but it can still be better than the floating window.

PR #560, opened 2015-09-16
New fixture - Robe Robin MiniMe

boxy321:
Created on 16/09/15.

I can no longer see so please put an evening aside to check this.

http://www.robe.cz/products/article/robin-minime/#downloads


mcallegari:
Wow, that's indeed a monster fixture.
Well done !

@boxy321 you might be interested in the latest changes to the Fixture Editor. Basically I unified the channel editing window with the capability editing window. It should help to speed up the creation/review of a fixture.
If you build from sources, please check it out.

image

PR #561, opened 2015-09-16
Fix directory names in variables.pri

daveol:
Fix the directory names in variables.pri so that they all have consistent capitalization across platforms.

### DIFF
--- a/variables.pri
+++ b/variables.pri
@@ -79,88 +79,88 @@ android:USERDATADIR    = .qlcplus
 
 # Documentation
 win32:DOCSDIR      = Documents
-unix:!macx:DOCSDIR = $$DATADIR/documents
+unix:!macx:DOCSDIR = $$DATADIR/Documents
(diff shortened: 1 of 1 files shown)


janosvitok:
I don't think this is improvement. Each platform has its own traditions, and I prefer if we stick to each platform's style rather than trying to unify them.

What benefits does this bring?

Besides that, to accept this more things are needed: 1. fix paths in tests (see travis output) 2. update debian/qlcplus.install

daveol:
Well first of all [mcallegari/qlcplus-extras](/mcallegari/qlcplus-extras) would actually work on linux without moving directories around, secondly most of the changes are in the user data directory and are names for things like RGB and since RGB is a acronym it should be in capitals.

I can fix the path in the test, but if you say you won't merge it I won't bother.

mcallegari:
Sorry, but I downvote this too.
It breaks everything currently installed and that's not what we want.

PR #563, opened 2015-09-22
E131 unicast configuration

plugz:
- Allow unicast input and output
- Configure unicast output port and input port
- Allow changing E1.31 universe for input and output

2015-09-22-150826_4480x1440_scrot

Tests done:
- Unicast input/output seem OK
- Multicast output seems OK
- Multicast input : not tested

### DIFF
--- a/plugins/E1.31/configuree131.cpp
+++ b/plugins/E1.31/configuree131.cpp
@@ -154,6 +154,8 @@ void ConfigureE131::fillMappingTree()
                     multicastCb->setChecked(false);
                     m_uniMapTree->setItemWidget(item, KMapColumnIPAddress,
                             new QLineEdit(info->outputUcastAddress.toString()));
+                    if (QHostAddress(controller->getNetworkIP()) == QHostAddress::LocalHost)
+                        m_uniMapTree->itemWidget(item, KMapColumnIPAddress)->setEnabled(false);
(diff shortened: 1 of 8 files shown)


plugz:
@mcallegari What about integrating this in TEST versions to get some feedback ? :)

mcallegari:
Uhm...I was looking at the screenshot.
I think a multicast class address can be autodetected, so there shouldn't be the need for a checkbox
I've seen there's a new API in Qt 5.6 (QHostAddress::isMulticast)
Code here:
http://code.qt.io/cgit/qt/qtbase.git/tree/src/network/kernel/qhostaddress.cpp#n1110

Seems quite simple after all. Maybe we can copy it ?

plugz:
According to the E1.31 spec, multicast should always be done one port 5568.
The checkbox allows disabling the port edit combobox, I find this makes things more explicit.
But yes, autodetection of the type of address is not complicated, maybe for 4.10.1 ? hehe

> [EDIT] in any case, a multicast checkbox on 127.0.0.1 would be wrong :)

I don't know. It's possible to join a multicast group on the loopback interface. I think the user should be able to do anything.
Plus, some software (including ola :( ) do not allow unicast configuration for E1.31.

mcallegari:
OK, let's go with this then :+1: 
Maybe we can refine it later if needed or depending on feedbacks

PR #564, opened 2015-09-23
New fixture - Strip LED RGB Code K2014

boxy321:
The datasheet is poor and the manufacturer website doesn't work so this is marginal between here and 'extras'.

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8946


boxy321:
Manufacturer omitted: PSL

mcallegari:
Hi, thanks.
PSL looks to me like the typical reseller that buys chinese stuff and rebrand it.
This is their site: http://www.professionalsoundlight.it/ (obviously only in italian...duh)
I would put this on qlcplus-extra

mcallegari:
Pushed to qlcplus-extras

PR #566, opened 2015-09-23
Updated catalan translation UI and OSC plugin

sbenejam:
Updated catalan translation UI and OSC plugin


mcallegari:
Thanks Santiago :+1:

PR #574, opened 2015-09-26
Language fixes

janosvitok:
1. added Czech to various files like NSIS installer, linux desktop files, etc.
2. added all languages that have translation to NSIS installer
3. added new types (qxi, qxmt) to mime-types definition

I suggest that translators have a look at these files as well - many of the translations are missing.

I didn't test NSIS installer how it looks like with these changes.


mcallegari:
@sbenejam and @tornel you might be interested in this too :)

tornel:
@mcallegari PR already sent!
:smile:

PR #575, opened 2015-09-26
Update Spanish translation (E1.31 + UI)

tornel:


mcallegari:
@tornel what are the changes to the italian translation ? :open_mouth:

tornel:
an error on my part when I clicked save and shouldn't (too little sleep and too much coffee  :disappointed:)
corrected in those two last commits

mcallegari:
Alrighty then, thanks :+1:

PR #580, opened 2015-09-29
New fixture - Cameo-LED-Moving-Head-60W-CLMHRGB60W

boxy321:
Original post:

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8971

Datasheet:

http://www.thomann.de/gb/cameo_moving_head_60.htm

### DIFF
--- a/resources/fixtures/Cameo-LED-Moving-Head-60W-CLMHRGB60W.qxf
+++ b/resources/fixtures/Cameo-LED-Moving-Head-60W-CLMHRGB60W.qxf
@@ -1,0 +1,151 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE FixtureDefinition>
+<FixtureDefinition xmlns="http://qlcplus.sourceforge.net/FixtureDefinition">
+ <Creator>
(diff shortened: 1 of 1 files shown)


mcallegari:
Yes it would :)
When the 4.10.0 storm has passed, I could dedicate some time to various pending things

boxy321:
The catalogue (pdf page 5): http://www.cameolight.com/startseite/?force_sid=l0jhu2d9cqktljdpjuem0980e2 shows the same image. However, there are no matching gobos in the library anyway.

mcallegari:
OK, I've had a look and actually we don't have similar matches.
Then there is no solution other than:
- leaving them blank
- ask Cameo to officially provide their gobos. If you want I can send them an email and see if they reply

For now I would go for the first option

boxy321:
You do first. I can do second option???


Sent from my Samsung device

mcallegari:
Let's see if tasaei replies in the forum.
Otherwise leave this to me and I'll sort it out.

boxy321:
👍


Sent from my Samsung device

PR #582, opened 2015-09-30
New fixture - American DJ Event Bar Q4

boxy321:
Original post here:

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8978

There were channels missing, no physical data and incorrect channel types.

There are two versions of this fixture with different capabilities. It is possible to use both by selecting the required mode.


mcallegari:
Great, thanks ! :+1:

PR #586, opened 2015-10-01
Update fixture - Cameo-LED-Moving-Head-60W-CLMHRGB60W with gobos

boxy321:
The author has provided the gobos and the definition refers to local the directory Cameo/xxx.png


mcallegari:
@boxy321 those gobos are not OK for the following reasons:
1- we don't have the Cameo approval to use them
2- the size of the pictures should be 64x64 pixels
3- the shape should be a circle
4- the circle should be black and all the rest transparent
5- file names should be gobo000xx.png

I tried to adapt them manually, but it's impossible to scale some of them (e.g. Gobo 2) without loosing quality. Please stand by on this. I'm gonna contact Cameo

The ideal solution for gobos would be using SVG files. Especially for QLC+ 5.
This would mean to review all of them and all the fixture files.

boxy321:
Taken on-board. SVG! That will be a job.


Sent from my Samsung device

mcallegari:
@boxy321 no reply from Cameo.
If they don't reply within tomorrow, I will merge your PR without gobos (the one I closed) and this fixture will remain without gobos.
I see no other option at the moment.

btw, new fixtures in the forum :smile: 
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8987
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8988

PR #587, opened 2015-10-05
VCSpeedial: put mult/div buttons outside of the dial

plugz:
- The multi/div buttons are applied AFTER the dial's value

- The resulting value is displayed on the side

- There is a reset button to reset the current factor

- There is a checkbox that makes the factor reset when
the dial value changes (example: click on the +/- button
or select a preset)

The speed dial:
speeddial
The properties:
speeddialproperties

- An input can be assigned to each button

I'm not sure about the style, it looks kinda big and ugly, and the result label is not visible enough. (translation: it looks like shit)
But the behavior is less random than the current implementation :)

### DIFF
--- a/ui/src/speeddial.cpp
+++ b/ui/src/speeddial.cpp
@@ -89,11 +89,9 @@ SpeedDial::SpeedDial(QWidget* parent)
     , m_previousDialValue(0)
     , m_preventSignals(false)
     , m_value(0)
-    , m_originalValue(0)
     , m_tapTime(NULL)
     , m_tapTickTimer(NULL)
     , m_tapTick(false)
-    , m_currentFactor(1)
     , m_visibilityMask(DEFAULT_VISIBILITY_MASK)
 {
     new QVBoxLayout(this);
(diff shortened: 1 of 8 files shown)


mcallegari:
Wow, lot of stuff. What happened to the "apply" button ?

I have some changes to the VC classes, but they will probably conflict with this :cry: 
Althogh they add custom feedbacks ! :smiley:

plugz:
I didn't know where to put the apply button so I haven't put it back yet. Maybe at the bottom of the dial, before the presets ?

>I have some changes to the VC classes, but they will probably conflict with this :cry:
Althogh they add custom feedbacks ! :smiley:

You can break the merge of this PR, I'll adapt it to the new VC classes :)

mcallegari:
Committed. Apparently this PR didn't break ! Yay :smile:

mcallegari:
@plugz merged. I reworked the layout a bit. How about this ?

image

plugz:
@mcallegari Yes nice :) I'll do the apply button soon

mcallegari:
Wait. I commit this first :)

PR #588, opened 2015-10-06
New fixture - Robe ColorSpot 575 AT

boxy321:
Manual here:
http://www.robe.cz/nc/support/file-detail/4723/

Requested here:
http://www.qlcplus.org/forum/viewtopic.php?f=6&t=8993

This is similar to the Robe Robin 300E Beam for which I'm still waiting confirmation of the channels:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=8987


mcallegari:
Gobos paths are not OK. They point to
/Applications/QLC+ 4.8.4.app/Contents/MacOS/../Resources/Gobos/Robe/

I'll merge and then fix this, since I realize there are others in the library with OSX paths

mcallegari:
@janosvitok the update-map.rb script fails on this, because Robert created it with a previous version of the Fixture Editor, so they still have xmlns="http://qlcplus.sourceforge.net/FixtureDefinition".

Can you please loosen it a bit or even better "auto-fix" those wrong URLs ?
Thanks

PR #595, opened 2015-10-06
Update vcspeeddial.cpp

bestdani:
Fixed bug that caused a doubled multiply / divide / reset when using midi input


mcallegari:
Makes sense, thanks !
Can't even imagine if one associates a fader to those buttons... (by mistake of course :smiley: )

PR #604, opened 2015-10-11
Also translate new strings

kripton:
Hi,
I felt that there were some strings missing (at least the two for the Apply-Button of the SpeedDial-widget) so I ran lupdate on all .pro-files. That action updated the line numbers on all .ts-files (I'm not sure if this is wanted) and added ts-files for the plugins GPIO, Dummy and UART (not sure if this is wanted as well).
Then I translated the missing strings.

Feel free to cherry-pick instead of merging this branch.

Oh, and make sure to merge the "german" branch first since this branch depend on "german"

### DIFF
--- a/etc/qlcplus.xml
+++ b/etc/qlcplus.xml
@@ -55,6 +55,7 @@
 		</magic>
 		<glob pattern="*.qxm"/>
 		<comment xml:lang="cz">QLC+ MIDI šablona</comment>
+		<comment xml:lang="de">QLC+ MIDI-Vorlage</comment>
 		<comment xml:lang="en">QLC+ MIDI template</comment>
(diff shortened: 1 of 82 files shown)


mcallegari:
Yeah, thanks but the Dummy plugin is there only for "educational" purposes so it doesn't need to be translated.
So basically I can't merge this as it is.
Would you mind if I commit the updated files without the Dummy plugin and then you commit your german translations ?

kripton:
Yes sure, take what you need and leave out the unwanted things. If you need any action from my side, just tell me :)

mcallegari:
@kripton OK, I have pushed the updated files with your translations included.
I didn't merge the one in UART plugin cause I changed the string to "output" only.
So you have just one string to translate now :)
Closing this then. Thanks.

PR #605, opened 2015-10-11
Update UI, DMXUSB, UART and GPIO spanish translations

tornel:


mcallegari:
A bit messy but it worked :smile: 
Thanks !

PR #612, opened 2015-10-16
New fixture - Acme-LED-MS50A-Genesis-Move

boxy321:
Original post here:

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9015

Datasheet pdf directly from google search.

I have removed the gobos from this definition. There are some images we could use but others are unique (see page 13).

This fixture is almost identical to the Acme MS350A apart from gobo colours and gobo ordering (I will add later)


mcallegari:
Uhm...this looks to me like a qlcplus-extra fixture.
ACME is chinese: http://www.acme.com.cn/Contact.aspx?lang=en

I would give the priority to more popular fixtures like Showtec-Dragon-F-350:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9010

mcallegari:
Pushed to qlcplus-extra

PR #618, opened 2015-10-28
New fixture - American-DJ-Starburst

boxy321:
Please check channel 6. I've used UV instead of purple as it is the nearest colour. It would be nice to select a generic colour instead of the fixed ones available.

http://www.americandj.eu/en/starburst.html


mcallegari:
Why not Magenta ?

PR #619, opened 2015-10-28
New fixture - Chauvet-LED-PAR-64-Tri-B

boxy321:
Same functionality as Chauvet SlimPAR 64 so copied it.


mcallegari:
Perfect, thanks

PR #620, opened 2015-10-28
New fixture - BoomToneDJ-Strob-LED-18

boxy321:
Posted on forum with datasheet, Mon Oct 26th.


mcallegari:
Same as before "BoomToneDJ" with no spaces

Issue #592, opened 2015-10-06
EFX Editor regression


mcallegari:
After the changes introduced by Giorgio Rebecchi, the EFX capabilities (Position, RGB, Dimmer) are now autodetected from the fixture heads.

Problem is that if a head has RGB channels but not Pan/Tilt (and the fixture definition do have them) the EFX editor doesn't show the "Position" capability.

Now, we cannot check every single fixture definition, so this has to work regardless, as it is a regression to what we had before

mcallegari:
Should be fixed.
The only counter effect in the EFX editor is that selecting a Fixture with 2 heads will show the Position capability for both heads.
Users should select just one head if they want to use just positions and not RGB or Dimmer

PR #625, opened 2015-10-29
Added few fixtures

lachmex:
Contest Evora DUO B2R
Contest Oz 37x15QC
Evora Beam 15R
Evora Beam 5R


mcallegari:
@lachmex I've seen the product. I think the correct heads should include the channels specific to the two bulbs.
The EFX editor didn't work for you cause a regression went into 4.10.x and it has now been resolved in GIT: see #592

There are TEST versions in the forum with the issue fixed. Can you please try one of those ?
http://www.qlcplus.org/forum/viewtopic.php?f=12&t=3135

lachmex:
wow, with pleasure :) I will try to test it on monday and will let you know ... Thank you

PR #627, opened 2015-10-30
DMX monitor: Show data from passthrough

janosvitok:
Change processing of DMX input data so that:
- the input data is not affected by the grandmaster and channel modifiers
- the input data is merged with QLC+ in a HTP fashion (for all channels, regardless of channel settings)
- the input data is shown in DMX monitor

Usage scenarios:
- QLC+ as DMX sniffer (this requires to patch enough dimmer channels)
- QLC+ as DMX visualizer (in 2D view mode, requires proper fixture patching)
- QLC+ on raspberry just forwards data from QLC+ on PC while programming scenes;
  when the workspace is transferred, Raspberry becomes main controller; the devices are
  always connected to RPi


plugz:
nice

PR #639, opened 2015-11-02
Fixture update - Chauvet-SlimPAR-Pro-RGBA

boxy321:
Added physical data and missing 10-channel mode with associated channels.

Datasheet:
http://www.chauvetlighting.co.uk/slim-par-pro-rgba.html


mcallegari:
I merged this manually cause you renamed the whole fixture file, thus breaking backward compatibility

boxy321:
My bad.

PR #640, opened 2015-11-02
Fixture update - Chauvet-SlimPAR-Pro-VW

boxy321:
Updated physical info and added 7-channel mode and related channels. NOTE: I entered 10-channel instead of 7 as the mode. Please update before adding.

Datasheet

http://www.chauvetlighting.co.uk/slimpar-pro-vw.html


mcallegari:
Merged manually cause of fixture rename

boxy321:
Fixing what I can on the way for consistency.

PR #641, opened 2015-11-02
Add manual sort for presets on vcxypad

grebecchi:
Hi everyone, this pull request add the possibility to manual sorting presets on vcxypad, that is very usefull for widgets with a high preset number.
I have implemented the behaviour and look & feel with up/down buttons similar to other components.

preset sorting

### DIFF
--- a/ui/src/virtualconsole/vcxypadproperties.cpp
+++ b/ui/src/virtualconsole/vcxypadproperties.cpp
@@ -524,10 +524,11 @@ void VCXYPadProperties::selectItemOnPresetsTree(quint8 presetId)
         if (treeItem->data(0, Qt::UserRole).toUInt() == presetId)
         {
             treeItem->setSelected(true);
-            m_presetsTree->blockSignals(false);
-            return;
+            break;
         }
     }
+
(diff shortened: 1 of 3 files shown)


ui/src/virtualconsole/vcxypadproperties.cpp
```diff
@@ -510,6 +514,22 @@ void VCXYPadProperties::updatePresetsTree()
...
+    {
+        QTreeWidgetItem* treeItem = m_presetsTree->topLevelItem(i);
+        if (treeItem->data(0, Qt::UserRole).toUInt() == presetId)
+        {
+            treeItem->setSelected(true);
+            m_presetsTree->blockSignals(false);
+            return;
+        }
+    }
+}
```
janosvitok: If the item is not found, the signals stay blocked forever. QSignalBlocker would be useful here, but it's a Qt5 class. It might be worth to backport it for our use until we drop Qt4 support.

Without QSignalBlocker, m_presetsTree->blockSignals(false) between lines 530 and 531 -OR- comment stating why it's not needed there.

Issue #629, opened 2015-11-01
Fixtures with missing physical dimensions


mcallegari:
Just a reminder for who wants to pick this up. @boxy321 ? :wink: 
The following fixtures miss physical dimensions.

EDIT 20151102: shortened list
EDIT 20151103: thanks to Jano now we have also Pan/Tilt information

```
BeamZ-LS-3DRG.qxf: PAN TILT
Eurolite-TB-250.qxf: PAN
JB-Lighting-Varyscan-2.qxf: PAN TILT WIDTH HEIGHT DEPTH WEIGHT
Laserworld-EL-200RGY.qxf: PAN TILT
Lite-Works-ColorChanger.qxf: WIDTH HEIGHT DEPTH WEIGHT
Sagitter-Miniscan-2001.qxf: WIDTH HEIGHT DEPTH WEIGHT
StageTech-LeaderScan-Roto.qxf: PAN TILT
```

janosvitok:
Do we want to change Fixture Editor so that it doesn't allow zero values? Alternatively add button to do final checks on the definition?

mcallegari:
Well, for personal use I don't care what people do. So a fixture without physical information is OK as long as they don't submit it in the forum.
As for a validation button, it is a good idea but I have higher priority tasks to carry on. If you want to implement it please do it :)

boxy321:
Regarding backward compatability here's an example:

Robe ClubSpot 150CT

Physical data: no problem, can add. Modes: there is one missing. If I add the missing one we have to rename the existing ones as nobody uses the names in the manual, ever. What effect would this have? Not good I expect, if somebody has a bunch of effects etc.

mcallegari:
> If I add the missing one we have to rename the existing ones as nobody uses the names in the manual, > ever. What effect would this have? Not good I expect, if somebody has a bunch of effects etc.

If you rename a mode, it breaks backward comaptibility. If you just add one it doesn't.

mcallegari:
Alrighty, closing this for now.
Thanks everyone for the efforts

PR #647, opened 2015-11-03
Fixture update - Showtec-Spectral-CYC650

boxy321:
Physical data added. Tour mode had a missing channel (ID address) so was incorrect in that mode.

Google for datasheet:

showtec spectral cyc650


mcallegari:
Hey Robert, I've updated this yesterday.
Check the updated list in #629 and sync the latest changes

PR #664, opened 2015-11-05
New fixture - Blizzard-Lighting-Lil-G

boxy321:
There is a 'cut sheet' link showing some gobos for this but there are only 7 (out of 9) and no info on the order of them, so they're blank.

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9124
http://www.blizzardlighting.com/products/moving-head-lights/item/176-lil-g


mcallegari:
Please don't shorten gobos capabilities like "1, 2, 3, ..."
Most of the times in the manual they're "Gobo 1, Gobo 2, Gobo 3, etc.."
Let's stick with it

boxy321:
Same as Modes though surely, they're implied. When there's 64 colour macros, do you need to repeat the word 'macro' 64 times?

Issue #672, opened 2015-11-09
Bug with cue list hitting next, then crossfader and then next again


anorod:
I have a cue list with different scenes, some of them have the fade timing set, but others I need to fade it manually because the scene depends on some actors. The problem with 4.10.1 that was not before is that when I hit next and then do the crossfade, when I hit next again later it continues since the last cue I did with the next button.

Eg:
I have 10 scenes in a cue list. The first 5 with timing so I hit the next cue button. After the 5th, I use the crossfader with 3 more so I am in cue 8th. If I hit next now instead of use the crossfader, I will jump to cue 6 instead of 9.

I've tested it with MacOS 10.9.5 and Windows with the same results. Also I've tested versions 4.10.1, 4.9.1, 4.8.5 and it's the same.

I've been looking at the code and while hitting the next button just do ch->next() in line 619:
  https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L619

In the sliders it change manually and it doesn't seem to update the current cue number. Maybe the bug could be solved updating the current cue number after the following lines:
 - For Slider1: https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L998
 - For Slider2: https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L1048

I'm guessing it may be something similar to *ch->setCurrentStep(ch->currentStepIndex() + 1);* but not sure.

Thank's for a great software!

anorod:
Tested plugz fix on the 4.10.1 and seemed to work fine except one thing:
- If at the end of using the sliders they slider are "upside down" (slider #1 to 0 and slider #2 at 100%) it does the next cue without the fade in value and then all times you hit next the DMX values are all 0 (probably because they are multiplying by the slider #1)

mcallegari:
Fixed by #674

PR #674, opened 2015-11-10
Cuelist: fix issue when using the next button after the crossfader

plugz:
Fix for #672 

When using the faders, the current step moves in the VCCueList,
but not in ChaserRunner.
ChaserRunner::m_lastRunStepIdx is the last step that was run
before using the faders.
This means that when calling ChaserRunner::next(), the next index
it not the one pointed at by the VCCueList.

Fix: let ChaserRunner::stopStep change the current chaser step, and inform the VCCueList through the currentStepChanged() signal.

### DIFF
--- a/engine/src/chaserrunner.cpp
+++ b/engine/src/chaserrunner.cpp
@@ -251,6 +251,11 @@ void ChaserRunner::stopStep(int stepIndex)
             delete step;
         }
     }
+
+    if (m_runnerSteps.size() == 1) {
+        m_lastRunStepIdx = m_runnerSteps.at(0)->m_index;
+        emit currentStepChanged(m_lastRunStepIdx);
+    }
 }
 
 void ChaserRunner::setCurrentStep(int step, qreal intensity)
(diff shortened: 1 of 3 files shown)


anorod:
Tested plugz fix on the 4.10.1 and seemed to work fine except one thing:

If at the end of using the sliders they slider are "upside down" (slider #1 to 0 and slider #2 at 100%) it does the next cue without the fade in value and then all times you hit next the DMX values are all 0 (probably because they are multiplying by the slider #1)

anorod:
Also if you move the sliders a bit and then come back to the original state (imagine that you start changing the scene but then decide to use the timer), the fade in of the next cue doesn't work either, only the fade out.

plugz:
Thanks for the feedback, I'll investigate this further.

anorod:
Thank you a lot.
In case it helps, to test it I create a fixture of 10 generic channels, then 10 scenes one putting up one channel at the time in order and then a chaser implementing all those scenes with infinite duration and some random fade in and fade out times. After that I just create a cue list in the VC and test to hit next or work with the sliders.

plugz:
I've found that the issues you highlighted already happened before applying this fix.

anorod:
I've tested them and it seems so. Problem was that before it was not easy to see them because of the issue #674.

Do you want me to add another new issue about it?

Thanks

mcallegari:
Is this PR complete now ?
Should we merge ?

anorod:
Yes. I will open another with the issues about crossfading that I said before because they are independent of this one.

Thanks Massimo

PR #676, opened 2015-11-10
Add support for the AVLdiy.cn UDMX clone with non-standard VID/PID

cheetahpaw:
Propose to add support for the UDMX clone from AVLdiy.cn with non-standard VID/PID 03eb:8888.

I bought this cable from DealExtreme and found out, that it doesn't work with QLC+ because of different USB vid/pid. Device is tested and works with changes, proposed in my commit.

Detects as:
[39684.701431] usb 3-1.3: New USB device found, idVendor=03eb, idProduct=8888
[39684.701437] usb 3-1.3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
[39684.701440] usb 3-1.3: Product: D512
[39684.701442] usb 3-1.3: Manufacturer: AVLdiy.cn
[39684.701445] usb 3-1.3: SerialNumber: AVLdiy.cn0001

PS: It's my first contribution to the open source project, so please point out any mistakes.

### DIFF
--- a/plugins/udmx/src/udmxdevice.cpp
+++ b/plugins/udmx/src/udmxdevice.cpp
@@ -35,6 +35,23 @@
 
 #define UDMX_SHARED_VENDOR     0x16C0 /* VOTI */
 #define UDMX_SHARED_PRODUCT    0x05DC /* Obdev's free shared PID */
+
+#define UDMX_AVLDIY_D512_CLONE_VENDOR     0x03EB /* Atmel Corp. (Clone VID)*/
+#define UDMX_AVLDIY_D512_CLONE_PRODUCT    0x8888 /* Clone PID */
+
+#define UDMX_USB_SIGNATURES_COUNT  2
+
+/* List of UDMX and clones VID/PID */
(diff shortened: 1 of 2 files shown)


mcallegari:
Looks OK to me. I'd only simplify the "signature" check with a code similar to what we have in DMX USB and Peperoni.

Out of curiosity: I am interested in having a Anyma adapter for testing, but the original one is freaking expensive (180€) !
Can you please post the link to the DealExtreme page where you found the clone ? Thanks.

mcallegari:
Can you please check if after 5488c06fc02981d96949a8babbcf9b352818d182 everything is still working OK ? Thanks.

cheetahpaw:
I checked and confirm that adapter works after 5488c06, which is great.

It was bought here: http://www.aliexpress.com/item/Free-Shipping-USB-to-DMX-Interface-Adapter-DMX512-Computer-PC-Satge-Lighting-Controller-Dimme/1948534712.html (not DX, but AliExpress), for about $19.

PR #679, opened 2015-11-15
Channel editor: allow temporary invalid cap ranges

plugz:
A message is displayed as long as the range is invalid, and the values are not applied.

http://qlcplus.org/forum/viewtopic.php?f=5&t=9167&start=10

### DIFF
--- a/engine/src/qlccapability.cpp
+++ b/engine/src/qlccapability.cpp
@@ -52,16 +52,7 @@ QLCCapability *QLCCapability::createCopy()
                                             m_resourceColor1, m_resourceColor2);
     return copy;
 }
-/*
-QLCCapability::QLCCapability(const QLCCapability* capability)
-{
-    m_min = 0;
-    m_max = UCHAR_MAX;
 
-    if (capability != NULL)
-        *this = *capability;
-}
-*/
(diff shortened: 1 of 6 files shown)


mcallegari:
Alright, apart for the issues that I have introduced ( :cry: ) the whole point of this was not allowing invalid ranges at all.
I would suggest to disable the dialog OK button as long as there are invalid ranges.
Thing is that the invalid capabilities should be somehow highlighted, for example in red.

Otherwise, on invalid ranges, when the OK button is pressed, a popup should show up, telling the user which capabilities are invalid.

Dunno which one is easier to implement.

plugz:
With this patch, the invalid ranges are still not allowed.
As long as the "invalid range" message is displayed, the comboboxes can still be moved, but the capability is NOT changed.

You can see the behavior in the gif I uploaded: http://qlcplus.org/forum/download/file.php?id=6713&mode=view

In the tree view, the values change only when the selected range is valid.

The invalid range cannot be applied to the channel by the user, even by accident.




It may not be the best solution, it's just one I came up with.

mcallegari:
OK, got it. Let's merge

PR #680, opened 2015-11-16
Update QT links

dadoonet:
Website has moved. Download pages are now at http://download.qt.io/official_releases/qt/


mcallegari:
Hi, thanks. Well, the official download page is actually this: http://www.qt.io/download/
But OK, a direct reference to the packages repository avoids confusion.

dadoonet:
Yeah. I was unsure about which link you want to provide... ;)

PR #686, opened 2015-11-18
New fixture - Showtec-Pixel-Bar-12

boxy321:
Fixture had only one mode and no physical data.

Original post:

http://www.qlcplus.org/forum/viewtopic.php?f=5&t=9203

Manual:

http://www.thomann.de/gb/showtec_pixel_bar_12.htm


mcallegari:
Fixture type "Other" ?? I would say this is a "Color changer".
Gonna fix it. Thanks

PR #687, opened 2015-11-19
Updated fixture - American-DJ-Nucleus-LED

boxy321:
Updated physical data and added click and go colours.


mcallegari:
I had to merge this manually.
Please do not rename mode names of existing fixtures cause you break backward compatibility !
Then users will get upset with me cause their projects got broken :(

PR #688, opened 2015-11-19
Updated fixture - American-DJ-Nucleus-PRO

boxy321:
Updated physucal data (only pan & tilt) and click and go colours.


mcallegari:
I'll revert the modes rename and the "PAn" typo

PR #694, opened 2015-11-23
Philips selecon pl profile1 mk ii

boxy321:
Corrected 16-bit colour LSB channels and other minor things. The full company name is clarified on page two of the datasheet:

http://www.qlcplus.org/forum/viewtopic.php?f=6&t=9219


mcallegari:
There are 2 fixtures in this PR. Is that intentional ?
In any case they look OK to me

boxy321:
Still getting the hang of rolling back a mistake in github without the world needing to know. I thought the first PR had not submitted so deleted and tried again...


Sent from my Samsung device

PR #698, opened 2015-11-24
Update catalan translation (UI and FixtureEditor)

sbenejam:
Update catalan translation (UI and FixtureEditor)


mcallegari:
Thanks Santiago :+1:

PR #699, opened 2015-11-25
VCSlider: ignore fade in and fade out times

plugz:
This way sliders in playback mode have the same behavior
as the sliders on a VCCueList.
Start a function with a slider -> manual fade in.
Stop a function with a slider -> manual fade out.

http://qlcplus.org/forum/viewtopic.php?f=12&t=9228

### DIFF
--- a/ui/src/virtualconsole/vcslider.cpp
+++ b/ui/src/virtualconsole/vcslider.cpp
@@ -1008,6 +1008,8 @@ void VCSlider::writeDMXPlayback(MasterTimer* timer, QList<Universe *> ua)
     {
         if (value == 0)
         {
+            // Make sure we ignore the fade out time
+            function->adjustAttribute(0, Function::Intensity);
             if (function->stopped() == false)
                 function->stop();
         }
@@ -1015,7 +1017,12 @@ void VCSlider::writeDMXPlayback(MasterTimer* timer, QList<Universe *> ua)
         {
             if (function->stopped() == true)
             {
-                function->start(timer);
+                // Since this function is started by a fader, its fade in time
+                // is decided by the fader movement.
+                function->start(
+                        timer, false, 0,
+                        0, Function::defaultSpeed(), Function::defaultSpeed()
+                        );
             }
             emit functionStarting(m_playbackFunction, pIntensity);
             function->adjustAttribute(pIntensity * intensity(), Function::Intensity);


mcallegari:
While this change is OK for me, my only concern is that it breaks the current behaviour.
There might be users with projects designed around the current behaviour and I guess they won't be happy about changing it.

Also, if a Scene attached to a slider in playback mode has a 2 minutes fade in, I might want to set the slider to 80% and wait for 2 minutes until the Scene gets to 80% intensity.
This might be a wanted behaviour.

In other words, I'm afraid this change should be a new option. On existing projects the behaviour should be the existing one (to preserve backward compatibility), while on newly created Sliders, the new behaviour should be the default.

As for the Cue List crossfaders behaviour, I don't remember who decided it (me ? you ?) but I suppose it should be consistent with the VCSlider default behaviour, thus "this change is OK for me" :)

What do you think ?

mcallegari:
On a second thought...do we really need this change ?
I mean, if you attach a function with no fade in/out the VCSlider already takes full control of the function intensity.

Can we just resolve this by placing a big bold message in the documentation like
** Note: if you want this slider to take full control over the attached Function intensity, then provide a Function with no Fade In and Fade Out times, so you will be able to manually perform fade in and fade out with the desired speed**

plugz:
The VCSlider + fadeout time issue has to be fixed anyway.
The current behavior for this is ugly. When you draw the slider down, the fadeout time is usually applied starting from a very low value.

I have no solution to correctly take fadeout time into account on a VCSlider, except ignoring it totally.

And since the fadeout time has to be ignored, the fadein time should be ignored too.
Bugs like "fade in works but fade out doesn't work" are sometimes reported, this would clear things out.

Of course, ignoring fadein time on VCSliders could be an opt-in option.
Ignoring fadeout time has to be done, because current behavior looks buggy.

mcallegari:
I totally agree in fixing the fade out issue.
As for ignoring or not the fade in, well to be honest I remain of the opinion that it could be easily solved without touching the code, but just by providing a Function with or without fade in.

In general I believe QLC+ is starting to have too many options and it's starting to be difficult to understand each one of them and the learning curve becomes steeper and steeper.
If we can avoid adding more options, that would be great.

In QLC+ 5 I'm planning to have an "Advanced" section of parameters everywhere it is needed, to separate edgy features from the basic ones, that should be easily accessed even without reading the documentation (that people don't read anyway... :laughing: )

plugz:
> I totally agree in fixing the fade out issue.

At least we agree on one thing :)

> As for ignoring or not the fade in, well to be honest I remain of the opinion that it could be easily solved without touching the code, but just by providing a Function with or without fade in.

This does not solve the fact that on a function with a fade in and a fade out time, controlling the function with a slider takes the fade in time into account, but not the fade out time.

It could also be solved by having the function controlled by a button, with a submaster on the side.

> In general I believe QLC+ is starting to have too many options and it's starting to be difficult to understand each one of them and the learning curve becomes steeper and steeper.
If we can avoid adding more options, that would be great.

> In QLC+ 5 I'm planning to have an "Advanced" section of parameters everywhere it is needed, to separate edgy features from the basic ones, that should be easily accessed even without reading the documentation (that people don't read anyway... :laughing: )

The current PR does not make this an option, I just said it could be made as an option :) IMO having fade in but no fade out is broken, I wouldn't even allow it. I don't know what others think, it would be nice to have some more feedback on the forum.

mcallegari:
> This does not solve the fact that on a function with a fade in and a fade out time, controlling the function > with a slider takes the fade in time into account, but not the fade out time.

Yes, and that's exactly what I meant: 
- fade out is always ignored
- if the Function has fade in, then see the case "2 minutes fade in & 80% intensity"
- if the function doesn't have fade in, then the slider will have full control of the intensity

I think it's just a matter of explaining it properly in the documentation

> The current PR does not make this an option, I just said it could be made as an option :) IMO having 
> fade in but no fade out is broken, I wouldn't even allow it.

Not necessarily (see above)

> I don't know what others think, it would be nice to have some more feedback on the forum.

Then we should post the proposal in the Issues & feedbacks forum, cause not everyone watch the development forum

mcallegari:
I moved your proposal in issues and feedbacks

mcallegari:
Let's merge this. 
I'll update the documentation accordingly

PR #701, opened 2015-11-27
Updated fixture - Stairville-MH-100-Beam-36x3-LED

boxy321:
The author has updated the definition with colours etc. See the post. Can't find an English version of the manual.

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9160

### DIFF
--- a/resources/fixtures/Stairville-MH-100-Beam-36x3-LED.qxf
+++ b/resources/fixtures/Stairville-MH-100-Beam-36x3-LED.qxf
@@ -3,98 +3,129 @@
 <FixtureDefinition xmlns="http://www.qlcplus.org/FixtureDefinition">
  <Creator>
   <Name>Q Light Controller Plus</Name>
-  <Version>4.7.5 GIT</Version>
-  <Author>Jasper Zevering</Author>
+  <Version>4.10.2 GIT</Version>
+  <Author>Antoni J. Canos</Author>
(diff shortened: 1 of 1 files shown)


mcallegari:
Not OK.
As I mentioned before, please do not rename mode names on existing fixtures. It breaks the backward compatilibty !
```
- <Mode Name="14-Ch Mode">
+ <Mode Name="14 Channels">
```

also, I don't quite like the shortening of Pan/Tilit channels like:
```
-  <Capability Min="0" Max="255">Pan (0° - 630°)</Capability>
+  <Capability Min="0" Max="255">0°-630°</Capability>
```
or
```
-  <Capability Min="0" Max="255">Pan fine</Capability>
+  <Capability Min="0" Max="255">Fine</Capability>
```

boxy321:
Whoops! I didn't change the mode names, the submitter did, but I should have noticed...

The removal of duplicate descriptions is a bit like the 'mode' word being implicit. You have a capability called 'Pan', in a channel called 'Pan', which is in a channel group called 'Pan', which has a pan icon.

On the other hand, I agree that other channel types can look strange if shortened to a single number, so macro 1, macro 2... won't be shortened to 1, 2... even if the channel is called Macros. I noticed that if the name is removed, the only clue to the function is by hovering the mouse and waiting for a tool-tip to appear.

PR #711, opened 2015-12-04
Update german translation. I'm not absolutely sure about the one with…

kripton:
… two newlines ...


mcallegari:
Looks OK to me, thanks.
When there are multiple lines, just copy-paste the original string and translate only the english part, keeping the original "layout"

PR #714, opened 2015-12-08
ArtNet plugin rework

plugz:
- Each controller binds on its own IP address instead of ANY address, allowing to send ArtNet to several different networks

- ArtPoll is sent every 5 seconds so lost packets are less of an issue

- Configuration: can select any input ArtNet universe

This has been tested on windows & linux, not on OSX.

### DIFF
--- a/plugins/E1.31/e131controller.cpp
+++ b/plugins/E1.31/e131controller.cpp
@@ -25,18 +25,18 @@
 #define TRANSMIT_FULL    "Full"
 #define TRANSMIT_PARTIAL "Partial"
 
-E131Controller::E131Controller(QNetworkInterface const& interface, QString const& ipaddr,
-                               Type type, quint32 line, QObject *parent)
+E131Controller::E131Controller(QNetworkInterface interface, QNetworkAddressEntry address,
+                               quint32 line, QObject *parent)
     : QObject(parent)
     , m_interface(interface)
-    , m_ipAddr(ipaddr)
+    , m_ipAddr(address.ip())
     , m_packetSent(0)
     , m_packetReceived(0)
     , m_line(line)
     , m_UdpSocket(new QUdpSocket(this))
     , m_packetizer(new E131Packetizer())
 {
-    qDebug() << "[E131Controller] type: " << type;
+    qDebug() << Q_FUNC_INFO;
     m_UdpSocket->bind(m_ipAddr, 0);
     // Output multicast on the correct interface
     m_UdpSocket->setMulticastInterface(m_interface);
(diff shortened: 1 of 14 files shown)


mcallegari:
Shall we discuss this a bit please ?

>> Each controller binds on its own IP address instead of ANY address, allowing to send ArtNet to several different networks

I'm not sure if this change is going to work in all cases.
The 2 nasty ones are: loopback and input/output on the same IP.
There are quite a few interesting usages where 2 ArtNet softwares communicate on the same PC. (e.g. 3D Visualizers, projection mapping, etc..) 
Sometimes a software doesn't support the loopback interface, so the only option is sending/receiving on the same PC IP address.
Have you tested this usage too ?

>> ArtPoll is sent every 5 seconds so lost packets are less of an issue

Why is this needed ? Have you eperienced packet loss in some scenario ? If so, which one ?

>> Configuration: can select any input ArtNet universe

Nice one

plugz:
> Shall we discuss this a bit please ?

Yes, of course :)

> The 2 nasty ones are: loopback and input/output on the same IP.

These 2 have been tested and work on both windows and linux

> Sometimes a software doesn't support the loopback interface, so the only option is sending/receiving on the same PC IP address.

The issue with this is that usually, to output ArtNet, a software will bind to the ArtNet port, disallowing another software to bind to it.
So with or without this PR, the issue is present.
QLC+ can't bind on 0:6454 (before PR) and QLC+ can't bind on <interface_address>:6454 (after PR).



> > ArtPoll is sent every 5 seconds so lost packets are less of an issue

> Why is this needed ? Have you eperienced packet loss in some scenario ? If so, which one ?

- ArtNet protocol specification says the ArtPoll packet should be sent every few seconds so the list can be refreshed

- Yes

- Wifi



> > Configuration: can select any input ArtNet universe

> Nice one

Nice

plugins/E1.31/e131controller.cpp
```diff
@@ -25,18 +25,18 @@
 #define TRANSMIT_FULL    "Full"
 #define TRANSMIT_PARTIAL "Partial"
 
-E131Controller::E131Controller(QNetworkInterface const& interface, QString const& ipaddr,
-                               Type type, quint32 line, QObject *parent)
+E131Controller::E131Controller(QNetworkInterface interface, QNetworkAddressEntry address,
```
janosvitok: Don't you want to keep the "const & interface" to avoid temporary object creation? The same for address?

plugz: Yes

mcallegari:
OK, I need some time to test this patch with all the cases I know (and on the RPi as well)

If you agree, I would postpone the merge after I release 4.10.2, which I would like to do on sunday (13th)

plugz:
Sure :)

mcallegari:
Alright, I haven't found the time to test this, so I'm gonna merge and test it later when I can.

mcallegari:
@plugz anything to mention in the changelog ?

plugz:
@mcallegari Will do

PR #721, opened 2015-12-15
New fixture - Chauvet-Rogue-RH1-Hybrid

boxy321:
http://www.chauvetprofessional.eu/products/rogue-rh1-hybrid/


mcallegari:
Awesome, thanks !

Just one minor thing. When Pan and Tilt are present, the Focus type should be "Head" and not "Fixed":
```
<Focus Type="Fixed" PanMax="540" TiltMax="260"/>
```

PR #730, opened 2015-12-18
Force universe ID and Index to be the same

plugz:
Forbid creating gaps in the universe list
by deleting a universe that is not the last one.

When loading a workspace that contains gaps in
the universe list, fill the gaps.

This fixes various issues and crashes that happen
when a gap is present in the universe list.

Related thread in the forum: http://qlcplus.org/forum/viewtopic.php?f=5&t=9303
This is open for discussion.

After struggling with the 'only use universe ID' solution (that touches a lot of files and can be seen here https://github.com/plugz/qlcplus/tree/universeindex-fix-2), and realising that the qmlui code uses universes indexes everywhere, and that I don't know shit about qml... I gave up and came up with a way more simple solution to this issue:
Force universeID==universeIndex

Pros of this solution: no more id/index confusion.
Bad with this solution: it forbids deleting a universe in the middle of the universe list.

### DIFF
--- a/engine/src/inputoutputmap.cpp
+++ b/engine/src/inputoutputmap.cpp
@@ -42,7 +42,6 @@
 InputOutputMap::InputOutputMap(Doc *doc, quint32 universes)
   : QObject(doc)
   , m_blackout(false)
-  , m_latestUniverseId(InputOutputMap::invalidUniverse())
   , m_universeChanged(false)
 {
     m_grandMaster = new GrandMaster(this);
@@ -127,15 +126,31 @@ quint32 InputOutputMap::invalidUniverse()
 
 bool InputOutputMap::addUniverse(quint32 id)
 {
-    QMutexLocker locker(&m_universeMutex);
-    if (id == InputOutputMap::invalidUniverse())
-        id = ++m_latestUniverseId;
-    else if (m_latestUniverseId == InputOutputMap::invalidUniverse()
-            || id >= m_latestUniverseId)
-        m_latestUniverseId = id;
+    {
+        QMutexLocker locker(&m_universeMutex);
+        if (id == InputOutputMap::invalidUniverse())
+            id = universesCount();
+        else if (id < universesCount())
+        {
+            qWarning() << Q_FUNC_INFO
+                << "Universe" << id << "is already present in the list."
+                << "The universe list may be unsorted.";
+            return false;
+        }
+        else if (id > universesCount())
+        {
+            qDebug() << Q_FUNC_INFO
+                << "Gap between universe" << (universesCount() - 1)
+                << "and universe" << id
+                << ", filling the gap...";
+            while (id > universesCount())
+            {
+                m_universeArray.append(new Universe(universesCount(), m_grandMaster));
+            }
+        }
 
-    m_universeArray.append(new Universe(id, m_grandMaster));
-    locker.unlock();
+        m_universeArray.append(new Universe(id, m_grandMaster));
+    }
 
     emit universeAdded(id);
     return true;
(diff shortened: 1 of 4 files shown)


plugz:
Forgot unit tests, will fix soon...

mcallegari:
Fair enough.
I'd suggest just one change: instead of allowing to delete any universe and then display a popup message if it's not the last in the list, why don't you disable the remove button if the selected universe is not the last one ?
Then we can add a note in the documentation explaining this behaviour.

> the qmlui code uses universes indexes everywhere, and that I don't know shit about qml... 

Lol ! Don't worry about the qmlui, it's still a huge work in progress. If I need to use IDs, then I'll adapt the code. Just tell me which solution I should go for.
As for shit :), I found a very interesting reading to learn QML: https://qmlbook.github.io/
QML itself is pretty simple. The headaches start when you have to interact with complex C++ structures. So far, it seems to me QVariant is the key :)

plugz:
> If I need to use IDs, then I'll adapt the code. Just tell me which solution I should go for.

As you wish: Do you want the easy code, or do you want to not lose a feature ?
I would go for the IDs-everywhere solution, it's better from a user point-of-view, but you have to know that this implies changes in the qml code that I cannot do myself. For both solutions, the code is ready for the current ui :)

> I'd suggest just one change: instead of allowing to delete any universe and then display a popup message if it's not the last in the list, why don't you disable the remove button if the selected universe is not the last one ?

Yes sure, I we decide to keep this solution.

> I found a very interesting reading to learn QML: https://qmlbook.github.io/

Interesting, thanks :)

mcallegari:
> As you wish: Do you want the easy code, or do you want to not lose a feature ?

IDs equals to indices is OK for me. In the QML UI I have to review the universe selection anyway.
Plus I haven't implemented add/remove a universe yet.
If you make the popup change, then we merge and I'll adapt qmlui to the new code if needed

plugz:
Popup change: done.

> IDs equals to indices is OK for me. In the QML UI I have to review the universe selection anyway.
Plus I haven't implemented add/remove a universe yet.

Well with this change you don't have to change anything in the qmlui :)

mcallegari:
Can you please also remove this ?
https://github.com/plugz/qlcplus/commit/a05495f5f959ef665bc70ec82dbd31aa5caaf8fa#diff-27f1c32c7d76fec06b8153b777e74253R350

Basically they are more strings to translate for translators and if it never happens then it's kinda useless now.

plugz:
Done

janosvitok:
I think that we can expect that a user will want to drop any universe (not just the last one). E.g. remove stuff related to a RGB panel, or to MIDI controller. If only last one can be deleted, the user has to delete stuff from the universe in the middle, move stuff from the next universe, repeat until the last one and then remove the last universe. If we want to keep Index== ID, this should be added sooner or later.

plugz:
Well if we want to be able to drop universes in the middle of the list, I have to create a PR for this branch: https://github.com/plugz/qlcplus/tree/universeindex-fix-2

As I said the code is ready for both methods, it's a matter of deciding which one to go for.

plugz:
After some more tests with the "clean" way of doing things, there is still a lot of work to be done for it to work.

The implementations of Function::write() and DMXSource::writeDMX() are all based on universeID==universeIndex.

The best for now would be to merge this PR, and only allow deleting the last universe.

> I think that we can expect that a user will want to drop any universe (not just the last one). E.g. remove stuff related to a RGB panel, or to MIDI controller. If only last one can be deleted, the user has to delete stuff from the universe in the middle, move stuff from the next universe, repeat until the last one and then remove the last universe. If we want to keep Index== ID, this should be added sooner or later.

The thing is, for now, deleting another universe than the last one will create a lot of issues. It's like a "break workspace" button.

mcallegari:
Let's merge this !

PR #734, opened 2015-12-21
(FIX): add a variable for the udev rules path

daveol:
Add the UDEVRULESDIR variable to let packagers change the location of the udev rules directory in one place. Because some distributions of linux don't use "/etc/udev/rules.d" for the rules but something like "/usr/lib/udev/rules.d"


mcallegari:
Makes perfect sense. Thank you very much !

PR #735, opened 2015-12-21
Fix translate

daveol:
This fix
- removes the "-silent" parameter from the call to lrelease since that is not a real parameter
- makes use of bash variables to specify the names of files in translate.sh
- makes the translate files install by putting "./" in front of the file names

### DIFF
--- a/qlc.pro
+++ b/qlc.pro
@@ -45,9 +45,9 @@ win32:coverage.commands = @echo Get a better OS.
 translations.target = translate
 QMAKE_EXTRA_TARGETS += translations
 translations.commands += ./translate.sh
-translations.files += qlcplus_de_DE.qm qlcplus_es_ES.qm qlcplus_fr_FR.qm
-translations.files += qlcplus_it_IT.qm qlcplus_nl_NL.qm qlcplus_cz_CZ.qm
-translations.files += qlcplus_pt_BR.qm qlcplus_ca_ES.qm qlcplus_ja_JP.qm
+translations.files = ./qlcplus_de_DE.qm ./qlcplus_es_ES.qm ./qlcplus_fr_FR.qm
+translations.files += ./qlcplus_it_IT.qm ./qlcplus_nl_NL.qm ./qlcplus_cz_CZ.qm
+translations.files += ./qlcplus_pt_BR.qm ./qlcplus_ca_ES.qm ./qlcplus_ja_JP.qm
 translations.path   = $$INSTALLROOT/$$TRANSLATIONDIR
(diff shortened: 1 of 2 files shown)


mcallegari:
-silent is a valid option of lrelease Qt4
```
    -silent
           Do not explain what is being done
```

daveol:
yes, but not on qt5 it doesn't recognize the option and doesn't make the file....

mcallegari:
That's weird.
There's plenty of Qt5 builds on OBS and we never encountered such issue:
https://build.opensuse.org/project/show/home:mcallegari79

mcallegari:
On Ubuntu:
```
/usr/lib/x86_64-linux-gnu/qt5/bin/lrelease -version
lrelease version 5.2.1
```
```
/usr/lib/x86_64-linux-gnu/qt5/bin/lrelease -h
Usage:
    lrelease [options] project-file
    lrelease [options] ts-files [-qm qm-file]

lrelease is part of Qt's Linguist tool chain. It can be used as a
stand-alone tool to convert XML-based translations files in the TS
format into the 'compiled' QM format used by QTranslator objects.

Options:
    -help  Display this information and exit
    -idbased
           Use IDs instead of source strings for message keying
    -compress
           Compress the QM files
    -nounfinished
           Do not include unfinished translations
    -removeidentical
           If the translated text is the same as
           the source text, do not include the message
    -markuntranslated <prefix>
           If a message has no real translation, use the source text
           prefixed with the given string instead
    -silent
           Do not explain what is being done
    -version
           Display the version of lrelease and exit
```

daveol:
hmm.... I also see a ``` -silent ``` in lrelease's help page, but in the script it just fails except when I remove the ``` -silent ```

janosvitok:
This is lrelease HEAD source: http://code.qt.io/cgit/qt/qttools.git/tree/src/linguist/lrelease

It does contain -silent; it unsets m_verbose. Funny thing is that -verbose is not documented.

http://koji.fedoraproject.org/koji/buildinfo?buildID=706809 contains -silent.

mcallegari:
@daveol  any updated on this ? Have you managed to figure out why it fails on your end ?

daveol:
apparently i had qt3.3 still installed and the package defined QTDIR in my environment

daveol:
Any reason to keep this out?

mcallegari:
Not really, but I don't understand why this is necessary.
The translate script already works right now

daveol:
It makes things clearer and prevents possible problems in the future. I don't care if you don't want to merge it but then just close the pull request

PR #739, opened 2015-12-24
Add appdata files and install them

daveol:
Appdata files get used by software-center on ubuntu, gnome-software-centre on fedora and other software centers. This also gives room for seperate icons and info for plugins if we implement a ".metainfo.xml" file for them. The plugins most likely also get listed underneath the package they extend.

For more info on appdata go [here](http://www.freedesktop.org/software/appstream/docs/chap-Metadata.html) and for more info on metainfo go [here](http://www.freedesktop.org/software/appstream/docs/sect-Quickstart-Addons.html).

### DIFF
--- a/etc/etc.pro
+++ b/etc/etc.pro
@@ -17,6 +17,10 @@ unix:!macx {
     mime.files += qlcplus.xml
     INSTALLS   += mime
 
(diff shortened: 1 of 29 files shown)


mcallegari:
I'll merge this as it is, but I'll push some changes afterward as there are a few things that I don't like

PR #740, opened 2015-12-29
Fix language declaration of italian translation

kripton:
Just saw that there were two "nl" entries. German translation will follow ...


mcallegari:
Ouch ! Thanks

PR #742, opened 2015-12-29
added input profile for behringer bcr2000

xarumanx:
using knobs as input type


mcallegari:
Excellent, thank you very much !

PR #745, opened 2015-12-30
Save in temp file before replacing destination file

plugz:
This way, in case of a crash during saving,
the current destination file will not become blank or corrupt.

Idea here:
http://qlcplus.org/forum/viewtopic.php?f=5&t=9227&p=40548&hilit=crash+while+saving#p40548

### DIFF
--- a/engine/src/qlcfixturedef.cpp
+++ b/engine/src/qlcfixturedef.cpp
@@ -301,7 +301,9 @@ QFile::FileError QLCFixtureDef::saveXML(const QString& fileName)
     if (fileName.isEmpty() == true)
         return QFile::OpenError;
 
-    QFile file(fileName);
+    QString tempFileName(fileName);
+    tempFileName += ".temp";
+    QFile file(tempFileName);
     if (file.open(QIODevice::WriteOnly) == false)
         return file.error();
 
@@ -329,6 +331,19 @@ QFile::FileError QLCFixtureDef::saveXML(const QString& fileName)
     error = QFile::NoError;
     doc.writeEndDocument();
     file.close();
+
+    // Save to actual requested file name
+    QFile currFile(fileName);
+    if (currFile.exists() && !currFile.remove())
+    {
+        qWarning() << "Could not erase" << fileName;
+        return currFile.error();
+    }
+    if (!file.rename(fileName))
+    {
+        qWarning() << "Could not rename" << tempFileName << "to" << fileName;
+        return file.error();
+    }
 
     return error;
 }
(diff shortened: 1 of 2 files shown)


mcallegari:
Neat idea, but it would be nice to understand what happened to the user that got a crash during the save process. Actually that was the first case I saw in 3 years.
Usually it happens the other way around: QLC+ crashes when loading a project.

A more generic approach would be like what QtCreator does: as long as you edit a file (and don't save it) the program creates a .autosave file for you.
If the program explodes or the power goes down, that's your backup point.

I don't know the save policy they adopt. could be based on the number of changes or time-based. Don't know. Surely we cannot afford to hook such feature to the Doc::setModified method.

Ideas ?

plugz:
> Neat idea, but it would be nice to understand what happened to the user that got a crash during the save process. Actually that was the first case I saw in 3 years.

It never happened to me either, but it could easily happen: When saving, the whole QLC+ instance is read through. 1 single issue may cause a crash.

I think both ideas are complementary.

It would be nice to have an autosave feature.
It would also be nice to no break the target file if QLC+ crashes while saving.

mcallegari:
I gave a second thought to this and I believe the autosave can cover both cases.

- When Doc::setModified is called, a timer is started, with one minute timeout, recurring every minute.
After one minute a .autosave file is created. If there are save issues and QLC+ crashes, you will loose one minute of work maximum.
- If more setModified are called and the timer is already on, no action is taken.
- If the user manually saves the project, the timer is killed and the autosave file is removed.

How does this sound ?

janosvitok:
I agree with david that those two are quite independent. For saving, it's safer to create the new file first and destroy the original only after the new one is created completely.

With this approach the recovery is to delete excess files, while with .autosave, you have to rename the proper file to proper name.

Problems I see is that it will not keep the created timestamp (but that can be handled separately).

That said, I agree that auto save is useful as well.

janosvitok:
The code in this commit seems like repeating the same pattern. Itmay be worth to create helper method (e.g. in qlcfile).

mcallegari:
OK, I think I see your point now, but can we decide a unified "extra" extension for both cases ?
By ".autosave" I meant "project_name.qxw.autosave".

Proposals: 
.bak
.temp
.autosave

plugz:
> OK, I think I see your point now, but can we decide a unified "extra" extension for both cases ?

So a normal saving would erase the autosave file right away ?

mcallegari:
A normal save would save on .autosave and then rename it to the original project name.

plugz:
The saving processes (autosave, manual save) should both use a temp file. We don't want a crash to create an invalid file.

This means having 2 extensions:
.temp extension. This type of file should never be seen by a user, it only exists if QLC+ crashes while saving. This is an invalid workspace file.
.autosave extention. This type of file can be used to recover work if QLC+ crashed while doing whatever. This is a valid workspace file.

kripton:
Just a hint: "projectname.autosave.qxw" might be better than "projectname.qxw.autosave" since it still allows displaying the file icon and opening the workspace more easily.

PR #748, opened 2015-12-30
Licensechecker

daveol:
This Pull Request adds a script that prints out files with UNKNOWN licenses and also fixes some files


mcallegari:
Adding licenses to .ui files is not a good idea for 2 reasons:
1- the Qt Company itself doesn't do it on released examples. See this for instance:
http://code.qt.io/cgit/qt/qtbase.git/tree/examples/widgets/painting/fontsampler/mainwindowbase.ui

2- UI files are normally edited with Qt Creator, which will probably delete the license header once we save the file

daveol:
Well I would like there to be a license header in the ui files, the way Qtractor does it seems like a good idea

mcallegari:
Alright. Is there a way to automate the change on each .ui file ?
I don't really have the time to do it manually or write a script myself :cry:

daveol:
I'll just do it myself :smile:

check-licenses.sh
```diff
@@ -0,0 +1,71 @@
...
+
+##config
+#file extentions
+FILE_EXT[0]="cpp"
+FILE_EXT[1]="c"
+FILE_EXT[2]="h"
+FILE_EXT[3]="js"
+FILE_EXT[4]="qml"
+FILE_EXT[5]="ui"
+#FILE_EXT[6]="xml"
```
daveol: We track files with the following extensions:
- cpp
- c
- h
- js
- qml
- ui

Are there any file extensions that I have missed? And are you happy with the ones I chose?

daveol:
I think this is ready. Thoughts @mcallegari, @plugz  ?

mcallegari:
Looks OK to me. I need to double check if it makes sense to add a license header to the Android manifest, but in case I can remove it later.

Thanks !

peternewman:
Would it be worth adding this to Travis CI @daveol ?

PR #754, opened 2016-01-05
New fixture - JB-Lighting-JBLED-A7

boxy321:
Here's a translation of the definition. There is an English version at the last part of the manual here:

Google: jb lighting a7 manual


kripton:
For reference: http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9335

PR #757, opened 2016-01-12
fix crash: select another input while rgbaudio is running

plugz:
I made a little test for audio input stability:

- Create an RGB matrix in audio mode
- Create a chaser with just the matrix in it, with duration=0
- play the chaser
- go to the I/O tab, and mess with the audio input configuration

This makes QLC+ crash quite quickly.
With this patch, QLC+ does not crash.

What it does:
- AudioCapture::uninitialize() does the Audio uninitialization (instead of doing it in the destructor)
- AudioCapture::initialize() and AudioCapture::uninintialize() are called in the AudioCapture Thread. This way Qt objects (QAudioInput) are created and destroyed from the same thread. Qt likes this better.
- This allow AudioCapture::run() to know when initialization could not be achieved, and abandon reading.

Tested on Linux (Qt5 version, Qt4 version)
I will test on Windows (Qt5) soon. I can't test the windows Qt4 version. Is this still used ?
I can't test on OSX at all. Is the OSX Qt4 version still used ?

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,5 +1,7 @@
 qlcplus (4.10.3) stable; urgency=low
 
+  * Audio Capture: Fix crash when selecting another audio input while a capture is running (David Garyga)
+  * Audio Capture: Fix crash when trying to use a wrongly configured audio input (David Garyga)
   * Scene Editor: Remember Channels Groups values when saving and loading a workspace (David Garyga)
   * Scene Editor: Remember fixtures even with no activated channel (David Garyga)
   * Show Manager: Fix crash when editing the total time of an empty chaser (David Garyga)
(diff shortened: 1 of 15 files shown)


mcallegari:
> I will test on Windows (Qt5) soon. I can't test the windows Qt4 version. Is this still used ?
I can't test on OSX at all. Is the OSX Qt4 version still used ?

Windows and OSX use Qt5 now. Qt4 is kept only on Linux cause we don't know on which distro QLC+ will run

mcallegari:
Please update the changelog for this and I'll merge

plugz:
> Windows and OSX use Qt 5.5.1 now. Qt4 is kept only on Linux cause we don't know on which distro QLC+ will run

Then can I remove the audio captures for qt4 on windows and osx ?

mcallegari:
> Then can I remove the audio captures for qt4 on windows and osx ?

The calls under ifdef, yes, the actual cpp/h files no (not yet), please

plugz:
Changing AudioCapture forced me to update AudioCapturePortAudio and AudioCaptureWaveIn, even though they are not in use anymore.

It's mostly for readability.



If I don't remove the files, then I won't remove the ifdefs either.

Issue #705, opened 2015-12-02
Bug in chaser when going from a show to the same scene as the end of the show


anorod:
Hi again:

I have a few shows working on a chaser, that when the show finish I have a scene at the end that is the same as the next scene of the chaser. This is because after the show I want to get back in control of the cues so I can continue hitting "next" or fading as usual.

What I saw is that when I use the same scene after the show that the one I use at the end of the show inside the show, I have a blackout. Eg:

I have scenes 5 scenes with different channels on. I have a show that goes 2->3->4 automatically with some timing, but scene number 4 is infinite duration. If I make a chaser that goes: 1->show->4->5, when I hit next after the show it should not make any change, but instead it makes a blackout.

Probably this is because at the same time the system is fading up 4 and fading it down.

Hope it helps too :)

plugz:
Hello,

Can you post a workspace file demonstrating the issue ?
You can post it in a forum thread there : http://qlcplus.org/forum/ and link the thread here.

Thank you

anorod:
Hi!,
Here is the example:
 - http://qlcplus.org/forum/viewtopic.php?f=5&t=9492
Thanks

plugz:
Fixed by #787

PR #775, opened 2016-01-27
Function Start/Stop source

plugz:
This is a huge engine change.

Basically put: A function keeps a list of 'what' started it. 'What' is called a 'function source' (but the name may be changed)

As long as at least one 'function source' is present in a function's list, the function will not be stopped.

As-is, this PR break qmlui build. I'll fix it before merging the PR, I just want to discuss the PR before.


plugz:
This fixes #705

mcallegari:
Uhm...why does this conflict ? Did I cause it ?

Can we rename "Source" into "FunctionSource" or "ParentSource" or "Parent" ?

Can you add some comments that explain the meaning of the SourceType enums ? (for example, what's the difference between AutoVCWidget and ManualVCWidget ?)
And who is this God after all ? :laughing: 
Can we call it "Master" or "MasterOfTheUniverse" or simply "He-Man" ? :laughing: 

What's the difference between God and Mask ?
Or is it a single entity called [Mixcoatl](https://en.wikipedia.org/wiki/Mixcoatl), the masked god ? :)

plugz:
> Uhm...why does this conflict ? Did I cause it ?

I probably caused this, don't worry

> Can we rename "Source" into "FunctionSource" or "ParentSource" or "Parent" ?

Yes, it could even go into it's own .h and .cpp files.

> Can you add some comments that explain the meaning of the SourceType enums ? (for example, what's the difference between AutoVCWidget and ManualVCWidget ?)

Will do


> And who is this God after all ? :laughing:

I don't know I never met him :(
But he has the power to stop running functions !

plugz:
continued as #787

PR #787, opened 2016-02-02
Function startstopsource mastertimerloop 2

plugz:
Rebase of #775 to fix merge issues

This is a huge engine change.

Basically put: A function keeps a list of 'what' started it. 'What' is called a 'function source' (but the name may be changed)

As long as at least one 'function source' is present in a function's list, the function will not be stopped.

As-is, this PR break qmlui build. I'll fix it before merging the PR, I just want to discuss the PR before.

> Can we rename "Source" into "FunctionSource" or "ParentSource" or "Parent" ?

Yes, it could even go into it's own .h and .cpp files.

> Can you add some comments that explain the meaning of the SourceType enums ? (for example, what's the difference between AutoVCWidget and ManualVCWidget ?)

Will do


plugz:
> What's the difference between God and Mask ?

Mask was not used and has been removed.

> Or is it a single entity called Mixcoatl, the masked god ? :joy:

Maybe... can this god light things up ? :D

plugz:
> Can you add some comments that explain the meaning of the SourceType enums ? (for example, what's the difference between AutoVCWidget and ManualVCWidget ?)

Some comments are present too if you look at the global changes, but I'll add some.

plugz:
- Renamed to FunctionParent
- Some comments have been added

PR #796, opened 2016-02-15
fix audiocapture_qt

### LINKED ISSUE #795: Audio input changes regression
As reported by @sbenejam here http://www.qlcplus.org/forum/viewtopic.php?f=12&t=9556&p=41999

@plugz: I went back in the GIT history and found that PR #757 caused it.

Basically, in the current code, this call  https://github.com/mcallegari/qlcplus/blob/master/engine/audio/src/audiocapture_qt.cpp#L132 always returns 0 bytes available. Commenting it seems to make the capture working.
However, even the call https://github.com/mcallegari/qlcplus/blob/master/engine/audio/src/audiocapture_qt.cpp#L135 always returns that 0 bytes have been read.

I've seen this kind of things before, and the explanation I gave myself is that QtMultimedia is crap.
It is VERY sensitive to threads and if you create an instance in one thread, you cannot use it in another thread, OR, it shows weird behaviour like the above.

Needless to say that this has to be fixed before 4.10.3.

plugz:
fix #795 

The issue is that neither QAudioInput::bytesReady()
nor QIODevice::bytesAvailable() are reliable here.

The only solution is to simply call QIODevice::read()
and use its return value.

### DIFF
--- a/engine/audio/src/audiocapture.h
+++ b/engine/audio/src/audiocapture.h
@@ -137,7 +137,7 @@ signals:
 
 protected:
     /*!
-     * Reads up to \b maxSize bytes from \b the input interface device.
+     * Reads up to \b maxSize uint16 from \b the input interface device.
      * Returns an array of the bytes read, or an empty array if an error occurred.
      * Subclass should reimplement this function.
      */
--- a/engine/audio/src/audiocapture_qt.cpp
+++ b/engine/audio/src/audiocapture_qt.cpp
@@ -92,6 +92,8 @@ bool AudioCaptureQt::initialize()
         return false;
     }
 
+    m_currentBufferPosition = 0;
+
     return true;
 }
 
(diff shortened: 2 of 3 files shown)


mcallegari:
> The only solution is to simply call QIODevice::read() and use its return value.

That's really weird. Apparently even the read function doesn't return the correct number of bytes read.
I'm using Qt 5.5.1, which Qt version are you using ?
By the way, QIODevice::read returns qint64: http://doc.qt.io/qt-5/qiodevice.html#read

I'm gonna test this later tonight, but for sure, before #757 the code was working, and I would like to understand what really happenend that caused the code to not work anymore.

As I mentioned in #795, I bet on some wrong thread handling, even though checking the changes of #757 didn't seem to be the case :disappointed:

plugz:
> Apparently even the read function doesn't return the correct number of bytes read.

Are you sure ? Why ?

> I'm using Qt 5.5.1, which Qt version are you using ?

I tested this with Qt 5.5.1 too.

> By the way, QIODevice::read returns qint64: http://doc.qt.io/qt-5/qiodevice.html#read

We don't try to read more than INT_MAX bytes anyway, but I can change the type of read_bytes


> I'm gonna test this later tonight, but for sure, before #757 the code was working, and I would like to understand what really happenend that caused the code not to work anymore.

Maybe the audio thread was slower in #757 because it was tied to the main thread somehow.
Really I don't know, I'll have to look at the QAudioInput source code.

The main difference is:

Before #757: QAudioInput is created and started from the UI thread OR the Engine thread.
Then it is stopped and destroyed from the UI thread OR the Engine thread.
It may be stopped from a different thread that it was started for, this causes a crash.

After #757: QAudioInput is created, started, stopped and destroyed from the same thread: the AudioCapture thread.

mcallegari:
> Apparently even the read function doesn't return the correct number of bytes read.
> Are you sure ? Why ?

I am sure. I saw it when I opened the issue report. Maybe it's just qDebug not able to print qint64...but I saw always 0 bytes returned.

> After #757: QAudioInput is created, started, stopped and destroyed from the same thread: the AudioCapture thread.

Makes perfect sense to me and in theory that is the way to go.
However there is still the variable QtMultimedia == CRAP that fucks everything up.

PR #816, opened 2016-03-06
German

kripton:
Update of german translations, including the long-overdue *.appdata.xml-files.

@mcallegari, if there are any more files that need to be translated that I forgot, please tell me :)


mcallegari:
Right on time for the release :laughing: 
Thanks !

PR #821, opened 2016-03-08
Vdpdmxpc12   12 channel switch pack

boxy321:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9646

### DIFF
--- a/resources/fixtures/HQ-Power-VDPDMXPC12---12-Channel-Switch-Pack.qxf
+++ b/resources/fixtures/HQ-Power-VDPDMXPC12---12-Channel-Switch-Pack.qxf
@@ -1,0 +1,129 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE FixtureDefinition>
+<FixtureDefinition xmlns="http://www.qlcplus.org/FixtureDefinition">
+ <Creator>
+  <Name>Q Light Controller Plus</Name>
+  <Version>4.10.2 GIT</Version>
+  <Author>DJ MoreBass</Author>
+ </Creator>
+ <Manufacturer>HQ Power</Manufacturer>
+ <Model>VDPDMXPC12 - 12-Channel Switch Pack</Model>
+ <Type>Other</Type>
+ <Channel Name="Channel 1">
+  <Group Byte="0">Intensity</Group>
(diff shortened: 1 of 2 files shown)


kripton:
boxy321, thanks for all fixtures you contributed in the past months. Amazing work!

However IMHO, we shouldn't have fixture definitions for dimmers. Of course they add information like power consumption, type of DMX connector and dimensions but I prefer using the "Generic Dimmer" fixture.
About switchpacks, I'm unsure ...

boxy321:
Interesting point. They're pretty boring things. If there are thousands of models of these about (I suspect there are), should something be put in the docs to explain to use generic dimmer if it's not already?

kripton:
Yes, Generic Dimmers are available in plethora of models. They are often used multiple times (okay, other fixtures are as well) and I usually add one Generic fixture with 18 channels if there are three dimmer packs with 6 channels each. It's not mentioned explicitly in the docs. @mcallegari any opinion if more documentation is needed?

boxy321:
I agree. A lot of these parts are made by the dreaded Velleman under made-up brand names. Feel free to remove them. I am often in two minds when a new manufacturer appears and judge whether to add them 'properly' based on the user manual quality half the time.

mcallegari:
My personal opinion is that dimmer packs should use the QLC+ generic dimmer built in functionality.
I believe also that fixtures with the same channels but different physical properties should be a single definition with 2 modes, eventually with the product code.

On the other hand, it's easier for users to find a ready-made definition. As far as I understand, some of them are scared by fixture definition creation. Don't know why. Maybe they are just lazy and don't even try sometimes. In fact, the "Fixtures requests" section of the forum has become quite popular...

Also, there is a technical reason. As we're rapdily approaching a 1000 definitions, thanks to Robert, each new def is a new line added to the FixturesMap.xml file, and that file has to be completely loaded and parsed when starts up QLC+. The bigger it is, the slower QLC+ will start up.
On a PC it takes a few milliseconds, but (for example) on a Raspberry Pi everything is amplified and much slower, so each millisecond saved is gold.

In conclusion, I would say we can live without dimmer packs definitions.

mcallegari:
@boxy321 so...what do we do with dimmer packs ?

PR #824, opened 2016-03-09
New fixture - Stairville-MH-X60th-LED-Spot

boxy321:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9676

PR #826, opened 2016-03-10
Fixture update: Original author found better gobo image for gobo 2.

boxy321:
I looked through the gobos 15 times but could not see 'Clay Paky/gobo0067.png'.


mcallegari:
>> Fixture update: Original author found better gobo image for gobo 2

Why not including the change in #824 then ?

>> I looked through the gobos 15 times but could not see 'Clay Paky/gobo0067.png'.

Sorry, I don't get this. Gobo 67 is here:
https://github.com/mcallegari/qlcplus/blob/master/resources/gobos/ClayPaky/gobo00067.png

boxy321:
The author found the black version after I pushed so I updated it. I was being blind.

PR #831, opened 2016-03-17
New fixture - Cameo-CLHB400RGBW

boxy321:
I have no idea what the OP was thinking here. The 12:20pm post is complete jibberish.

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=9715


mcallegari:
> The 12:20pm post is complete jibberish.

Yeah...italian people...what do you expect ? :wink:

PR #834, opened 2016-03-25
osc rework: fix receiving packets on wrong interface

### LINKED ISSUE #832: OSC plugin stops sending signals when setting an input profile
Sequence:
- select OSC input line
- select OSC feedback line
- check that it emits signals (joystick blinking)
- set an input profile
- no more signals sent to QLC+ (joystick not blinking anymore)

The plugin is actually working and receiving data, but it doesn't reach the QLC+ UI

Doesn't happen with other plugins. MIDI plugin is OK for example.

### LINKED ISSUE #833: OSC plugin on 127.0.0.1 receives packets from other interfaces
This is so wrong, and it is caused by binding a socket only by port and not by port+IP.

To reproduce:
- activate the OSC 127.0.0.1 input line
- send OSC packets from another device to (for example) the eth0 IP address.

Result:
- packets are received and signals are emitted like if they were received on 127.0.0.1.

This is another regression as it was working correctly before.

plugz:
Also use safer pointers

Fix #833

I also tested #832
Without this patch, QLC+ crashes when unchecking/checking 2 times the input of a line with feedback enabled.
With this patch, no crash, and the input still works.
So maybe it fixes #832 too, or maybe my test procedure was wrong.

### DIFF
--- a/plugins/E1.31/e131controller.cpp
+++ b/plugins/E1.31/e131controller.cpp
@@ -436,5 +436,4 @@ void E131Controller::processPendingPackets()
                 << ", that does not look like E1.31";
         }
     }
-// TODO? use this test?                if (senderAddress != m_ipAddr || info.inputMulticast == false)
 }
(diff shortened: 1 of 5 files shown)


mcallegari:
Just tested. #832 and #833 seem to be both fixed.
I haven't had the chance to test something on 127.0.0.1 yet, but this Pr surely improve things.
I'll update the changelog
Thanks

PR #838, opened 2016-04-04
XYPad with ranges + reversed - Test file

werepak:
Added  writeReverseRange() and writeRange() for test XYpad ranges (full, symmetric and asymmetric)
Please use this with the original vcxypadfixture.cpp


mcallegari:
Hi, can we close this then ?

werepak:
Hi Massimo, I followed the directions of janosvitok. This is the PR with the test failing before the fix. If you decide to accept the PR with the fix, I think that this can be closed.

PR #839, opened 2016-04-04
XYPad with ranges + reversed possible patch

werepak:
in writeDMX() changed the math:
xmul = m_xMax - xmul -> xmul = 1 - xmul;
ymul = m_yMax - xmul -> xmul = 1 - ymul;

### DIFF
--- a/ui/src/virtualconsole/vcxypadfixture.cpp
+++ b/ui/src/virtualconsole/vcxypadfixture.cpp
@@ -485,9 +485,9 @@ void VCXYPadFixture::writeDMX(qreal xmul, qreal ymul, QList<Universe *> universe
     ymul = ((m_yMax - m_yMin) * ymul) + m_yMin;
 
     if (m_xReverse == true)
-        xmul = m_xMax - xmul;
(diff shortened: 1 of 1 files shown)


ui/src/virtualconsole/vcxypadfixture.cpp
```diff
@@ -557,7 +557,9 @@ void VCXYPadFixture::readDMX(QList<Universe*> universes, qreal & xmul, qreal & y
         return;
 
     if (m_xReverse == true)
```
mcallegari: 1- the language to be used is english
2- adding those debug lines breaks the if condition
3- please remove those lines, as writeDMX is a high frequency called method

werepak: Thanks for notice me. I saw them too late. These two lines were there only for private test and I forgot them there -.- New pull request will have them removed.

werepak:
I'm closing this pull request because it will be useless soon: the next pull release will bring the corrected version along with test sources, that **should work**

PR #840, opened 2016-04-04
XYPad reverse test with fix

werepak:
This is the package that **should** compile with travis-ci.  It contains the modified test .cpp with latest corrections along with the **"should be"** fixed vcxypadfixture.cpp giving the intended results.

### DIFF
--- a/ui/src/virtualconsole/vcxypadfixture.cpp
+++ b/ui/src/virtualconsole/vcxypadfixture.cpp
@@ -485,9 +485,9 @@ void VCXYPadFixture::writeDMX(qreal xmul, qreal ymul, QList<Universe *> universe
     ymul = ((m_yMax - m_yMin) * ymul) + m_yMin;
 
     if (m_xReverse == true)
-        xmul = m_xMax - xmul;
+        xmul = 1 - xmul;
     if (m_yReverse == true)
-        ymul = m_yMax - ymul;
+        ymul = 1 - ymul;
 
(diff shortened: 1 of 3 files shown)


mcallegari:
Adding the reference to the forum discussion:

@janosvitok since you followed the discussion, is this PR OK ?
Otherwise I'll try to check it when I can

janosvitok:
Luca,

I've changed your code a bit: I believe that the "reverse" flag should change just direction but not the range. Example: 0%-20% reversed is 20%-0% rather than 100%-80%. I hope that this way it is more intuitive.

I've also rewritten a bit your tests. I hope they are now more readable. This is the technique I've used: http://doc.qt.io/qt-5/qttestlib-tutorial2-example.html

Finally, I'm sorry it took me so long to process this PR, but as you know, I'm doing this in my free time, and I don't have much of it lately...

Thank you for your work! Also, you get the bonus points for unit tests :)

werepak:
Hey Janosvitok, thank you for your help. I'm not a programmer, so I'm already happy to have done something that worked (in one way or another xD). 
You're right about the behaviour of the "reversed" flag: I was missing the way how to not reverse also the range, but this was my terrible math deficit, and not pretty sure if it was a real problem during programming (I was hoping in a windows version to put to the test this, but you did a better work :P)
Thank you again for your corrections! I will read the article you've linked in case I will need it again.

PR #846, opened 2016-05-01
Rgbscript ballscolor

prutseltje:


mcallegari:
Sorry, I don't get the meaning of this.
Can't we just add the color selection on the existing balls.js script ?

Also, I noticed on both (original and yours), the parameter "Collision" has a typo

mcallegari:
Ugh, closed
I just wanted to understand if we could merge everything into a single script (if that makes sense...)

prutseltje:
Hi,

I've already merged it already into one script, but before pushing it to
git I want to make sure everything works. I'll push it to git when I am
home from work.

Rob Nieuwenhuizen
Op 3 mei 2016 17:44 schreef "Massimo Callegari" <notifications@github.com>:

PR #847, opened 2016-05-03
Add fixed color selection to balls rgbscript.

prutseltje:


mcallegari:
Hi @prutseltje, sorry for the delay, I got dragged into other things.

So, I've checked both versions of the balls color script and I have to say I changed my mind and I prefer the first version you submitted.

Reasons:
- the 2nd version breaks backward compatibility cause of the "Color Type" parameter. The default should be QLC+ to preserve compatibility, but instead it's "fixed", so users using the balls script would see their matrices changing
- in the 2nd version you added 10 preset colors (while on 1st version there were 5) and this is what happens on a "low" resolution screen (1366x768):
image

So, if you don't mind, I would merge the "Balls (colors)" version. In case, please let me know if 5 colors are enough (I think so) or if we really need more.

Thanks !

prutseltje:
Hi Massimo,

I don’t have the old script anymore :( I understand that the backwards compatibility is a issue.

Currently I am moving to a new home and am not able to make a video of the new script on my 15*33 rgb led panel. On this panel the 10 bright colored balls are really nice to see!
Is it possible to make the properties window scroll when the text won’t fit anymore?

Maybe we can call this script balls (colors) since it has a addition of the colors.

The other option is an extra option pulldown menu with colors fixed yes/no apart from the old color selection.
Or name the presetColors back to presetRandom and just add the extra option here.

Please let me know what you prefer so I can adjust the script.

Kind Regards,
Rob Nieuwenhuizen

mcallegari:
Hi Rob

>> I don’t have the old script anymore :( I understand that the backwards compatibility is a issue.

No worries. GitHub has memory of it, and in any case I saved a copy. I will commit it later today, also with the typo fixed.

>> Is it possible to make the properties window scroll when the text won’t fit anymore?

That's what I wanted to avoid. This is already addressed on QLC+ 5 and at this stage I'd like to avoid to dedicate more effort to QLC+ 4

I'd say, let's go for the "old" 5 colors script and eventually we can increase the number of colors in the future

mcallegari:
Pushed ! 0b170f62669f16cd82de0114873586532b27a23c

PR #857, opened 2016-05-27
American dj event bar pro

boxy321:
I missed the master dimmer from the heads first time round.

http://www.qlcplus.org/forum/viewtopic.php?f=3&t=10017


mcallegari:
Mode names are just numbers. They should be "xx channel".
I'll fix it.
Thanks

PR #860, opened 2016-05-29
Update spanish translation

tornel:


mcallegari:
Thanks Luis :+1: 
Unfortunately this didn't make it in time for the release :(
My bad indeed for not informing you earlier

PR #862, opened 2016-06-01
New fixture - Pro-Lights-LumiPIX-12Q

boxy321:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=10044


mcallegari:
"X Ch Mode" should be "X Channel"

PR #863, opened 2016-06-01
New fixture - Stairville-DCL-Flat-Par-18x4W-CW-WW

boxy321:
Not sure of the use of amber for warm white, but...

http://www.qlcplus.org/forum/viewtopic.php?f=6&t=9958


mcallegari:
Modes names not OK. I'll fix them.
Thanks

PR #864, opened 2016-06-02
Adding iSolution iColor4 fixture

pade:
Hi,

Is it possible to add this fixture to the main ? I test it with the product.
If you think I need to modify something, please let me know.

Thanks


mcallegari:
Added, thanks.
Next time please share also the product manual cause it helps to review the definition
https://www.maxlight.ru/downloads/acme/icolor-3-4.pdf

PR #865, opened 2016-06-02
New fixture - ETC-ColorSource-Spot

boxy321:
http://www.qlcplus.org/forum/viewtopic.php?f=6&t=10009

Physical and photometric data is from the ETC website (I have used maximum values).


mcallegari:
Thanks

PR #867, opened 2016-06-02
New fixture - American-DJ-Chameleon-QBar-Pro

boxy321:
Datasheet here:

[http://www.adj.com/chameleon-qbar-pro](url)

Lots and lots and lots of colour macros.


mcallegari:
Massive definition !
Thanks :+1:

PR #868, opened 2016-06-03
New fixture - American-DJ-Vizi-Beam-RXONE

boxy321:
http://www.americandj.eu/en/vizi-beam-1rx.html


mcallegari:
Pity for the gobos.
Thanks :+1:

PR #872, opened 2016-06-22
show pagenumber on multipage frames when frame is collapsed

mgubisch:
Pagenumber is now shown for collapsed frames
The page navigation buttons are still hidden

hide the navigation button on initializing a collapsed multipaged frame too

The discussion thread is here:
http://www.qlcplus.org/forum/viewtopic.php?f=12&t=10125


mgubisch:
Can't figure out why appveyor build is failing, project builds fine on my machine (windows 10 and ubuntu) using the provided build instructions from the wiki.

janosvitok:
Appveyor timed out. See tab messages in appveyor. The build can't take more than one hour and it is not easy to do here... and it even doesn't run tests... Since travis was successful, the commit should be OK (buildable & doesn't break tests).

janosvitok:
Thanks!

PR #873, opened 2016-06-29
New fixture - Proel-PLLEDMLBG

boxy321:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=10144

The creator has added an 8-channel, 8-bit mode not mentioned in the manual. I left it in.


citofono:
Hi Robert
I confirm the presence of 8 channel mode, even if there's no tables specifing that.

The following information is provided in the manual at page 19:
" CHSL
CHSL is defined as Channels Switch which total are two option include ON & OFF. When it is in “ON” position, the light is maximum channels. When it shows “OFF”, the light is the last two channels. 16 Bit is useless. 
"

PR #876, opened 2016-07-25
New fixture - Martin-MH2-Wash

boxy321:
http://www.qlcplus.org/forum/viewtopic.php?f=3&t=10189


mcallegari:
Pan/Tilt fine groups are wrong.
Rest is OK, thanks

PR #879, opened 2016-08-26
Adds Indigo as additional channel color

Habefaro:
While working with the Source Four LED lustr from ETC I wrote the fixture, which I will add to the forum later. Since it has Indigo as LED color, I added the possibility to choose it in the fixture creator.


mcallegari:
Looks perfect to me !
You also fixed some Lime missing code :+1: 
Thank you very much.

What name should I put in the changelog ?

Habefaro:
Thank you for this great program, it was a pleasure!
The fixture for the Source LED lustr comes soon in the forum with the
new color =).

The real name is: Axel Metzke

Am 28.08.2016 um 19:01 schrieb Massimo Callegari:

PR #884, opened 2016-09-15
Hysteresis

karrika:
GPIO inputs are noisy. Even with good power supplies and proper electronics in the front you get occasional false triggers (I had 2 last summer. One by itself before the show and one when someone stepped on some cables). This patch requires the value to be stable for 200ms before we accept the trigger as real. The hysteresis work both ways so you also check for release of the button.

It does slow down the response from a button press by 200ms. But on the other hand you can also use internal pull-ups and cheap power supplies without problems with your buttons.

The best thing is that you get no false triggers. And that is quite important for the end user.

### DIFF
--- a/plugins/gpio/gpioreaderthread.cpp
+++ b/plugins/gpio/gpioreaderthread.cpp
@@ -22,6 +22,8 @@
 #include "gpioreaderthread.h"
 #include "gpioplugin.h"
 
+#define HYSTERESIS_THRESHOLD  3
+
 ReadThread::ReadThread(GPIOPlugin *plugin, QObject *parent)
     : QThread(parent)
     , m_plugin(plugin)
@@ -110,14 +112,17 @@ void ReadThread::run()
             uchar newVal = dataRead.toUInt();
             if (newVal != gpio->m_value)
             {
-                gpio->m_count += 1;
-                if (gpio->m_count > 3) {
+                gpio->m_count++;
+                if (gpio->m_count > HYSTERESIS_THRESHOLD) 
+                {
                     qDebug() << "Value read: GPIO:" << gpio->m_number << "val:" <<  newVal;
                     gpio->m_value = newVal;
                     gpio->m_count = 0;
                     emit valueChanged(gpio->m_number, gpio->m_value);
                 }
-            } else {
+            } 
+            else 
+            {
                 gpio->m_count = 0;
             }
         }
--- a/plugins/gpio/gpioreaderthread.cpp
+++ b/plugins/gpio/gpioreaderthread.cpp
@@ -110,9 +110,15 @@ void ReadThread::run()
             uchar newVal = dataRead.toUInt();
             if (newVal != gpio->m_value)
             {
-                qDebug() << "Value read: GPIO:" << gpio->m_number << "val:" <<  newVal;
-                gpio->m_value = newVal;
-                emit valueChanged(gpio->m_number, gpio->m_value);
+                gpio->m_counter += 1;
+                if (gpio->m_counter > 3) {
+                    qDebug() << "Value read: GPIO:" << gpio->m_number << "val:" <<  newVal;
+                    gpio->m_value = newVal;
+                    gpio->m_counter = 0;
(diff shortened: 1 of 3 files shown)


janosvitok:
- you may [squash](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) the commits to one -- it makes nicer history (but you don't have to).
- make 3 a constant ("THRESHOLD"?; maybe with explanation of the relationship between this number and actual time)

karrika:
I was thinking about making 3 a constant and even adding comments. But the number 3 is related to usleep(50000) in the same file. It is not a constant either. It doe not even have 50ms in a comment. It may be easier to read what happens when they are both numbers. The style of the entire qlcplus is very sparse and I don't want to change the style.

plugins/gpio/gpioreaderthread.cpp
```diff
@@ -110,9 +110,15 @@ void ReadThread::run()
             uchar newVal = dataRead.toUInt();
             if (newVal != gpio->m_value)
             {
-                qDebug() << "Value read: GPIO:" << gpio->m_number << "val:" <<  newVal;
-                gpio->m_value = newVal;
-                emit valueChanged(gpio->m_number, gpio->m_value);
+                gpio->m_counter += 1;
```
mcallegari: += 1 is python syntax. In C/C++ we use ++

plugins/gpio/gpioreaderthread.cpp
```diff
@@ -110,9 +110,15 @@ void ReadThread::run()
...
+                if (gpio->m_counter > 3) {
```
mcallegari: You said you don't want to change the code style, but still your style places parenthesis at the end of the line.

plugins/gpio/gpioreaderthread.cpp
```diff
@@ -110,9 +110,15 @@ void ReadThread::run()
...
+                    qDebug() << "Value read: GPIO:" << gpio->m_number << "val:" <<  newVal;
+                    gpio->m_value = newVal;
+                    gpio->m_counter = 0;
```
mcallegari: This means that even switching from high to low requires to pass the threshold check. Doesn't make much sense to me. Once the state change is triggered, going back to the original state should be immediate

mcallegari:
> The style of the entire qlcplus is very sparse and I don't want to change the style.

Can you please elaborate this ? Does it refer only to defines or also to other parts that you can mention ?

karrika:
Noise works both ways. If switch back is instant then we also get a second unwanted trigger 200 ms after the instant switch to zero. So I would like to filter edges both ways.

karrika:
Thank you Massimo. I was not sure if using #define's would be better than "const int definitions".
I was abroad for a long week-end working with dance. So I could not make the changes without a computer.
I have even been thinking of making the Hysteresis_count a variable. At zero it would not affect reading GPIO in any way and at 3 it would have the preferred operations for Raspberry Pi. But extra settings may just confuse the user so it may be best to have it as it is now.

mcallegari:
2 questions, maybe unrelated:
- where does 200ms come from ? I mean, would it work with 100ms or 150ms or would it be best 300ms ?
- I'm not an electronic expert, but to clean a noisy line, isn't condensers the most appropriate component to be put on the line ?

karrika:
Yes. It would work with 1 = 100ms, 2 = 150ms and 3 = 200ms. I tried 200ms and was happy with it. There is no engineering reason behind it. But of course it is extremely likely that a noise spike would be recorded in the same state every 50 ms 4 times in a row. Usually you get less than one spike per 4 hours. So 100ms should be ok too. 300ms is definitely overkill.

The problem is noise generated by the chip itself or noise coming through the power supply. The filtering electronics take care of noise coming from the wires. This is a bit disappointing for me to realize that Raspberry Pi is a toy. But I have no other explanation for all these random trigger discussions on the Raspberrypi.org forum. I was able to record a spurious trigger even with no wire connected to the GPIO pin - just a 10k pull-up resistor. You may generate triggers by touching the board while it is running.

mcallegari:
Alrighty, let's merge this

PR #895, opened 2016-09-29
New fixture - High-End-Systems-Studio-Beam

boxy321:
Had to reclone to add this... Things got a bit screwed up.

http://www.qlcplus.org/forum/viewtopic.php?f=6&t=10351&start=10


mcallegari:
Thanks :+1:

PR #899, opened 2016-10-19
Update qlcplus.install

trebmuh:
Fixes for the udev rules location.
Following https://lintian.debian.org/tags/udev-rule-in-etc.html recommendation.

### DIFF
--- a/debian/qlcplus.install
+++ b/debian/qlcplus.install
@@ -1,8 +1,8 @@
 usr/*
 
 # udev rules
-lib/udev/rules.d/z65-dmxusb.rules
-lib/udev/rules.d/z65-peperoni.rules
-lib/udev/rules.d/z65-anyma-udmx.rules
-lib/udev/rules.d/z65-spi.rules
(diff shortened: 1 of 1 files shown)


mcallegari:
I'm not sure about this one.
First, it breaks backward compatibility, and possibly doesn't work on older Debian/Ubuntu versions.
Secondly, Lintian says that /etc/udev/rules.d is reserved for user-installed files, so it is still a valid folder and doesn't seem a big deal to keep QLC+ rules over there.

trebmuh:
Point agreed on backward compatibility.
For the other one, it's not the best solution since in this case "user-installed" files refers to "user manually installed" file. Not that goods in regards of keeping a clean system.

Anyway, you're free to take it or not, I've got no issue with that. Bear in mind that I'm working "the debian way" (or at least trying to), so it would need to be done anyway if QLC+ is expected to enter in Debian on one of these days.

trebmuh:
I'm closing this PR and removing my forked branch.

PR #902, opened 2016-11-16
RGB Matrix with generic dimmers

### LINKED ISSUE #827: RGB Matrix with generic dimmers
Opening this so it doesn't get forgotten
As explained by @janosvitok here: http://www.qlcplus.org/forum/viewtopic.php?f=5&t=8088&start=10#p42488

When creating a RGB matrix with generic dimmers, the bahaviour is not correct, as all the heads go to 100% intensity.

This is because the RGB Matrix function doesn't handle properly the intensity channel. We need to add an extra check for generic dimmers.

janosvitok:
This should fix #827. First commit adjusts fixture definitions, the other code.

### DIFF
--- a/engine/src/efxfixture.cpp
+++ b/engine/src/efxfixture.cpp
@@ -99,7 +99,8 @@ void EFXFixture::setHead(GroupHead const & head)
             (fxi->tiltMsbChannel(head.head) != QLCChannel::invalid()))
         modes << PanTilt;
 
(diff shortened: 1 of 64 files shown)


mcallegari:
2 comments for now, since this is quite a big patch:
- some fixtures actually have separate dimmer channels for separate "heads". Is your change to the fixture library considering that ?
- do you have a sample workspace you used to test the changes that I can use as well in conjunction with my other (colorful) projects ? :)

PR #908, opened 2016-11-25
uninitialized variable in GPIO plugin

rawbengal:
m_enabled was not initialized causing some pins to never get
set as exported


rawbengal:
Argh...  I tried to submit these as three separate requests but somehow Git decided to put them together into one.  Two bugs... uninitialized member variable... incorrect universe when changes made through UI.  One style change.

mcallegari:
Good catches !
Thanks for this patch 👍

PR #909, opened 2016-11-27
New fixture and a small fix

McBochi:
Corrected the type of the "ADJ FREQ 16" and added the "ADJ FREQ 5" as a new fixture.

### DIFF
--- a/resources/fixtures/American-DJ-FREQ-16-Strobe.qxf
+++ b/resources/fixtures/American-DJ-FREQ-16-Strobe.qxf
@@ -8,7 +8,7 @@
  </Creator>
  <Manufacturer>American DJ</Manufacturer>
  <Model>FREQ 16 Strobe</Model>
- <Type>Color Changer</Type>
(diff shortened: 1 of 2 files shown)


janosvitok:
Please fix channels vs heads in the new fixture:
- 7 channel mode: channels 6 and 7 should not be included in any head; there should be only 5 heads
- 2 channel mode and 1 channel mode: no head should be defined

McBochi:
Ah, sorry for that - going to fix that!

McBochi:
Quick follow-up question: Got another fixture ready - an RGB PAR-can.
However, the included LEDs can't be controlled separately, so should I even define a single head for all of them?

janosvitok:
if the LEDs are controlled together as one light, don't create a head. Heads are needed only for lights with >= 2 separately controlled parts.

mcallegari:
Physical information missing.
I don't understand why everybody submit fixtures through the forum and you came directly here, despite the message that this is an area for developers

PR #915, opened 2016-12-26
Debian packaging

trebmuh:
As suggested by https://lintian.debian.org/tags/maintainer-script-lacks-debhelper-token.html


mcallegari:
I tried to read something about this, but I'm still confused.
I just hope it doesn't end up in weird post installation results :(

trebmuh:
Let me know if so.
Thanks for the merge.

Issue #439, opened 2015-01-19
Crossfade sliders problem when transitioning and fixture didn't change 


albertolopez:
I am trying to use the crossfade sliders but I've found a problem.

In my cue list, there are some scenes where the fixture intensity is not changing. However, when use the crossfade sliders, those fixtures start from intensity 100%, then go to 50% and then to 100% again.

For example:

Scene 7. Fixture 1 => Intensity = 255, fade in = 0, fade out = 0, hold = infinite, duration = infinite

Scene 8. Fixture 1 => Intensity = 255, fade in = 0, fade out = 0, hold = infinite, duration = infinite

So, in this case when transitioning from scene 7 to scene 8 there should be no changes in the intensity. However this is what is happening:

Scene 7. Fixture 1 => Intensity = 255
Scene 7,5(in the middle of the transition to 8) . Fixture 1 => Intensity = 128
Scene 8. Fixture 1 => Intensity = 255

More info here
http://sourceforge.net/p/qlcplus/discussion/general/thread/a44f3846/
http://sourceforge.net/p/qlcplus/discussion/general/thread/d6f0ed17

albertolopez:
I started looking at the code and trying to understand the problem. But I may need help from a dev of the project.

First I will need to know where the value for the intensity is read. I think it might be here.
```
void GenericFader::write(QList<Universe*> ua)
{
...
        // Apply intensity to HTP channels
        if (grp == QLCChannel::Intensity && canFade == true)
            value = fc.current(intensity());

```
However, when using the crossfade sliders linked the m_current is always the active one, even after the slider passed 50% so I'm missing something here. Once I understand how it works I probably can move forward.

A possible solution I thought it could be setting the intensity by channel instead of using the GenericFader. However I'm sure there are other better alternatives.

janosvitok:
Crossfaders are handled in VCCueList::slotSlider1ValueChanged and VCCueList::slotSlider2ValueChanged. See (this line)[https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L751](setting the other slider), (this line)[https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L758](controlling step 7 intensity from 100% to 0%) and (this line)[https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L801](controlling step 8 intensity from 0% to 100%).

[...]

PR #916, opened 2016-12-29
Fixed issue #439

anorod:
Now when we have a crossfader on a cuelist and we have the faders linked the faders will act as:
> Scene = Crossfader1 * SceneValues1 + Crossfader2 * SceneValues2

Tested with several Cuelists and different objects. Only adds when there are two scenes.

Also in the process:
* Added the function addFrom to the Scene and Function class so we can add the values of the scene from another scene.
* Added the attribute CrossfaderId which identifies the scenes to be added. Must be at 0 so there is no addition when not needed.

This is very helpful when QLC+ is used in theatre lighting.


anorod:
The tests that fail are from EFX_Test which I didn't modify. Will check but maybe are unrelated to this pull request?

mcallegari:
> The tests that fail are from EFX_Test which I didn't modify. Will check but maybe are unrelated to this pull request?

Test units don't lie. Your change broke the EFX Function. Before your change all the test unit pass.

In any case I won't merge this, because you insisted to make changes to the code without having the proper knowledge of how QLC+ works and most likely how C++ works.

Things like "CrosfaderId" (double s maybe ?) means you haven't even reviewed your code.
Same story for "#CONFIG         -= debug" in variables.pri
Another example is "function->type() == 1" in mastertimer.cpp, which should tell you you are in the wrong place for making such check.
And again adding a cross fader attribute to EVERY function doesn't make any sense to me.
And in general, when changing other's code, you should follow the code style adopted, in this case by aligning brakets to new lines, and not at the end of them.

Plus, will you take the responsibility for hundreds of users for the ugly change in MasterTimer where you loop through the whole Function list and eventually make a copy of a Function ?
Have you got the faintest idea of what that means when several Functions are running on a Raspberry Pi ?

anorod:
Thanks for your reply. I will check all of what you said and of course make sure all the test pass fine.

Because the changes will surely not merge I will not worry for optimizing for Raspberry users, as it's not my case or the people I work or share the program with. So far I tested the performance with my last theatre show (60 fixtures, 350 cues and 7 simultaneous chasers) and moved fine.

I was just trying to help to solve this. Thank you for your time and sorry for the inconvenience. I will not disturb you again.

mcallegari:
If it works for you then good !
But one thing is if it works for one, another is if it works for everyone,
Making changes to QLC+ is not easy at all, considered the large amount of usage cases out there.
As maintainer of the project, I need to consider everything, and it often happens that making some users happy has the downside of making others unhappy.
The more I limit this, the more successful is the project. I hope you understand there is nothing personal on this.

In any case, in these days I'm processing a few "most wanted" features, and #439 is one of them.

anorod:
I understand it's not personal, and I thank you for taking your role as maintainer seriously.

Even though I think this last interactions were a rough start, I can't argue that all you said was right and thinking about all the users of QLC+ and not just theatre ones. I didn't think on Raspberry Pi cases, and the Crossfader attribute is a crappy fix that worked for me but will probably give problems with future developments. To make it right involved changing big things that I didn't fully understand and will surely involve complex problems.

But this is the beauty of open source. I can report bugs when I see them: some of them get fixed and some of them are features that hang in there for some time. But if I need something I can particularize the code to do it myself somehow, and offer those changes back to the community in case they can help. Sometimes they fully do, sometimes partially, and sometimes they don't. It's fine in all cases.

Thank you for making QLC+ open source and for processing #439 in these days. I can assure you I will happily test it and report if I see something strange while using it.

engine/src/function.cpp
```diff
@@ -1235,14 +1236,23 @@ void Function::adjustAttribute(qreal fraction, int attributeIndex)
...
+    }
     emit attributeChanged(attributeIndex, m_attributes[attributeIndex].value);
 }
 
 void Function::resetAttributes()
 {
-    for (int i = 0; i < m_attributes.count(); i++)
-        m_attributes[i].value = 1.0;
+    for (int i = 0; i < m_attributes.count(); i++){
+        if(!QString::compare(m_attributes[i].name, "CrosfaderId", Qt::CaseInsensitive)){
```
janosvitok: This line is the one that broke EFX tests. The "!" should probably be not there.
You effectively set Intensity for all functions to 0 here.

anorod: Thanks!
I tested and seems is not the problem. When I quit the "!" I get the error in chaserunner because now it seems I'm putting the intensity to 0 as you said.
    
    FAIL!  : ChaserRunner_Test::adjustIntensity() Compared doubles are not the same (fuzzy compare)
      Actual   (m_scene1->getAttributeValue(Function::Intensity)): 0
      Expected (qreal(1.0))                                      : 1

In the doc it says compare will return 0 when true, so that's why I put it like that:

    [static] int QString::compare(const QString &s1, const QString &s2, Qt::CaseSensitivity cs = Qt::CaseSensitive)
    Compares s1 with s2 and returns an integer less than, equal to, or greater than zero if s1 is less than, equal to, or greater than s2.
    If cs is Qt::CaseSensitive, the comparison is case sensitive; otherwise the comparison is case insensitive.
    Case sensitive comparison is based exclusively on the numeric Unicode values of the characters and is very fast, but is not what a human would expect. Consider sorting user-visible strings with localeAwareCompare().

      int x = QString::compare("aUtO", "AuTo", Qt::CaseInsensitive);  // x == 0
      int y = QString::compare("auto", "Car", Qt::CaseSensitive);     // y > 0
      int z = QString::compare("auto", "Car", Qt::CaseInsensitive);   // z < 0

On the other hand, I saw that when I initialize the CrosfaderId (will change it to two s in next commit) to '0' [in this line](https://github.com/anorod/qlcplus/blob/develop/engine/src/function.cpp#L130), I get the test errors. If I do `registerAttribute(tr("CrosfaderId"));` instead of `registerAttribute(tr("CrosfaderId"), 0);` I pass all the unit tests. Still working on why.

anorod: Got it!

Had to include also the CrossfaderId in the attributes of the efx [here](https://github.com/anorod/qlcplus/blob/develop/engine/src/efx.h#L83) Doing that passes all the test.
That's because I put the attribute in Function and not in Scene.

Anyway, it works for me now and passes all the unit tests. Thanks to all and have a great year!.

PR #917, opened 2017-01-01
Added direction to RGB panel creation dialog.

raivisr:
Had to work with panels where LED pixels were interconnected top to bottom instead of side to side and it was easier for me to add this feature to code than to figure how to transpose matrices myself. Hopefully others will find this feature useful as well.
If any code release forms have to be signed by me, I will gladly do that.

### DIFF
--- a/resources/docs/addrgbpanel.html
+++ b/resources/docs/addrgbpanel.html
@@ -109,6 +109,18 @@ Let's have a look at every option in this panel:
   of the end of a row is on the first column of the next row.</TD>
  </TR>
 
+ <TR>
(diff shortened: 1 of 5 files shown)


mcallegari:
Looks good to me, thanks !
As this increases the possible combinations, have you tested that this doesn't break some cases ?

raivisr:
I actually tried creating all possible combinations with horizontal and vertical layouts and they all looked OK to me, but that was on 1st of January, so second pair of eyes might be good idea.

mcallegari:
Alright, let's merge this and I'll do some more tests myself.
Thanks for the contribution

PR #918, opened 2017-01-01
Memory optimisation

janosvitok:
This is a little experiment in decreasing memory footprint inspired by the "70 MB" showfile.
Since it's too much for my computer with 4GB to handle, I extracted one sequence with 9600 steps to separate file, and run valgrind --tool=massif on it (the file has no VC widgets and still it's 3 MB).
Baseline peak memory usage was 299 MB.

First commit makes SceneValue's destructor non-virtual which saves us size of virtual table (one pointer here).
Massif peak usage 199 MB. 30% down. Nice...

Second commit avoids some conversions and uses QString::splitRef instead of split to avoid creating QStrings and copying the data. It doesn't save memory, but may help with fragmentation. Massif says 194 MB. Not so big difference but still something. Note: splitRef is supported from QT 5.4.

Third commit combines fixture id and channel to one quint32. This saves one quint32 out of each SceneValue. It's a memory/time tradeoff. The time penaulty is decreased by the fact, that SceneValue comparison is simplified to quint32 comparison now. Massif says 159 MB (47% saved from the baseline).

Some memory could be saved by storing SceneValues in QVector instead of QList in ChaserStep (one pointer per SceneValue). Problem here is that Scene keeps SceneValues in QMap, and QMap::keys() provides the values as QList, so conversion to QVector must be provided in some place.

### DIFF
--- a/engine/src/channelsgroup.cpp
+++ b/engine/src/channelsgroup.cpp
@@ -76,7 +76,7 @@ void ChannelsGroup::slotFixtureRemoved(quint32 fixtureId)
     while (channelsIt.hasNext())
     {
         SceneValue scv(channelsIt.next());
-        if (scv.fxi == fixtureId)
+        if (scv.fxi() == fixtureId)
         {
             channelsIt.remove();
             hasChanged = true;
@@ -170,20 +170,20 @@ QString ChannelsGroup::status(Doc *doc) const
 
     foreach (SceneValue value, m_channels)
     {
-        Fixture *fixture = doc->fixture(value.fxi);
+        Fixture *fixture = doc->fixture(value.fxi());
         if (fixture == NULL)
             return QString();
         const QLCFixtureMode *mode = fixture->fixtureMode();
         QString chInfo("<TR><TD>%1</TD><TD>%2</TD><TD>%3</TD></TR>");
         if (mode != NULL)
         {
-            info += chInfo.arg(fixture->name()).arg(value.channel + 1)
-                .arg(mode->channels().at(value.channel)->name());
+            info += chInfo.arg(fixture->name()).arg(value.channel() + 1)
+                .arg(mode->channels().at(value.channel())->name());
         }
         else
         {
-            info += chInfo.arg(fixture->name()).arg(value.channel + 1)
-                .arg(QString(tr("Channel %1")).arg(value.channel));
+            info += chInfo.arg(fixture->name()).arg(value.channel() + 1)
+                .arg(QString(tr("Channel %1")).arg(value.channel()));
         }
     }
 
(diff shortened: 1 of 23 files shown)


mcallegari:
Nice analysis, but I'm afraid the only change that can go upstream is the virtual destructor one.

splitRef: Unfortunately we still have to keep the Qt4 compatibility, so unless everything is placed under a Qt version check, that can't go up

quint32: I have plans for the fixture ID and I will send a separate email for that. Thing is that I need it to stay 32bit. By the way your change create a mix of 32/16bit fixture IDs, which can cause weird things, especially with the Fixture::invalidId() number, which is 32bit.

Instead, I was thinking of another optimization that can be done on Sequences. My initial idea was to write every non-zero value of each step in the XML file, to still allow manual change/inspection of the values.
However I realized it's nearly impossible to understand which is which from a comma separated list of numbers. So the idea to reduce the XML file size is to write full data only of the first step, and then save just the delta values of the next steps.
This will mean: DO NOT touch the XML file :)

janosvitok:
> By the way your change create a mix of 32/16bit fixture IDs

Yes, I limited fixture IDs and channels per fixture to 16 bits. I think it is sufficient. 
Handling of Fixture::invalidId() is added [here](https://github.com/mcallegari/qlcplus/pull/918/commits/bb1391e9263d22a4509d5cfbafb86254ca06c961#diff-cb2732b8c6be3e6a1f6483dbf073bf2cR100) and few lines below. It adds another "if" but keeps the logic hidden within the class.

Rewriting XML will make the qxw file smaller, but the problem I'm solving is memory usage when the file is loaded (although in this case RGB Matrix is better solution, so it may not be so hot).

FixtureID + channel may even be extracted to separate class, let's say FixtureRelativeAddress or FRA :)
This combination is used in many places throughout the code, and creating FRA may avoid splitting the address in many places.

I will wait for the mail :)


I agree about splitRef for now. I will remove it from PR. BTW. do we have publicly available requirements for building QLC+? i.e. min supported QT4 version, min. QT5 version, min gcc version?

mcallegari:
We can even think of having 16bit IDs everywhere, but there's a "but" :)
I believe what Heikki intended with 32bit IDs everywhere was not just for having 4 billion possible fixtures (who cares ?) but to have 32bit words that a compiler and a processor can easily handle.
Now, while 15 years have passed and PCs are more powerful and 64bit, I think you can't really tell the difference on a PC, but it might still help less powerful platforms like the Raspberry Pi.
This also includes some Android/iOS platforms where QLC+ 5 will run.

janosvitok:
Let's close this. I've committed the first commit and I'll leave the rest as it is.

PR #922, opened 2017-01-18
Increase maximum fixture group dimensions that UI accepts as valid.

raivisr:
Fixture group maximum dimensions are limited in UI. One can easily create fixture group larger than 99px by 99px by adding RGB panel, but can not enter dimensions larger than 99px by 99px in fixture group editor or create dialog.
I increased limit to 999, which should suit most of us for time being.


mcallegari:
Thanks :+1:

PR #924, opened 2017-01-18
Fixed disabled widgets in frame after page switch

siegmund42:
Had to call setEnabled() on the frame itself to make it work, but only when activating.
There is still an issue present that the disable state of a frame in a multipaged frame is not saved. But this is not so easy to implement with the current architecture.


mcallegari:
Hey, thanks for this, but please see 
http://www.qlcplus.org/forum/viewtopic.php?f=33&t=10732&p=46263#p46263

Issue #923, opened 2017-01-18
Getting unexpected MIDI feedback from another page


mcallegari:
Reference for tracking:
http://www.qlcplus.org/forum/viewtopic.php?f=5&t=9587

PR #925, opened 2017-01-18
Fixed unexpected feedback from another page

siegmund42:
Uses recently added acceptsInput() method since a disabled widget should not send any feedback from my point of view. If you care, it could be renamed or another sendsFeedback() could be added. Just let me know if you want something to change.
Fixes: #923

### DIFF
--- a/ui/src/virtualconsole/vcwidget.cpp
+++ b/ui/src/virtualconsole/vcwidget.cpp
@@ -672,20 +672,23 @@ void VCWidget::sendFeedback(int value, QSharedPointer<QLCInputSource> src)
         if (src->needsUpdate())
             src->updateOuputValue(value);
 
-        QString chName = QString();
-
-        InputPatch* pat = m_doc->inputOutputMap()->inputPatch(src->universe());
-        if (pat != NULL)
+        if (acceptsInput())
         {
-            QLCInputProfile* profile = pat->profile();
-            if (profile != NULL)
+            QString chName = QString();
+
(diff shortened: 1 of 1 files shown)


mcallegari:
Looks OK, thanks.
Let's just hope this doesn't break something else somewhere else :(
Acting on bottom classes (in this case VCWidget) is always risky

mcallegari:
@siegmund42 oh, and if you want your last name in the changelog please just let me know :wink:

siegmund42:
> Let's just hope this doesn't break something else somewhere else :(
Well I hope this, too. I tested the version at an event on Friday and everything went quite well.

"Lukas Jähn" would be fine, thanks

siegmund42:
Well it finally happened. A user claims it's important to him having feedback to an external control while not be able to control the vc widget with the external control:
http://www.qlcplus.org/forum/viewtopic.php?p=46515#p46515

What to do?

- Consider this as a not typical use case, ignore it and leave it as it is
- Differentiate between "disabled" widgets and widgets that are on a page that is currently not active 
  - so either get if the widget currently is on an active page (could be hard to implement getting the current page from a widget - nested frames?)
  - or simply introduce another property like `m_onActivePage` property that changes on page switches.

siegmund42:
Actually I realized the difference is the visibility of vcwidgets. So widgets should only send feedback if they either accept input (we still need this for collapsed frames) or are visible.

I don't know If you re-open the PR my already pushed changes show up here.

PR #927, opened 2017-01-21
Fixed feedback not working with buttons on some controllers

siegmund42:
This happens for example at BCF 2000 and probably also APC40 mk2 due to some kind of internal feedback.
Only thing to do was to add updateFeedback() as soon as a button on external controller is released so that it has a proper LED state and not the one given by its "internal" feedback.
The change works great at my BCF 2000, since I have no other controllers at hand maybe someone could give it a try.

Bug has been reported here: http://www.qlcplus.org/forum/viewtopic.php?f=5&t=6900&p=29106
And I tried to describe it well here: http://www.qlcplus.org/forum/viewtopic.php?f=24&t=10158&p=45921#p45916

### DIFF
--- a/ui/src/virtualconsole/vcbutton.cpp
+++ b/ui/src/virtualconsole/vcbutton.cpp
@@ -534,11 +534,19 @@ void VCButton::slotInputValueChanged(quint32 universe, quint32 channel, uchar va
             else if (isOn() == true && value == 0)
                 releaseFunction();
         }
-        else if (value > 0)
+        else
         {
-            // Only toggle when the external button is pressed.
-            // Releasing the button does nothing.
-            pressFunction();
+            if (value > 0)
+            {
+                // Only toggle when the external button is pressed.
+                pressFunction();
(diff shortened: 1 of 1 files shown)


mcallegari:
See my comment here http://www.qlcplus.org/forum/viewtopic.php?f=24&t=10158&p=46443#p46443

False alarm ?

siegmund42:
I answered your post and had a short look at the code again and I think it is not possible to use a flash button with this configuration and the current implementation. Since the "extra signal generator" could not know if the external input is assigned to a flash button and the flash button does not know whether a signal comes from the external input or from the "signal generator" (last one is maybe implementable).
So either you end up with being unable to use a flash button at all or again a toggling button, although it should be flashing (depending on if the "generate extra signal" is enabled or not).
I would still stick to my code since it gives the ability to use flash buttons and is much more simple than implementing a method to distinguish signals as I wrote above.

What are your thoughts?

PR #928, opened 2017-01-21
Fixed choose external input does not consider page number

siegmund42:
This fixes the problem that the "Choose external input" dialog does not consider page numbers if it's executed in a multipage frame.

Reported here: http://www.qlcplus.org/forum/viewtopic.php?f=33&t=10612&p=46219


mcallegari:
As simple as that !
Thanks for this :)

PR #929, opened 2017-01-22
Improve speeddial

siegmund42:
This PR should improve the vc speeddial widget.

First commit fixes that the tap button blinking stops to work after one has changed the widgets value by any other method than the tap button.

Second commit adds feedback capability for the tap button. If desired, one could add custom feedback.


enbyted:
It works great, actually I was using similar modification (the feedback on tap) for quite some time - just never got around to cleaning it and making a pull request. 
However I think it would be more useful if the feedback was a short blink on each timer pulse (like 1/5th of the time). That's what I did, but it has some drawbacks:
- Twice as much feedback messages
- With short speedial settings feedback light might not light up
- It is inconsistent with visual feedback on screen
Still, that is the way that most 'professional' tempo LEDs work...
Maybe user should be given a choice?

siegmund42:
I thought about that possibility, too, since I know this behavior from a lot of sound consoles.
In my opinion the biggest problem is the inconsistency - if you could apply the blink also to the button in vc I would be fine with it.

One further thing would be to make the tap button always blink and indicate the current speed (even if the speed was not adjusted by tap). I haven't had a look at how complicated this is to implement, yet.

enbyted:
Changing the button in VC shouldn't be a problem.

I'll take a look at the second one - changing tap speed from other sources than just tap.
BTW. Is it possible to send pull request on top of another pull request? Like on top of your branch?

siegmund42:
Cool!
I don't know - but if you set up a PR on my branch and I merge it it should appear here so let's give it a try!

enbyted:
It probably won't copy my comment prom PR, so here I link it:
https://github.com/siegmund42/qlcplus/pull/2

enbyted:
Now, about changing how the tap button is flashed.

My idea is to speed up `m_tapTickTimer` in `SpeedDial` 5 times from the desired value and make `m_tapTick` be true for 1 tick and false for 4 ticks. I would also limit `emit tapTimeout()` to only when `m_tapTick` actually changes. This imposes addition of a counter variable thus increasing memory footprint by at least 1 byte, but in this case it doesn't matter - it's ui stuff.

I think this would be an elegant solution for making the tap feedback more 'professional'.
What do you think?

siegmund42:
I thought about adjusting the interval of `m_tapTickTimer`: in `updateTapTimer()` it could be `m_value/5` and depending on if `m_tapTick` is true or not it should be `4*m_value/5` or `m_value/5`.
At least this would save the extra byte. Don't know if there will be other problems coming up but it could be worth a try...

enbyted:
Hmm, I don't like it when timer changes it's period by itself, specially when this can be avoided.
Potentially it could also lead to desyncing with output - I'm pretty sure that changing timer's period resets it, thus we're losing a little bit of time with each tick.

Additionally it may lead to hiding bugs in future - when the timer updates it's period by itself based on current value the `updateTapTimer()` call might seem redundant, but in reality it would serve two purposes:
- Synchronizing tap timer with outputs
- Starting tap timer if it's stopped

But changing speedial's value without calling `updateTapTimer()` would change blinking period thus making an impression that this is enough. This could be traced with an assert, however.

Also there is another issue with dividing timer's interval - it's integer. And even 500ms/5 is not a round number, so in either scenario we're desyncing with output... Just realized that.

enbyted:
Try this https://github.com/enbyted/qlcplus/commit/f44d604ba1fcb2d55ea8039c9d77e1b0ece92049
It's the third option implemented.

siegmund42:
I tested this with a BCF2000 and it works quite well.

Just a suggestion: We could save a slot if we connect the timeout of the new timer to the already existing `slotTapTiemout()` method and in there only start the new timer if `m_tapTick == false`.

In any case, we should add an option to decide between the two modes. I would prefer two radio buttons in the external input tab of the speed dials' settings dialog under "Tap" section.
Depending on the decision we could simply start the new timer or not.

enbyted:
I think that would work, I'll test in a moment when I finish with sliders in webaccess.

Would you do the UI stuff? I don't have any experience with Qt's UI.

siegmund42:
Okay, the new feedback style is there now. The only drawback is, that the minimum timer duration for a proper visual feedback in ui is around 50ms. If you care, one could add an "if clause" to switch off feedback for times <50ms.

I don't have any experience in Qt (or at least c++ in general), either. But I can give it a try - always willing to learn something.
Before, could you @mcallegari  please state if you are even willing to merge this? Thank you!

enbyted:
> The only drawback is, that the minimum timer duration for a proper visual feedback in ui is around 50ms. If you care, one could add an "if clause" to switch off feedback for times <50ms.
Hmm, probably you are right. Do you think that making the button constantly lit would be good behavior? Or the opposite - make it dark? If this is implemented it should be also stated somewhere, so the user knows about it.

I would love to learn Qt in depth, just don't have the time... Like @mcallegari  curses Qt4's builds, I curse deadlines :P I just find this commit message -> 8624997 <- absolutely hilarious, sorry, had to mention it.

siegmund42:
> Hmm, probably you are right. Do you think that making the button constantly lit would be good behavior? Or the opposite - make it dark? If this is implemented it should be also stated somewhere, so the user knows about it.
I think leaving it constantly lit would be the best. This could be added to the documentation.
I tested this with the current implementation and it kind of works at least til 23ms (less I was not able to create by tapping with the mouse). But then it looks more like being on than being of all the time.

>  I just find this commit message -> 8624997 <- absolutely hilarious, sorry, had to mention it.
I saw this too and it definitely made my day :D

enbyted:
Looks great, man.

See the PR I created. I've implemented your suggestion and fixed these CI errors in the merge commit.

ui/src/speeddial.cpp
```diff
@@ -91,6 +91,7 @@ SpeedDial::SpeedDial(QWidget* parent)
     , m_tapTickTimer(NULL)
     , m_tapTick(false)
     , m_visibilityMask(DEFAULT_VISIBILITY_MASK)
+    , m_tapFeedbackType(Blink)
```
janosvitok: Siegmund, move this line before the above one. We have enabled check that initialization here  must be in order of declaration in the .h file, since that's how it happens anyway (.cpp order doesn't matter, the check is to avoid false information from constructor). This will fix travis build.

siegmund42: Thank you for the hint, the latest commit should fix this!

siegmund42:
Thank you for your suggestions!
I slept over the <50ms thing and I'm no more sure if setting the tap button simply to be on all the time is the way to go. The possibilities are:
 - Turn the tap button on (not the best regarding consistency, could be mixed up with long duration)
 - Turn the tap button off (see above)
 - Scale the flashing time in another way, e.g. use 1/4 if `m_value < 50 ms`, 1/3 if `m_value < 40 ms`, or so. Depending on how implemented, this limits the minimum displayable frequency.
 - Use always blinking feedback when `m_value < 50 ms`. Not fully consistent, too, but at least it blinks and in such frequencies this should be hard to notice, anyway. In addition to the fact that this would be easy to implement, it will not limit the minimum frequency other than it is already limited in the official code.
 - Set the flashing time to a fixed value if `m_value` is lower than some border. This again limits the minimum displayable frequency (but this limit will be around 10ms)

What do you guys think? I'd appreciate hearing what you like best.

enbyted:
The thing with people is that we can't really differentiate frequencies above 15Hz, I can think about a different solution: We could make the minimum displayable feedback be 50ms (20 Hz) and if `m_value` is less than this just set the tap timer to be 50ms. I'm sure that nobody will notice it and it will be consistent with other behavior. 
It should be noted, however, that for the old feedback style and `m_value` set to 50ms the blink frequency is in reality 10Hz. Still it's quite high for human eye and I think we could safely disregard it.

As for the new feedback style:
- I think the tap elapsed timer should not be shorter than 20 ms, because hardware feedback might start to be less noticible when this time is shorter. (keyword - might)\

I hope what I've written makes sense, I'm quite tired.

siegmund42:
> I'm sure that nobody will notice it and it will be consistent with other behavior.

Well, it will not be consistent with the timing itself. So basically this is kind of option #5 in my comment.

> I think the tap elapsed timer should not be shorter than 20 ms, because hardware feedback might start to be less noticible when this time is shorter. (keyword - might)\

I tested this with a BCF2000 and until 50ms everything works fine (feedback was noticeable, too).

siegmund42:
# Summary

1. The tap button will now blink/flash always and update, if the speed dials value has changed (whatever changed it)

2. Since in a lot of audio consoles the tap buttons are flashing instead of blinking, this feedback mode was added to the code. The user is given the choice which feedback mode he prefers in the properties dialog.
screenshot - 12 02 2017 - 00 12 45

3. Feedback is now sent to the external input assigned to the tap button.

The only thing we where not sure is what to do with the tap feedback on minimum speeddial values - especially if the flashing feedback is enabled. I decided to go with @enbyted's suggestion:
> I think the tap elapsed timer should not be shorter than 20 ms, because hardware feedback might start to be less noticible when this time is shorter. (keyword - might)\

So the flashing feedback is know limited to 20/200ms, in between the duration of a flash is `m_value/5`.

PS. Again a .deb package.
[improve-speeddial.zip](https://github.com/mcallegari/qlcplus/files/768986/improve-speeddial.zip)

ui/src/speeddial.h
```diff
@@ -81,6 +81,13 @@ class SpeedDial : public QGroupBox
         Infinite     = 1 << 7,
     };
 
+    // Feedback types for flash button
```
mcallegari: flash -> tap

ui/src/virtualconsole/vcspeeddialproperties.ui
```diff
@@ -51,10 +51,20 @@
...
+     <property name="standardButtons">
+      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
+     </property>
+    </widget>
+   </item>
    <item row="1" column="0" colspan="2">
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
+      <number>1</number>
```
mcallegari: And you changed again the default page displayed.
When you save a .ui file, you should save it with the first tab selected.

siegmund42: Sorry for that - I was not aware that the last saved state implies the default tab.
The latest commit processes both of you comments.

siegmund42:
To pick up the flashing duration problem: I was able to test a minimal duration on a sound console (Behringer X32) this weekend and it completely disables feedback below ~125ms duration.
So this could be a way to go, too.

mcallegari:
My opinion on this PR is that there shouldn't be an additional option for the tap flashing behaviour.
It adds confusion and it has to be documented (and it has to be ported to QLC+ 5)
When I implemented it, for simplicity I did the half/half flashing cause it came "free of charge", but since you implemented the traditional console flashing, I'd say that's the one everybody should use.

I haven't tested it yet on the controllers I have, but we need to be extremely careful on the flash time, cause at high speeds it creates a lot of feedback traffic to the plugins.

mcallegari:
And I'd also add: with some monitor resolutions or contrast, or ambient lights, flashing just the border of the tap button might be not be clearly visible, so I'd go with a solution that makes the whole button background to flash, with a proper color.

siegmund42:
> My opinion on this PR is that there shouldn't be an additional option for the tap flashing behaviour.
It adds confusion and it has to be documented (and it has to be ported to QLC+ 5)

I'd be okay with that, although I could document and port it if you want.

> I haven't tested it yet on the controllers I have, but we need to be extremely careful on the flash time, cause at high speeds it creates a lot of feedback traffic to the plugins.

That could be a problem indeed. So what do you think about disabling any feedback on frequencies below 125ms as it is on the Behringer X32?

> And I'd also add: with some monitor resolutions or contrast, or ambient lights, flashing just the border of the tap button might be not be clearly visible, so I'd go with a solution that makes the whole button background to flash, with a proper color.

Sorry but I don't like this idea so much especially for two reasons:

1. It would be not consistent with the status indication of other buttons.
2. The indication of the tap button is as important to notice as the indication of other buttons - these buttons have nearly the same indication mechanism as the tap button has at the moment and are pretty noticeable.

Anyway, if you still want to have it implemented like this let me know (especially what's a "proper color" for you). I have some time over the next few days and will fix the PR according to your decisions.

mcallegari:
>> And I'd also add: with some monitor resolutions or contrast, or ambient lights, flashing just the border of the tap button might be not be clearly visible, so I'd go with a solution that makes the whole button background to flash, with a proper color.
Sorry but I don't like this idea so much especially for two reasons:

> It would be not consistent with the status indication of other buttons.
The indication of the tap button is as important to notice as the indication of other buttons - these buttons have nearly the same indication mechanism as the tap button has at the moment and are pretty noticeable.
Anyway, if you still want to have it implemented like this let me know (especially what's a "proper color" for you). I have some time over the next few days and will fix the PR according to your decisions.

Lukas, that was just a proposal, no worries. Please let me explain why I came up with it.
When I implemented VC Buttons highlight, there was quite a long discussion in the forums, whether the "on" status were visible enough or not. A "proper color" in that case was a green border, but some people preferred the LED status (which ATM I'm not even sure someone uses)
Later on, when "presets" came in, I adopted a blue border, to distinguish them from the green border of VC buttons. The doubt I had is that in a real environment, where the operator is basically in the dark, maybe a blue thin border is not quite visible with a quick glimpse at the console. And since the tap feedback might be important for some people, I thought  maybe the whole button could blink.

But nevermind, let's go with the least impacting solution first.

As for the minimum blinking time, I'd say 125ms is a reasonable time, but as I said, we need to test with as many controllers as we can. Maybe for some other 200ms is better. In general though I don't like magic numbers in the code.

siegmund42:
I can follow your fear, although I have never had an environment where the blue borders weren't enough.
Anyway, this should be a simple change as soon as it's clear how the coloring should be.

> As for the minimum blinking time, I'd say 125ms is a reasonable time, but as I said, we need to test with as many controllers as we can.

I'll implement this in the next days and set up a post in the forum so maybe some people could help testing.

> Maybe for some other 200ms is better. In general though I don't like magic numbers in the code.

Me neither but I'm afraid there is no other possibility in this particular case.

enbyted:
I've used QLC in many different environments and hardware and never had any problems with blue borders. Even on very small and low contrast screens (like 6" old LCD displays, they're terrible!).

>>Maybe for some other 200ms is better. In general though I don't like magic numbers in the code.

>Me neither but I'm afraid there is no other possibility in this particular case.

This sounds like a good place for config or command line option.

BTW. I'm sorry for abandoning my work, as much as I love this project I just cannot find the time :(

siegmund42:
> This sounds like a good place for config or command line option.

I would definitely not go with a command line parameter, but probably with a config file option.
Maybe the flashing style could be adjusted in the config file, too.

ui/src/virtualconsole/vcspeeddialproperties.ui
```diff
@@ -221,10 +231,44 @@
          <property name="title">
           <string>Tap</string>
          </property>
-         <layout class="QHBoxLayout" name="m_tapInputLayout"/>
+         <layout class="QHBoxLayout" name="m_tapLayout">
```
mcallegari: Remove ?

siegmund42: Sorry - I knew I forgot something.

ui/src/virtualconsole/vcspeeddial.h
```diff
@@ -24,6 +24,7 @@
 #include <QSet>
 
 #include "vcwidget.h"
+#include "speeddial.h"
```
mcallegari: No need for this. It's already in vcspeeddial.cpp

mcallegari:
Alright, I've tested it with a BCF2000, with an AKAI APC mini and with Touch OSC.
While this works pretty good with a MIDI controller, it didn't convince me with a network controller.
Basically there is another (magic number)[https://github.com/mcallegari/qlcplus/pull/929/files#diff-6d1bfd87700ec896f5e47cde5bfa7a31R292] which can make the difference.
Over the network (especially WiFi), packets don't have guaranteed timings, so Touch OSC looses half of the feedbacks times below 1 second, because the on/off sequence is too fast.

For now I'll merge this, but I need to investigate if there's something that can be done for network plugins.

siegmund42:
That is a problem indeed. I'll think about it, too so maybe we can find a way to make everybody happy.
Anyway, thanks for merging!

PR #930, opened 2017-01-27
Basic page shortcut functionality

siegmund42:
I took the time and tried to process a recent request for direct page shortcuts, mentioned for example here: http://www.qlcplus.org/forum/viewtopic.php?f=18&t=9971&p=43621

Since I am not experienced in cpp programming at all I learned a lot in the past few days and did a lot of copy&paste from the vcspeeddialpreset class.

So far I got it working pretty good, but I got stuck with an issue regarding key sequences:
As soon as you switch the page manually (by selecting from dropdown -> not by key sequence) the key sequence is working no more until you click into the frame, then everything works normally.
I am not providing a sample workspace since I haven't implemented XML saving/loading yet. Just add a frame in the vc, enable multipage, insert >1 frame and define a key sequence for any external input.
So this seems definitely related to the qComboBox but I can't figure out how to workaround this. Maybe it is this issue: https://www.sencha.com/forum/showthread.php?170131-how-to-remove-focus-from-combobox-on-select Which, if still present, would mean the dropdown widget will not work in our case ):

What's not yet implemented:
 - XML load/save
 - feedback
 - styling

Please let me know if you want me something basic to change or if you are not willing to merge this.

Have a nice weekend!


siegmund42:
I was able to fix the problem with key sequences and focus by simply focus the frame after a page change.

Qt4 build should hopefully also pass now...

siegmund42:
I added basic XML load and save. There is one thing I was not able to resolve so far:
When loading from XML, there is some strange bar displayed at the top left corner of the frame. It goes away if you accept the properties-dialog once. I have no idea what this is...

In addition to that, the styling of the` QComboBox` is quite basic. I was not able to remove the padding at the top and bottom of the popup so far.

TODO:
 - implement/test feedback
 - Think about what to do on collapsing the frame
 - Final styling

siegmund42:
So this should be ready to merge now.

Since I do not have my BCF 2000 at hand, I was not yet able to test feedback but it should work.

The styling of the popup of the QComboBox is still very basic but I was not able to do better.

I thought about what to do on collapsing the frame and I came to the conclusion that nothing should be done. Since that the external inputs for prev/next page are still working, the external inputs for shortcuts should still work, too. The drop down should not be a problem at all.

I agree that the whole thing with drop down lists gets a bit complicated when it comes to a lot of pages, but I don't think this is a typical use case for the page shortcuts. In addition to that, most people will only use the external input or could produce "real" buttons for every page by using loopback.

janosvitok:
Yesterday I quickly tried your PR. I didn't find any problems yet. I really appreciate it, since it's something I wanted for long for my BCF2000 (use one row of buttons to switch pages). I will check more with actual BCF later.

One more addition: it may be helpful if you could add names to pages ("Front Lights", "Washes", "Movers", etc.)

siegmund42:
I appreciate hearing your feedback!

> One more addition: it may be helpful if you could add names to pages ("Front Lights", "Washes", "Movers", etc.)

Great idea, I'll give it a try.

siegmund42:
Finally implemented names for pages along with some smaller fixes and improvements. Enjoy!

mcallegari:
I see PRs are piling up, but ATM I'm coding a quite big feature for QLC+ 5 so I don't really have time to review and test all the PR, sorry.

If you have the chance, can you please post a screenshot of the change (either property dialog and how the frame looks like) so I can see it without building the PR ? (also in the Speed Dial PR if you can)

The only downside I see for now is that I need to port all these new features in QLC+ 5 :cry:

siegmund42:
Ok, I'll try to summarize what has been done in here:

1. The page label on the top right of frames was replaced by a `QComboBox`, so now it opens a drop-down list when clicking it. From this list, it is directly possible to select a page to jump to.
screenshot - 11 02 2017 - 23 43 38
screenshot - 11 02 2017 - 23 45 07

2. In addition to that, the caption of frames was changed to show the caption and (if specified) a special page name, which can be chosen in the properties dialog of the frame.
screenshot - 11 02 2017 - 23 47 32

3. In the pages tab of the properties dialog of a frame, there is a new section regarding pages. It includes another dropdown menu to select the page to edit, a text input to fill in a name for every page and an input selection widget. With this, it is possible to assign an external input (or key sequence) as shortcut for a page.
screenshot - 11 02 2017 - 23 50 07
On every page change, there is also some feedback sent to the assigned input (will test this tomorrow with external controller).

I used the speeddial widget and its presets as template for my implementation, so it is very similar to that.
If you want to know some more technical details in a short review, please let me know.

The only thing that I am not completely happy with at the moment, is the styling of the drop down of the page label. It has some odd padding on top and bottom, which I was not able to sort out. Nothing really bad, but not as perfect as I wanted it to be.

PS. This is the workspace I used for testing
[shortcuttest.qxw.zip](https://github.com/mcallegari/qlcplus/files/768977/shortcuttest.qxw.zip)

PPS. Thought you use linux so I packed a .deb into a .zip - maybe you could give it a short try.
[page-shortcut.zip](https://github.com/mcallegari/qlcplus/files/768975/page-shortcut.zip)

siegmund42:
Tested with my BCF 2000 and feedback works very well!

siegmund42:
Forgot to mention:

> The only downside I see for now is that I need to port all these new features in QLC+ 5 :cry:

I absolutely see your point, but I would feel bad if this prevents new features being added - especially since there is no time frame for v5 release yet.
Anyway, if you need help porting things I'm only a message away!

mcallegari:
Thanks for the screenshots and explanation. It makes easier to review this.

Correct me if I'm wrong, but are you really creating a QPushButton for each page ?? Why ?
I'm referring to [this line](https://github.com/mcallegari/qlcplus/pull/930/commits/42274a7e585d1aaa9abd025c3f347e541c023c3d#diff-9d3bf91cd0d096946d2609967398f462R456)
I mean, pages are clearly an array, so a QList or a QVector is just enough to handle them.
If you copied the code from SpeedDial presets, then you missed the point that presets are actual items in the widget space, so they are represented by a button or a knob.
In this case you already have your widget which is the ComboBox, where the clicked item index is exactly the frame page index.

siegmund42:
You are absolutely right. I copied the code and thought I'll need the button for external input or something later but indeed it is not needed at all. Thank you for the hint, I'll remove it.

mcallegari:
Wow, so fast ! :)
Looks OK now, but please give me some more time to check it in details.
Lately I struggle to manage my QLC+ activities time...too many things to follow !

mcallegari:
ehm...clicked on the wrong button lol

siegmund42:
> Looks OK now, but please give me some more time to check it in details.

No problem - since it's quite some code being changed/added it should be reviewed carefully. If you need some additional info or help, please let me know.

mcallegari:
I am finally picking this up, and I'm afraid this is not OK.

1) on Kubuntu 16.04 1920x1080 resolution, QLC+ in italian looks like this:
image
image
The QComboBox stylesheet is not OK. Can't imagine on Windows or macOS how it looks like.

2) more serious, loading an existing workspace with frame pages doesn't fill the combo box. Load the attached
[testFrameMultipage.zip](https://github.com/mcallegari/qlcplus/files/801079/testFrameMultipage.zip)

For now, these alone are a no-go.
I might add some more comments specific to the code

mcallegari:
I have almost given up with this. It is taking me too much time to fix it.
Please test with my latest changes and see if it still works.

mcallegari:
I think I'm done. I spent like 4 hours to get this straight and to be honest it shouldn't be like this with a PR.
There was quite a lack of knowledge of the Qt specifics and I hope @siegmund42 you understand how much difficult it is to make changes to QLC+ and what is the degree of responsibility I take for every change that goes upstream.

Now please let's test this and see if it can finally be merged.

siegmund42:
Hey Massimo, I'm glad to see things going in here.
Actually you are right - this was my first touch with Qt ever and I'm really sorry that you where the one who had to deal with it. Although I've learned a lot while coding this, of course it was only a short view and I will keep looking into it.

These are the issues I see after a short test:

1. Styling of QComboBox: It is still not perfect. Especially, there is a problem with long page names. The name is cut in both the actual item of the box and in the list view (see pictures).
screenshot - 26 02 2017 - 00 00 16
screenshot - 26 02 2017 - 00 00 50
For this reason I initially decided not to show the page name in the QComboBox. Only the page number is more predictable. Anyway, different languages (and so different lengths of texts) should be taken into account here.
Another reason I didn't want to show (only) the page name is that if the page name is empty, the corresponding entry in the QComboBox is empty, too. This especially will happen with any workspace that has no page names (so at least any which has been saved before the change). I don't think this is good practice since it is not even clear where to click on.
screenshot - 26 02 2017 - 00 10 09
Because of this, I recommend to at least add a leading "1: " to the page name.
Let me know what you decide to put in the Box and I'll code it.

2. Properties of selected page are not filled in after bringing up properties dialog.
screenshot - 26 02 2017 - 00 11 48
After changing the QComboBox in the dialog everything works as expected.

3. There is definitely something wrong with saving/loading. After every save/load, the order of the page names seems to be shuffled.

4. Key sequences are not saved. After changing the key sequences for a shortcut, the frame is still listening to the old key sequences.
It could also be the case that this is related to 3 and there is already something wrong with the internal storage order.

I did not test the shortcut functionality with loopback yet.

I had a short look into 3 but was not able to find out whats going on.

I'll have a look at all these in the next few days and let you know.
Sorry again for the messy request and I really appreciate your work as well as I understand that integrating a new feature is always quite a big step along with a huge responsibility.

mcallegari:
Issues are very possible after my changes. If you are so kind to share here a zip file with a test project I can use to test shortcuts+input+key sequences, it would save me some time to fix it.

As for 1, the reason is simple: the page name label width is fixed to 100 pixels. Please keep in mind that there's also the collapsed state, which is 200px, if I remember.
Basically, one can't have all. If they set a very long name, there should be a limit somewhere, and most likely that limit is already there with the 100px.
We can study a more flexible solution, but for now let's keep it simple and make this work.

siegmund42:
I'm pretty sure now 3&4 are related to a wrong internal storage order.
When looking into the xml code of the workspace, everything seems to be fine: Page with index 0 has the Name "Page 1 Name", the key sequence 1 and loopback input channel 0. And so on for all pages.
After opening the workspace in QLC+ and bringing up the properties dialog for the frame, you'll see issue 2 and if you have a look at the drop down, the pages are shuffled. The mapping is still correct but the order of the pages (items) in the QComboBox is not - most likely this is the only issue to resolve. Fix could be to ensure the correct order after adding a shortcut.
[shortcuttest.qxw.zip](https://github.com/mcallegari/qlcplus/files/803484/shortcuttest.qxw.zip)

Regarding 1, it's okay for me to have a limit for the page name. The only thing that in my opinion really needs to be changed before merging is to add at least a number in front of the page name, otherwise the whole QComboBox drop down menu would be empty (click on the QComboBox in the following workspace and you'll see what I mean).
[shortcuttest2.qxw.zip](https://github.com/mcallegari/qlcplus/files/803490/shortcuttest2.qxw.zip)

The same applies to the QComboBox in the properties dialog.

mcallegari:
@siegmund42  Please give it another go.
It should be almost working. I need to check input sources.
Please trash shortcuttest2.qxw cause I think you saved it after my changes and the shortcut order got messy

siegmund42:
It seems to work well now. Thank you for all the work!

Only one suggestion on your last commit: Honestly at first the slot was completely redundant, but after some time I discovered a problem, that key sequences are not recognized as long as the QComboBox has the focus. So I used this slot to set the focus back to the frame and didn't drop it.
You can reproduce the issue with the shortcuttest.qxw. Just pick a page using the QComboBox and then try to use key sequences to change the page again - it will not work.

> Just a generic suggestion for the next times: I noticed you placed tr("Page ..") in several places of your code. That means a lot of work for the translators, and a lot of redundancy which can cause inconsistencies sometimes.

Thank you for the useful hint! Just out of curiosity: is there any solution other than hiding the translation somewhere? I don't know much about the translation mechanism (especially how to trigger it to look for new untranslated phrases). If you have a link where I could learn about the translation thing I really would appreciate your sharing.

janosvitok:
@siegmund42: check http://doc.qt.io/qt-5/internationalization.html
the command to update translation files is lupdate (it is mentioned in several places in the above document. e.g. in http://doc.qt.io/qt-5/linguist-programmers.html

mcallegari:
Focus issue should be fixed now.
Merging on master

siegmund42:
@janosvitok Thank you for the hint!

@mcallegari Works great now. Thanks for merging - I'm sure a whole bunch of users will highly appreciate this.

mdmayfield:
There appears to be a regression in this commit - for Frames which have no pages, their titles are not shown at all in Virtual Console.

You can see this by creating a new project, creating a Frame and giving it a name -- or by opening any existing project with named Frames in it. After enabling Page mode, the name shows up in the header of the Frame.

I'm looking into the code now to try to help find the source of  this issue.

Issue #933, opened 2017-01-29
Web console password protection


enbyted:
It would be nice if there was an option to protect web console with basic auth, nothing fancy and really secure - just to stop curious eyes.

I'll work on in, but would love to hear comments from others - how would you see it in the web ui?

My personal rationale: I'm using QLC+ in my company for controlling architectural lighting and web console is great tool for that. More often than not clients put these drivers in the same network that clients can access, that creates a risk that someone might try to be funny and for example changes the IP address or plays with simple desk. I would live it if it made it's way into master, cause I don't really have the time to figure out ARM builds myself.

siegmund42:
Such a feature would be great. But still in most cases the user don't want authentication by default so it should be adjustable somewhere - maybe another command-line-parameter?
Maybe it could be done by configuration of the web server (I don't know how the web server is implemented at the moment) to require an authentication.

enbyted:
I though about adding tab in settings that would allow setting of username and password and if they are empty don't require authentication.

I like the idea of overriding credentials from command line, might be useful.

mcallegari:
Sorry, but I have no plans in taking an adventure through this.
And by the way, as explained before, features request go through the forum, not here

mcallegari:
Implemented by #980 ! :+1:

PR #934, opened 2017-01-29
Webaccess improvements

enbyted:
I'm opening this PR to start a discussion about web console.
It is a great idea, but still requires a lot of improvements.

I'm starting this with fix for sliders in inverted and percentage modes.
Test console for this one: http://pastebin.com/effRheTr

As stated in the commit I see these possible improvements for sliders:
- Properly using min/max values to span the whole slider
  range from min to max, not 0 to 255.
- Sliders in 'knob' mode - they should use some other input
  method, maybe a simple numeric up/down (or textbox) ?
  That might be useful for someone.
- It would be nice to have horizontal sliders, but that also
  means implementing them in VC, so probably won't be done.

I'll also work on #933
What other things should be prioritized as far as webaccess is concerned?
Personally I would love to have speeddial implemented, at least in some way. I would see it as just the presets + display of current value. Maybe add multipliers, though I don't think they fit usecases for web console.

What are your thoughts?

### DIFF
--- a/webaccess/res/websocket.js
+++ b/webaccess/res/websocket.js
@@ -52,9 +52,10 @@ window.onload = function() {
     }
   }
   else if (msgParams[1] == "SLIDER") {
+    // Slider message is <ID>|SLIDER|<SLIDER VALUE>|<DISPLAY VALUE>
     obj.value = msgParams[2];
     var labelObj = document.getElementById("slv" + msgParams[0]);
-    labelObj.innerHTML = msgParams[2];
+    labelObj.innerHTML = msgParams[3];
   }
   else if (msgParams[1] == "CUE") {
     setCueIndex(msgParams[0], msgParams[2]);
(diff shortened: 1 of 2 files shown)


mcallegari:
> I'm opening this PR to start a discussion about web console.

A pull request is not meant to start a discussion. The QLC+ forums are.

> It is a great idea, but still requires a lot of improvements.

In fact there's quite a long TODO list even in the documentation (see last paragraph): 
http://www.qlcplus.org/docs/webinterface.html

> Sliders in 'knob' mode - they should use some other input method, maybe a simple numeric up/down (or textbox) ? That might be useful for someone.

Not quite. Knobs should look like knobs.

> It would be nice to have horizontal sliders, but that also means implementing them in VC, so probably won't be done.

That'll come with QLC+ 5

mcallegari:
Thanks for the fixes

enbyted:
I'm just a very frequent user of web interface, so I thought I'd participate :>
I'll definitely work on some of the stuff from the list you've linked!

PR #936, opened 2017-02-05
VC: Fixed frame's keySequences not copied

siegmund42:
This fixes the problem, that when copying a frame in vc, the keySequences for enabling and previous/next page are not copied.


mcallegari:
Thanks ! :+1:

PR #937, opened 2017-02-05
Fixed no feedback from widgets in disabled frames

siegmund42:
This fixes the issue reported here: http://www.qlcplus.org/forum/viewtopic.php?f=33&t=10796
I'm still not sure if the use case is intended by QLC+, so if a disabled frame should send any feedback. But if it is considered as expected, we could simply use the visibility status of a widget in addition to the `acceptsInput` method (still needed for collapsed frames, in which the widgets are hidden).

### DIFF
--- a/ui/src/virtualconsole/vcwidget.cpp
+++ b/ui/src/virtualconsole/vcwidget.cpp
@@ -673,7 +673,7 @@ void VCWidget::sendFeedback(int value, QSharedPointer<QLCInputSource> src)
     if (src->needsUpdate())
         src->updateOuputValue(value);
 
-    if (acceptsInput() == false)
+    if (acceptsInput() == false && isHidden())
         return;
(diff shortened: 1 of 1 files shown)


siegmund42:
This also fixes that no feedback is sent to the enable button when disabling the frame. So I assume it should be merged since it is more near the current behavior, too.
Mentioned here: http://www.qlcplus.org/forum/viewtopic.php?p=46699#p46699

mcallegari:
Alright, let's merge this, even though I'm pretty sure it will make some users happy and some others not

siegmund42:
> Alright, let's merge this, even though I'm pretty sure it will make some users happy and some others not

Sure - but this applies to nearly every change. And this PR brought the behavior closer to the original one (before "Feedback from other pages" - fix).

PR #939, opened 2017-02-07
Implemented catching sliders feature

siegmund42:
As requested here: http://www.qlcplus.org/forum/viewtopic.php?f=18&t=10814&p=46597#p46597
It would be great if sliders could be "caught" by external controls, if they have no motor faders.

I tried to implement this. It depends on if there is a feedback patched to the universe - if not, the new behavior is applied.
To catch the slider, simply move the external control to the current level of the slider and the slider will follow your external control as soon as you reached a certain area around the current level with your external control.

There is one drawback: If you move very fast across the current value, the slider might not be caught, but I wouldn't consider this is a big problem.

### DIFF
--- a/ui/src/virtualconsole/vcslider.cpp
+++ b/ui/src/virtualconsole/vcslider.cpp
@@ -175,7 +175,7 @@ VCSlider::VCSlider(QWidget* parent, Doc* doc) : VCWidget(parent, doc)
 
     m_resetButton = NULL;
     m_isOverriding = false;
-    m_isCatched = false;
+    m_isCaught = false;
 
     /* Bottom label */
     m_bottomLabel = new QLabel(this);
@@ -778,6 +778,8 @@ void VCSlider::slotResetButtonClicked()
 
     m_priority = DMXSource::Auto;
     m_doc->masterTimer()->requestNewPriority(this);
+    setCaught(false);
+
     emit monitorDMXValueChanged(m_monitorValue);
 }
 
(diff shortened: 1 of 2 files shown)


siegmund42:
Works well with a BCF 2000.
With the last commit, drawback when moving fast across the value is resolved.

xhpohanka:
Hi,
I was again around qlc and noticed that this PR has not been merged yet. Is there any problem here? I would really appreciate this function...

thanks all

siegmund42:
Since someone has again requested this feature [here](http://www.qlcplus.org/forum/viewtopic.php?f=33&t=11658&p=50004#p50004) I would like to dig this up. I think there are a lot of users who would be happy to see that feature (at least from 4 I know).
I'm happy to change the PR to resolve conflicts or other issues if you are willing to merge.

mcallegari:
@siegmund42  Implemented my way: 29ec50d51fca1cec6c9a978ac587c2670593d090
It has less impact than your patch. Please check if everything is OK. 
I'll leave this open for discussion until the feature is OK.

siegmund42:
Thanks for the repsonse! I replied in the [forum thread](http://www.qlcplus.org/forum/viewtopic.php?f=33&t=11658&p=50072#p50072), there may be some issues left.

mcallegari:
@siegmund42 thanks for the loopback example
I think I've fixed the remaining cases here 3e556d0fc116bb60170926918c067c2c4f22773e
The problem was the 'surpassing' case (for fast values change) that did not consider page switch (show/hide) and disable state.
Please check if everything is fine now, thanks

siegmund42:
Thank you for fixing! Seems to work fine now - closing.
Thanks again, I'm sure there will be lots of users who are appreciating this!

PR #941, opened 2017-02-14
left-align blend and link buttons in vccuelist's xfade section

edogawa23:
cosmetic change to make these two buttons look a bit tidier


janosvitok:
This is how it looks:

Before | After
----------|-----------------
before | after

PR #942, opened 2017-02-14
VC/Cue List Widget: Fixed sliders Mode not copied

siegmund42:
This fixes the issue that the sliders Mode status is not copied.
Reported here: http://www.qlcplus.org/forum/viewtopic.php?p=46703#p46703


janosvitok:
Thanks!

PR #944, opened 2017-02-15
Set fixed size policy for XFade percentage buttons

edogawa23:
Whenever one of the two Qlabels (sorry, I wrote buttons in the commit message by mistake) that display fader movement in percent changed to show "100%", the list itself moves a bit to the right because the grid layout changes width for that column. This patch sets both Qlabels to a fixed size of 32x24 pixel, just big enough to show the 100% mark, and thus stops that nervous jittering around.

### DIFF
--- a/ui/src/virtualconsole/vccuelist.cpp
+++ b/ui/src/virtualconsole/vccuelist.cpp
@@ -333,6 +333,9 @@ bool VCCueList::copyFrom(const VCWidget* widget)
     setPlaybackKeySequence(cuelist->playbackKeySequence());
     setStopKeySequence(cuelist->stopKeySequence());
 
+    /* Sliders mode */
+    setSlidersMode(cuelist->slidersMode());
+
     /* Common stuff */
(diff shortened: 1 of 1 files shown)


mcallegari:
Have you tested this on Windows, Linux and macOS ?
Since I guess the answer is no, the thing is that every OS has a different default font, which has different metrics.
Setting a fixed pixel size is usually not the way to go with texts.

The proper way to do this is to pre-calculate the width of the "100%" text with QFontMetrics and then set the label width to the result.

edogawa23:
OK Massimo, I'll look into this and propose another patch;

I wondered about this, in fact, but had no clue how to deal with it, the only test i ran was to change fontsize in Linux, and it seemed to look ok at sizes up to 14 pt. (I'm using Tahoma on my PCs, btw.). When going massively bigger like e.g. 28 pt., text got cropped, but that happened just about everywhere in that widget, somewhat expectably, so I thought I could leave it as-is... anyway, I'll find time to look into QFontMetrics later today.

PR #945, opened 2017-02-16
Fixed input profiles not honored on other page than 1

### LINKED ISSUE #940: Input Profiles not honored on other than Page 1
opened for tracking:
http://www.qlcplus.org/forum/viewtopic.php?f=33&t=10826#p46640

siegmund42:
This should fix #940 
I have no external controller at hand atm and don't work with input profiles so much so somebody should test if this breaks something else beforehand.


mcallegari:
Merging. Let's hope this doesn't cause side effects.
Every time there's a change on VCWidget I feel a breeze along my spine... :worried:

PR #946, opened 2017-02-17
VCCuelist: fixed width for Crossfader top labels, honouring current font settings

edogawa23:
Second attempt at a patch to stop the cuelist frame jittering when a crossfader reaches 100%, taking current font metrics into account

### DIFF
--- a/ui/src/virtualconsole/vccuelist.cpp
+++ b/ui/src/virtualconsole/vccuelist.cpp
@@ -333,6 +333,9 @@ bool VCCueList::copyFrom(const VCWidget* widget)
     setPlaybackKeySequence(cuelist->playbackKeySequence());
     setStopKeySequence(cuelist->stopKeySequence());
 
+    /* Sliders mode */
+    setSlidersMode(cuelist->slidersMode());
+
     /* Common stuff */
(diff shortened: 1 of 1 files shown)


ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -112,8 +114,11 @@ VCCueList::VCCueList(QWidget* parent, Doc* doc) : VCWidget(parent, doc)
     m_linkCheck = new QCheckBox(tr("Link"));
     grid->addWidget(m_linkCheck, 1, 0, 1, 2, Qt::AlignVCenter | Qt::AlignLeft);
 
+    QFontMetrics m_fm = QFontMetrics(this->font());
+    
     m_sl1TopLabel = new QLabel("100%");
     m_sl1TopLabel->setAlignment(Qt::AlignHCenter);
+    m_sl1TopLabel->setFixedWidth(m_fm.width("100%",-1));
```
mcallegari: You don't actually need to pass -1 as argument as it is the default.
C++ allows to omit them.
Same below

edogawa23: OK I'll change that...

PR #952, opened 2017-03-05
Fix vcframe label

mdmayfield:
The new page name functionality caused the frame names not to show when not in multipage mode. This change uses the previous behavior for setting the label text when multipage mode is not enabled.


mcallegari:
Matt, you're doing a mess here !
Please learn how to submit pull requests on GitHub first, cause you are proposing 3 different patches committing all of them in the same branch.
You need to create 3 branches on your end, and separate changes.

I'm gonna close all your PR until you submit them properly.

PR #954, opened 2017-03-05
Added Lixada-12-LEDs PAR fixture

scemama:


mcallegari:
1- Lixada is not even a lighting maker. It's probably chinese stuff re branded
2- the definition doesn't even pass the [online validator](http://www.qlcplus.org/fixture_validator.php)
3- there's a forum to [submit fixtures](http://www.qlcplus.org/forum/viewforum.php?f=3) and unless they're perfect (so that I just need one click to merge them) that's the place to use

Closing.

PR #956, opened 2017-03-12
Enable Alt-key shortcuts on Mac OS

mdmayfield:
By default, Qt does not enable standard Alt-key mnemonics on MacOS. This commit lets QLC+ on MacOS respond to Alt-key shortcuts the same as on other platforms.

This change also allows MacOS users to type the first letter of a dialog box button, the same as on Windows and Linux - for example, Y for "Yes" to "There are still running functions. Stop them?".

### DIFF
--- a/ui/src/app.cpp
+++ b/ui/src/app.cpp
@@ -64,6 +64,10 @@
 #   include "hotplugmonitor.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


mcallegari:
Looks OK, thanks :+1:

PR #957, opened 2017-03-19
Copy crash

siegmund42:
This fixes a crash when trying to copy a multipage frame as reported here: http://www.qlcplus.org/forum/posting.php?mode=reply&f=33&t=10948


siegmund42:
Sorry, too many commits - closing

PR #958, opened 2017-03-19
VC/Frame: Fixed crash when copying multipage frame

siegmund42:
This fixes a crash when trying to copy a multipage frame as reported here: http://www.qlcplus.org/forum/posting.php?mode=reply&f=33&t=10948


mcallegari:
Thanks :+1:  and sorry for the delay.
I'm full immersed in the QML world :wink:

PR #959, opened 2017-04-05
Looped audio

raivisr:
I needed support for perfectly looped audio samples being played in infinite loop until explicitly stopped, so I created it. Someone else might find it useful.

### DIFF
--- a/engine/audio/src/audio.cpp
+++ b/engine/audio/src/audio.cpp
@@ -244,15 +244,15 @@ void Audio::adjustAttribute(qreal fraction, int attributeIndex)
 
 void Audio::slotEndOfStream()
 {
-	if (m_audio_out != NULL)
-	{
-		m_audio_out->stop();
-		m_audio_out->deleteLater();
-		m_audio_out = NULL;
-		m_decoder->seek(0);
-	}
-	if (!stopped())
-		stop(FunctionParent::master());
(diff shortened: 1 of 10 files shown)


mcallegari:
Please fix the C++ code indentation. It doesn't comply with the existing code.

raivisr:
Is that only issue/potential issue you see with the code?

mcallegari:
> Is that only issue/potential issue you see with the code?

I haven't reviewed it in depth yet. I just had a quick look and noticed the indentation.
By the way, what is the result of placing an Audio function into a looping Chaser ? Isn't that precise enough ?

raivisr:
I am lazy, so I did try looping chaser first, but there are two issues, 1) unless you set step duration to exactly the length of audio, it jumps over audio immediately 2) there is long delay. while audiorenderer thread is created and loop is not seamless anymore.

mcallegari:
Alright, I've reviewed it.
It looks OK, but if you follow what I've done in the Video class for looping videos, you will see that I used the existing [setRunOrder](https://github.com/mcallegari/qlcplus/blob/master/ui/src/videoeditor.cpp#L202) method.
If you don't mind, I'd do the same for the Audio class. Otherwise I'll merge this and I'll do it myself.
Thanks

raivisr:
I'll fix that.

PR #961, opened 2017-04-08
Updated catalan translation (UI)

sbenejam:
Updated some new strings


janosvitok:
Thanks!

janosvitok:
It seems that translations with "vanished" are not needed anymore. See http://stackoverflow.com/questions/31808275/vanished-type-for-the-translation-element-in-the-ts-files.

I'm not sure what to do here. I guess next run of lupdate will bring back the type="vanished".

As I understand it: These translations are not a problem but probably we can remove them. Only point why we'd want to keep them is when a similar phrase appears somewhere else.

mcallegari:
My bad I guess. I always forget to do "lupdate -no-obsolete". That should fix the vanished thing

PR #964, opened 2017-04-26
added a checkbox to enable webaccess direct from the osx launcher. also added information about power consumption and weigth to the workspace and single universes

chrisderock:
with this change osx users can start the webaccess via the launcher


engine/src/fixture.cpp
```diff
@@ -1236,3 +1236,23 @@ QString Fixture::status() const
 
     return info;
 }
+
+int Fixture::getPowerConsumption()
+{
+  if (m_fixtureMode != NULL)
```
mcallegari: 4 spaces indentation please

ui/src/fixturemanager.cpp
```diff
@@ -866,6 +866,11 @@ void FixtureManager::initActions()
                                tr("Remap fixtures..."), this);
     connect(m_remapAction, SIGNAL(triggered(bool)),
             this, SLOT(slotRemap()));
+
+    m_wsInfoAction = new QAction(QIcon(":/info.png"),
```
mcallegari: You forgot to commit the png file...

ui/src/fixturemanager.cpp
```diff
@@ -1719,6 +1725,22 @@ void FixtureManager::slotExport()
     file.close();
 }
 
+void FixtureManager::slotWsInfo()
```
mcallegari: Indentation

chrisderock:
Sry, its my first pull request ever.. i only wanted the checkbox to be integrated but the request contains all of my commits..

chrisderock:
about the workspace information.. of course we can add the information to single universes too but for packaging i dont load only a single universe to one truck so i need the complete weight. for the next version you suggested more then only fixtures and i think trussing & co. have also weights

chrisderock:
Back to the original topic: for windows user you can use

CreateShortCut '$SMPROGRAMS\$R0\Q Light Controller Plus (webaccess).lnk' $INSTDIR\qlcplus.exe "--web"

i have no windows and cant test this, so i write it ad comment...

mcallegari:
Sorry, I did it my way.
Please check this out 02e91c5d529542ae367171ed50dc6d96a32d6d25

As for macOS options, I'll need more time to think about it. If we add a web interface option, then we should add all the other QLC+ options.
In general I would like to get rid of the QLC+ launcher and give the possibility to launch QLC+ or the Fixture Editor directly, but I haven't figured out a proper way to do it yet.

PR #965, opened 2017-05-01
adding support for APC 40 MK II

sandinak:
- has correct init string for Mode 2


janosvitok:
Manual (APC40Mk2_Communications_Protocol_v1.2.pdf) confirms the message. I can't test it more than that.

PR #877, opened 2016-08-10
Alternatespeed

plugz:
This could be split in 2 pull requests:

- fix #758, #445, #193.
- introduces chaser step speed change through VCSpeedDial.

Details in the commit messages.

Issue #445, opened 2015-01-29
[Maybe a Bug] Chaser applies Fade-In of parent Chaser, ignoring his own values


Moehritz:
To start off, I´m now trying to describe my Problem as good as I can, just throw your questions at me in case you´ve got one.

So what my basic setup is:
Chaser main
 | Step 1: Collection x (Fade in set to .20)

Collection x
 | Contents: Chaser inner; Scene BackgroundLight

Chaser inner
 | Step 1: Szene abc (Fade in set to 10 sec)

By starting the inner chaser, scene abc will fade in within 10 seconds as intended. But when starting the main chaser, it will start the collection, wich will start the inner chaser - and scene abc will fade-in within 0.20 seconds - not 10 seconds as it would be logic. Removing the fade-in of the main chaser isn´t a solution for me, as I want the scene BackgroundLight to fade in with the start of the inner chaser.

To make things clear, I´ve created a small example project:
https://gist.github.com/Moehritz/35e1603f5dd32b66f55d

Are there any other workarounds for this? It would be really cool if anyone could help me out, thanks!

forresthopkinsa:
Right now, I've just had to set all of my top-level chasers to "Default" fade-in and fade-out times. I, too, would love to see a solution.

I made a [Forum topic](http://www.qlcplus.org/forum/viewtopic.php?f=5&t=8974) for this a while back.

matthijskooijman:
I'm running into the same problem, but with fade outs. I have a similar setup as described above (but with zero fades in the inner chaser for a disco chaser) and want to gradually fade out that one by defining a fade out in the outer chaser, but that causes the inner chaser to become fading rather than instantaneous.

Issue #193, opened 2013-11-07
Chases in cuelists bugs


NiKoyes:
I upload a file for describing the bug and test it : http://dl.free.fr/rwgddTQi1

I created 2 dimmers
2 scenes : S1 := D1@255, S2 := D2@255
1 chaser : 
Chase := 
1. S1 .50s
2. S2 .50s
   no fades

Then, create a cuelist (chaser) : 
1. S1
2. Chase
3. S2

And test, while changing the cuelist fades

I notice a 1st bug (easier to see when no fade) if you step from 2 to 3 when D2=255 (during the chase) it jumps the 3d step to go back to the 1st

Other bug or real problem :
when you add fades in the cuelist, the chaser automatically gets the same values in it, but it should not according to me, and if you put more than 5s fade, the chaser nearly stops, instead of coming in fade and going out fade. (as it sould according to me...)

Hope explainations are clear, otherwise, get the file above, everything is clear in it, test with  the DMX monitor.

Thanks !

PS : system is ubuntu, tested w/ versions 4.5.0 & 4.5.1

PR #966, opened 2017-05-04
Alternate speeds

### LINKED ISSUE #758: RGB matrix fadein is broken when starting an RGB matrix from a VC slider
Because of this:
https://github.com/mcallegari/qlcplus/pull/699
(sorry)

This can be fixed by simply ignoring overridefadein and overridefadeout.
Or use them differently (use them as global fade in and fade out speed instead of step fadein and fadeout speed).
Or revert 699.

The issue is the same for chasers, and a fix could fix this issue too: https://github.com/mcallegari/qlcplus/issues/445

plugz:
This is basically a redo of #877, which fixes #758, #445, #193.
See #877 for info on the alternateSpeeds deal.

What's more in this PR, is that all the speed-related code is now separated from the Function class.
It is present in the FunctionSpeeds class, and the Speed class.
Speed contains the helper functions (add, toString, etc...) and a speed value.
FunctionSpeeds is the group of {fadeIn, hold, fadeOut} speeds.

This PR could be merge as-is, as it should not break the current version of QLC+, but it breaks the build of QLC+5.

What still needs to be done:
* fix qmlui build
* better integrate bpm stuff
* find a way to use the outerspeed
* add a "speed override priority" system


mcallegari:
@plugz Hi David, welcome back ! :)

Wow, this is a massive PR and I'm afraid I don't even know where to start reviewing it unless some more documentation is provided.
Most importantly: the more we document the QLC+ engine, the better. So, please document the new methods you added, especially to the Function class.
I can't really understand the difference between speeds / overrideSpeeds / alternateSpeeds /  speedsEdit / commonSpeedsEdit.
To fix QLC+ 5, I need to know exactly what I'm doing.

Also, I noticed you removed the totalDuration methods. Those have a precise meaning which is different from 'duration'. Have you checked that all the Show Manager features still work ? For example stretching a Chaser involved totalDuration.
Honestly I don't remember why I placed totalDuration in the Function class, and maybe it can be moved into ShowFunction. In any case we need it.

This for now. Thanks !

plugz:
Hello Massimo,

I had another feature to add in this PR, it's done now.

> Wow, this is a massive PR and I'm afraid I don't even know where to start reviewing it unless some more documentation is provided.

I'll now start adding comments everywhere, and I'll try to explain as best as I can here.

> To fix QLC+ 5, I need to know exactly what I'm doing.

I might try to fix it myself

> Also, I noticed you removed the totalDuration methods. Those have a precise meaning which is different from 'duration'. Have you checked that all the Show Manager features still work ? For example stretching a Chaser involved totalDuration.
Honestly I don't remember why I placed totalDuration in the Function class, and maybe it can be moved into ShowFunction. In any case we need it.

I'm pretty sure I broke the stretch feature, I'll look into it :)

PR #971, opened 2017-05-18
Fix for duplicate scene bug (thanks @janosvitok)

mdmayfield:


mcallegari:
Great catch, thanks ! :+1:

PR #974, opened 2017-05-26
VC/Frame(PageShortcuts): Fixed pages not loaded properly when loading…

siegmund42:
… workspace file that has been saved without shortcuts. This caused trouble in the number of pages available because there was no initial shortcut present.

### DIFF
--- a/ui/src/virtualconsole/vcframe.cpp
+++ b/ui/src/virtualconsole/vcframe.cpp
@@ -1058,6 +1058,8 @@ bool VCFrame::loadXML(QXmlStreamReader &root)
         else if (root.name() == KXMLQLCVCFrameMultipage)
         {
             setMultipageMode(true);
+            // add initial shortcut
+            addShortcut();
             QXmlStreamAttributes attrs = root.attributes();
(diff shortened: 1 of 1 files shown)


mcallegari:
Hmmm...this looks like a workaround instead of a solution.
2 lines below your change, there's this:
```
            if (attrs.hasAttribute(KXMLQLCVCFramePagesNumber))
                setTotalPagesNumber(attrs.value(KXMLQLCVCFramePagesNumber).toString().toInt());
```
setTotalPagesNumber should take care of adding the necessary shortcuts. If it doesn't, then the issue is somewhere around there.

mcallegari:
Here's the fix 22340b5b57ff6c4062fc40aff489b0589f4dd6fc

siegmund42:
Much better solution! But I would like to suggest to add a `resetShortcuts()` call before adding the first one to really be sure that there is only one present.

mcallegari:
I think there's no need.
The transition in setMultipageMode does a addShortcut on enable and a resetShortcuts on disable.
When loading the XML, multipage is disabled by default.
If we loose track of the shortcuts it means there's a bug somewhere in the code (and we'd need to fix that code), but I would exclude the XML part.

PR #975, opened 2017-05-26
VC/Clock: Added external inputs for controlling countdown/stopwatch

siegmund42:
As requested [here](http://www.qlcplus.org/forum/viewtopic.php?f=18&t=10923&p=47289) I added external input support for the clock widget in countdown and stopwatch mode.
I was not yet able to test feedback, but will this weekend.
I'm not completely sure about the order in the properties dialog but think as it is now is sufficient.

### DIFF
--- a/ui/src/virtualconsole/vcaudiotriggersproperties.ui
+++ b/ui/src/virtualconsole/vcaudiotriggersproperties.ui
@@ -7,14 +7,14 @@
 
   Copyright (c) 2015 Massimo Callegari
 
-  Licensed under the Apache License, Version 2.0 (the "License");
+  Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);
   you may not use this file except in compliance with the License.
   You may obtain a copy of the License at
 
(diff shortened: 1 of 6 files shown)


mcallegari:
Tested and it's working pretty well. Feedbacks seems to be OK too.
Well done !

I just took the liberty of changing the config page layout a bit cause when external input was enabled, the page got too crowdy. A tab solution is cleaner (e.g. see audio triggers widget)

siegmund42:
Tabs are much better, but I think it is a bit confusing that the Input tab is empty when selecting clock. Maybe the input widgets can be grayed out in this case?

mcallegari:
I agree, and I already tried to hide the input tab when selecting the clock mode.
Unfortunately Qt doesn't provide a way to just hide a tab. You need to destroy it and recreate it later.
I think the easier solution for now is to just display a text saying "No input controls available in clock mode"

mcallegari:
Done a135b63a3c50924aeb2bbce9c390a2fab4460ebc

PR #978, opened 2017-06-13
Add copy & paste buttons to speed dial functions page

jpxd:
I often have lots of functions connected to a speed dial in the virtual console where i want to control the fade times in the same way. To save a lot of clicking in the dropdown boxes I added a button to apply the factors from the selected function to the others (see screenshots).

bildschirmfoto 2017-06-13 um 11 11 54

results in 

bildschirmfoto 2017-06-13 um 11 17 57

### DIFF
--- a/ui/src/virtualconsole/vcspeeddialproperties.cpp
+++ b/ui/src/virtualconsole/vcspeeddialproperties.cpp
@@ -270,12 +270,12 @@ void VCSpeedDialProperties::slotRemoveClicked()
         delete it.next();
 }
 
-void VCSpeedDialProperties::applySelectedFactorsToAllClicked()
+void VCSpeedDialProperties::copySelectedFactorsClicked()
 {
-    QList <QTreeWidgetItem*> selectedItems = m_tree->selectedItems();
-    if (selectedItems.isEmpty())
+    QList <QTreeWidgetItem*> allSelected = m_tree->selectedItems();
+    if (allSelected.isEmpty())
         return;
-    QTreeWidgetItem* selected = m_tree->selectedItems().first();
+    QTreeWidgetItem* selected = allSelected.first();
 
     // clear focus on the combobox to apply the selected value
     QApplication* app = static_cast<QApplication*>(QApplication::instance());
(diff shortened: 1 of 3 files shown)


siegmund42:
Great and highly desired feature!
Tested on Xubuntu 14.04 and works great.
One small issue, which is most probably related to Qt's `QComboBox`:
When choosing a new value from the `QComboBox` and directly clicking the new "Apply to all functions" Button, nothing happens. You have to deselect the function and select it again to get it to work.

I would appreciate if this would be included in the next release!

ui/src/virtualconsole/vcspeeddialproperties.cpp
```diff
@@ -270,6 +270,43 @@ void VCSpeedDialProperties::slotRemoveClicked()
...
 }
 
+void VCSpeedDialProperties::applySelectedFactorsToAllClicked()
+{
+    QList <QTreeWidgetItem*> selectedItems = m_tree->selectedItems();
+    if (selectedItems.isEmpty())
+        return;
+    QTreeWidgetItem* selected = m_tree->selectedItems().first();
+
+    // clear focus on the combobox to apply the selected value
```
jpxd: @siegmund42 
This now clears the focus on the open combobox before the factors are copied.
Let me know if there is a more elegant way in QT to do this (i'm a qt noob)...

mcallegari:
Thanks for this patch.
Since we're at it, I was wondering if a Copy & Paste mechanism would be even more effective. In this way I could copy one Function factors to the items I select, not necessarily all of them.
What do you think ?

jpxd:
@mcallegari Good point!

screenshot 2017-06-18 22 20 19

screenshot 2017-06-18 22 20 41

mcallegari:
Excellent, thank you !
I'll review this one within this week and hopefully release 4.11.0 next weekend.
I'm not too convinced about the focus code block, but I'll probably merge this as it is and then see if there's a better solution

jpxd:
Nice!
Yes, the unfocus block is not nice but I didn't find a better QT-ish way to do it in a reasonable amount of time...

mcallegari:
Merged :+1:

mcallegari:
Found a better solution (I guess). See the comment in the commit.
Footnote: we use left aligned parenthesis and not at the end of a line.
Also, slots should start with the "slot" word, so they can be easily recognized within the code.

PR #981, opened 2017-06-22
Allow for multiple LibFTDI devices with the same VID, PID, and serial.

MDTech-us-MAN:
I bought a few cheap China USB to RS485 bridges ([here](http://www.dx.com/p/usb-to-rs485-module-black-221402)) they use FTD2xx chips but the only thing is that they all have the same exact VID, PID, and serial numbers. This caused a problem when I had a few connected at once. QLC always tried to open the same one. This is obvious because these 3 identical fields are the ones that were used to open the device (using `ftdi_usb_open_desc`).
This fix may or may not seem like a dirty one. Basically, what it does is add another property to the `DMXInterface` class (and `LibFTDIInterface`). Then, that property is used in the `open()` function to loop through the available devices until a device with matching [USB Bus](http://libusb.sourceforge.net/api-1.0/group__dev.html#gaf2718609d50c8ded2704e4051b3d2925) and [USB Device Address](http://libusb.sourceforge.net/api-1.0/group__dev.html#gab6d4e39ac483ebaeb108f2954715305d). I didn't want to use [Port Number](http://libusb.sourceforge.net/api-1.0/group__dev.html#ga14879a0ea7daccdcddb68852d86c00c4), because it is not guaranteed to be unique.

Maxwell.

### DIFF
--- a/plugins/dmxusb/src/dmxinterface.cpp
+++ b/plugins/dmxusb/src/dmxinterface.cpp
@@ -21,13 +21,15 @@
 #include "dmxinterface.h"
 
 DMXInterface::DMXInterface(const QString& serial, const QString& name, const QString &vendor,
-                           quint16 VID, quint16 PID, quint32 id)
+                           quint16 VID, quint16 PID, quint32 id, uint8_t usbBus, uint8_t usbAddr)
     : m_serial(serial)
     , m_name(name)
     , m_vendor(vendor)
     , m_vendorID(VID)
     , m_productID(PID)
     , m_id(id)
+    , m_usbBus(usbBus)
+    , m_usbAddr(usbAddr)
 {
 }
 
@@ -65,6 +67,16 @@ quint32 DMXInterface::id() const
     return m_id;
 }
 
+uint8_t DMXInterface::usbBus() const
+{
+    return m_usbBus;
+}
+
+uint8_t DMXInterface::usbAddr() const
+{
+    return m_usbAddr;
+}
+
 quint8 DMXInterface::busLocation()
 {
     return 0;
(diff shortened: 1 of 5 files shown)


MDTech-us-MAN:
FYI: continuous-integration/travis-ci/pr failed because of a problem in the runner: g++ could not be found.

mcallegari:
Hi, thanks for this PR.
On Linux this could be OK, but we need to find a proper way to handle identical devices on Windows and macOS too.
I might order 3-4 of those adapters and try this myself on all the supported OSes.

MDTech-us-MAN:
That would be great! Especially since these little adapters really are cost-effective! (Only $4.40 a piece against $65 for an Enttec DMX Open. These things win hands down.)

janosvitok:
Did you try to change serial number using [FT_Prog](http://www.ftdichip.com/Support/Utilities.htm#FT_PROG) utility?

Will the device bus/address be same next time you start the computer when you (un)plug other devices?
E.g. when the port will be used by more or less devices?

I'm afraid more devices with same serial number is against USB specification and may cause other problems (e.g. with drivers). https://msdn.microsoft.com/en-us/library/windows/hardware/hh998535(v=vs.85).aspx

mcallegari:
For the record I just ordered 3 of those adapters :wink:

MDTech-us-MAN:
@janosvitok - This adapter seemed to be locked or something causing the FT_Prog utility not to work.

janosvitok:
Several articles on the internet say that genuine FTDI chip costs $4 so when the whole board is cheaper than the chip, the chip is counterfeit. That may mean it does not have proper eeprom etc.

https://github.com/eswierk/ft232r_prog contains FT_PROG linux alternative.

MDTech-us-MAN:
@janosvitok - And its even cheaper from [AliExpress](https://www.aliexpress.com/item/USB-2-0-to-TTL-RS485-Serial-Converter-Adapter-FTDI-Module-FT232RL-SN75176-double-function-double/32687850884.html). Only $2.30... (though 30 day shipping) But still... the key for opening should be the bus and port, just in case, who knows what crap manufacturers could make. This is one of the reasons I added the `ftdi_usb_open_bus_addr` in the libftdi library (for another project).

mcallegari:
@MDTech-us-MAN I just received the 3 adapters that I ordered.
One of them is broken, but to my surprise, the other 2 are working and they have different serial numbers !
So basically they work together with no issue.

I suppose you got some of them from a bad batch, where they forgot to program the SN.
I haven't tried them on Windows yet, but I suppose everything will be fine.

MDTech-us-MAN:
@mcallegari So what do you think? Should this go in? I suppose that this may prevent collisions, but I dunno...

mcallegari:
@MDTech-us-MAN sorry but this is a no go.
Even if this PR is completed with the Windows code, I won't risk to compromise the plugin for some crappy chinese adapters. In other words: QLC+ requires adapters with different SN.

PR #985, opened 2017-07-08
Added support for RBG pixel panels

petermarks:
This is my first pull request on GitHub, so please advise if I'm not doing something right.

I have a chain of LED pixels using WS2811 chips. For some reason they are wired in RBG channel order. I've added support for RBG in pixel panels following the examples of other channel orderings. This change seems to work in my setup, so I thought others might find it useful.

Peter


mcallegari:
Looks good to me, thanks a lot :+1:

petermarks:
Thanks Massimo.

PR #997, opened 2017-08-26
Fix slider not returning to 0 on stopping of playback function

siegmund42:
This fixes a problem, that an external slider does not return to 0 after a playback function is stopped.

### DIFF
--- a/ui/src/virtualconsole/vcslider.cpp
+++ b/ui/src/virtualconsole/vcslider.cpp
@@ -898,6 +898,7 @@ void VCSlider::slotPlaybackFunctionStopped(quint32 fid)
             m_slider->setValue(0);
         resetIntensityOverrideAttribute();
     }
+    updateFeedback();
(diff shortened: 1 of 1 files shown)


ui/src/virtualconsole/vcslider.cpp
```diff
@@ -898,6 +898,7 @@ void VCSlider::slotPlaybackFunctionStopped(quint32 fid)
             m_slider->setValue(0);
         resetIntensityOverrideAttribute();
     }
+    updateFeedback();
```
mcallegari: wrong: should be inside the if clause

mcallegari:
see bc785aa976335dc58a782a002342147ed797ad3d

ui/src/virtualconsole/vcslider.cpp
siegmund42: of course, thanks for fixing!

PR #999, opened 2017-09-09
Fixture tool

janosvitok:
This PR adds a command line tool that re-saves fixture definition (for now).

It allows to normalize tag/attribute order, so that further edits in Fixture Definition Editor will not cause
unrelated changes (e.g. diffs will contain only the new changes).

It allows to convert qxf files when the format is extended and also possibly migrate updating FixtureMap.xml from ruby to C++.


mcallegari:
Is there a specific issue that requires this tool ?
It looks like a tool that a packager should launch before releasing QLC+, so I don't see a reason to build and install it so that the end user can actually launch it.
Instead it looks more like a "support" tool, that can be written in Python (or whatever) and placed in the fixtures folder and launched when needed.

janosvitok:
I added shell script so that install is not needed anymore.
I don't want to install nor package this tool (yet?). It's to be run on demand by developers/maintainers, e.g. at the same time when update-fixture-map.rb is run.

The idea behind it for now is that I want to explicitly bind LSB and MSB channels in the fixture definitions, and with that comes autodetect mechanism. I wrote one in C++ inside QLC+ and one in ruby in update-fixture-map.rb. Then I decided that I just want to use the C++ one so that 1. I don't have to write the same thing twice 2. I need to deploy the C++ to users so I need to test it.

This tool helps with testing in two ways:
1. first I use the tool to normalize the definitions, so they do not change when they are opened and saved in fixture editor/fixture tool again
2. second I will use the tool to run the MSB/LSB autodetect mechanism on all current fixture, so that I can check if it works properly (and the diffs will be much smaller after step 1)

This is MSB/LSB change: https://github.com/janosvitok/qlcplus/commit/e027eca5dcc385c387588957841fcdbf3b44a267
It does not merge cleanly to current master so I have to update it.

Similar thing could be done later with shutter/strobe values (add manual editor and use current guess mechanism to prefill the values).

mcallegari:
Sorry, but I still don't get the point.

- Again, is there a particular issue that requires 557 files to be changed or is this a development exercise ?
- I don't get why we need [250 more lines of code](https://github.com/janosvitok/qlcplus/commit/e027eca5dcc385c387588957841fcdbf3b44a267) for something that already works
- by "installation" I mean [this](https://github.com/janosvitok/qlcplus/blob/29935deba0f8d1f36b8185aa957c4565a9cb8981/qlc.pro#L21). A shell script doesn't prevent that to happen
- "I need to deploy the C++ to users so I need to test it.". We both know that nobody will test it, so what's the deal ? Windows and macOS user don't even know what a command line tool is.

janosvitok:
> I don't get why we need 250 more lines of code for something that already works

The problem is that it does not work. QLC+ guesses what channels are most probably related. So far it's only used for PAN/TILT, so in most cases the there are only 2 PAN channels and 2 TILT channels, but there are fixtures with 4 pan channels. And things get more complicated if we want to process other 16-bit channels.

Those 250 lines of code is an attempt to improve guessing. I could do it in ruby or manually, but then user-defined fixtures that are not updated will not work. I think it's good to provide "heuristics" to pre-fill values rather then leave them empty. Nevertheless, I don't object to stripping them off.

> by "installation" I mean this. A shell script doesn't prevent that to happen

OK, I need to include the dir so that the tool compiles, but I will remove INSTALLS += target from fixturetool.pro.

> "I need to deploy the C++ to users so I need to test it.". We both know that nobody will test it, so what's the deal ? Windows and macOS user don't even know what a command line tool is.

I'm not referencing the tool itself, I'm referencing those 250 lines of code that I need to test/tune.

> Again, is there a particular issue that requires 557 files to be changed or is this a development exercise

Idea here is that many of the fixtures have attributes still in "old"/"random" order. When you open in in fixture editor and save, the file will have many changes in attribute order.
So when I will fill the MSB/LSB relations there will be two type of changes: the interesting ones related to the MSB/LSB relations and noise caused by fixing attribute order.
When I first fix attribute order (which I don't need to check, since it's tested code long time in production), then the MSB/LSB diff (that has to be manually checked, since it's mostly guess/best effort) will be much smaller.

Maybe when I add the MSB/LSB commit things will be clearer.

mcallegari:
> The problem is that it does not work. QLC+ guesses what channels are most probably related.

The current code does not "guess" anything. In your statement, you're ignoring the fact that Fixture::channelNumber works by head number. So it's not guessing. It gathers channel numbers exactly as specified in the fixture definition. If the definition is wrong, then it's a creator's fault. QLC+ cannot prevent user errors.
If a fixture has 4 pan channels, most likely it means it has 2 heads, otherwise it wouldn't make any sense. Unless you prove I'm wrong with a real example.

PR #1001, opened 2017-09-16
Script: fix load/save cycle

janosvitok:
Excess empty line was created after each cycle.
Reported by benste at http://www.qlcplus.org/forum/viewtopic.php?f=21&t=11467


mcallegari:
Not ok.
Try this:
- create a script
- add a wait command
- hit backspace
- save

The command is lost

mcallegari:
The fix is this:
```
    while (result.count() && result.last().isEmpty())
        result.takeLast();
```

PR #1003, opened 2017-10-25
Fixture: Added 9th channel (rotation of lasers) to the Eurolite-LED-FE-1500

dandanx:
Added missing channel to the Eurolite LED FE 1500.


mcallegari:
Thanks

PR #1005, opened 2017-11-07
focus video player

veecue:
As of now, the video player shows up behind other Applications (like Powerpoint) if QLC+ is not focused. I suggesst this solution, however I wasn't able to compile and test it under windows.
Cheers

### DIFF
--- a/qmlui/videoprovider.cpp
+++ b/qmlui/videoprovider.cpp
@@ -192,6 +192,8 @@ void VideoContent::playVideo()
                                   Q_ARG(QVariant, QVariant::fromValue(m_video)));
     }
 
+    m_viewContext->setFlags(m_viewContext->flags() | Qt::WindowStaysOnTopHint);
+
     if (m_video->fullscreen())
     {
         m_provider->setFullscreenContext(m_viewContext);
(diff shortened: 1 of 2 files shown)


mcallegari:
Does launching QLC+ after Powerpoint help at all ?

veecue:
No it doesnt. Focusing QLC+ before pressing the MIDI button helps, but is very impractical. My proposed change doesn't help at all, but m_videoWidget->setWindowFlag(Qt::WindowStaysOnTopHint); helps. Do you think that this is a change that could be simply made or should there be an extra checkbox for `Always on Top`?

mcallegari:
To be honest I'd like to avoid adding more options.
In general, if a user requests a video to play, it's kind of implicit that they want to see it on top of everything else.
Which platform are you using by the way ? Windows ?

veecue:
I am using Linux, but atm I have to work with windows. I haven't experienced this Problem in Linux (Gnome/Wayland), but on windows. As I understand you, I will now change this PR to the solution I found to be working (adding the always on top flag to the Video Player window).

veecue:
I have now tested this under windows and it works. However, the first time a video is played, for some reason the window does not show.

mcallegari:
Alright, let's merge this. I'll test it later when I spend (waste :laughing:) some time on Windows.
Thanks :+1:

PR #1006, opened 2017-11-16
addfixture.ui tweaks

wardrums:
fixed the height of the fixture number and offset fields so the window scales a bit nicer.


wardrums:
Please ignore this, my first try doing a pull request

PR #1010, opened 2017-11-20
litte fix for the launcher

chrisderock:
When building on OSX with the new qmlui, this fix enables the launcher to run the binary "qmlplus-qml" instead of "qlcplus"


mcallegari:
Rejected. QLC+ 5 will have no launcher.
The Fixture Editor will be a separate bundle.
The script to edit is create-dmg.sh
If you have some patience I will make the change, but the macOS bundle is the last of my concerns right now

chrisderock:
ok, no problem

mcallegari:
@chrisderock 97d88b1354a10a811a9208fdb0bfb85b3ef66ce8
This should do:
`create-dmg.sh CONFIG+=qmlui`

PR #1013, opened 2017-11-26
Fix some warnings in the german translations

kripton:
Modified whitespaces to align with source strings. No news strings to be translated.


mcallegari:
Great, thanks :+1:

PR #1016, opened 2017-12-02
Remove color from Off value in Color Macro channel.

janosvitok:
Otherwise the black color will override R/G/B values in 2d monitor.

I suspect other fixtures may have similar problem.
Maybe new test for fixture validator: Color channel must have value without color attached.

### DIFF
--- a/resources/fixtures/American-DJ-12P-Hex.qxf
+++ b/resources/fixtures/American-DJ-12P-Hex.qxf
@@ -55,7 +55,7 @@
  </Channel>
  <Channel Name="Color Macro">
   <Group Byte="0">Colour</Group>
-  <Capability Min="0" Max="3" Color="#000000">Off</Capability>
+  <Capability Min="0" Max="3">Off</Capability>
   <Capability Min="4" Max="7" Color="#ff0000">Red</Capability>
   <Capability Min="8" Max="11" Color="#00ff00">Green</Capability>
   <Capability Min="12" Max="15" Color="#0000ff">Blue</Capability>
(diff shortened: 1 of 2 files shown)


mcallegari:
I see the point, but how does this work in the real world ?
If a fixture has both color macros and RGB channels, which color will win ?
I think in QLC+ 5 I checked if color macro color != black then it will win over RGB channels.
So basically it is possible to fix this in the C++ code rather than reviewing all the fixture definitions.

mcallegari:
Fixed 91f1cf3eae7d9042330d3d7e8a2bfad4cdc05eef
Now, since you implemented the strobing preview, can you explain why this is strobing ??

[monitor_black_wheel.zip](https://github.com/mcallegari/qlcplus/files/1524708/monitor_black_wheel.zip)

janosvitok:
It depends on the fixture itself, or better said, on light source.

For RGB LED fixtures, color wheels are virtual, and they override the RGB value. For example GLP Volkslicht channel 5 - color macros has one value that means No Macro and you need to set this value in order to be able to use RGB channels (6,7,8). Any other value forces its color.

Fixtures with subtractive color mixing (e.g. halogen bulb + color wheels, the newer have white LED source + color wheels) the color subtracts from the white. That means black color should stop the light completely. In this case to pass the light unchanged "White" or "Open" is more appropriate.

I've implemented it in QLC+4 as follows:

https://github.com/mcallegari/qlcplus/blob/master/ui/src/monitor/monitorfixtureitem.cpp#L133
https://github.com/mcallegari/qlcplus/blob/master/ui/src/monitor/monitorfixtureitem.cpp#L334
When there was no color= tag in the capability, the light is unchanged.

When you ignore black color, I saw one fixture that had #000000 as "Blackout", that will not work with your version.

janosvitok:
monitor_black_wheel.zip

I can't reproduce. Can you provide exact steps to reproduce? :^)

janosvitok:
Let's close this. Workaround is committed, and most problably this whole thing will go away with the new fixture format.

PR #1023, opened 2018-01-09
E1.31: fix crash on wrong packet format

plugz:
Hello!

When receiving bad/strange/incompatible packets on E1.31 port, QLC crashes because it tries to read more bytes in the packet than the packet contains.

This PR fixes the crash by rewriting the length contained in the packet data.

Another solution would be to make this check in checkPacket(), and discard the packet on wrong length.

cheers

### DIFF
--- a/plugins/E1.31/e131packetizer.cpp
+++ b/plugins/E1.31/e131packetizer.cpp
@@ -211,7 +211,6 @@ bool E131Packetizer::fillDMXdata(QByteArray& data, QByteArray &dmx, quint32 &uni
 {
     if (data.isNull())
         return false;
-    dmx.clear();
 
     universe = (data[113] << 8) + data[114];
 
@@ -220,7 +219,8 @@ bool E131Packetizer::fillDMXdata(QByteArray& data, QByteArray &dmx, quint32 &uni
     int length = (msb << 8) | lsb;
 
     qDebug() << "[E1.31 fillDMXdata] length: " << length - 1;
-    for (int i = 126; i < 126 + length - 1; i++)
-        dmx.append(data.at(i));
+
+    dmx.clear();
+    dmx.append(data.mid(126, length - 1));
     return true;
 }


mcallegari:
Hi there !
I was thinking...a for cycle when we have `QByteArray &QByteArray::append(const QByteArray &ba)` ?

What about this:
`dmx.append(data.mid(126, qMin(data.size() - 125, length)));`

Simple and should cover all cases

plugz:
Yes, that's less of a mess.
As dmx buffer is cleared at the start of the function, we can even forget the append() and just do dmx = data.mid(...)

mcallegari:
> As dmx buffer is cleared at the start of the function, we can even forget the append() and just do dmx = data.mid(...)

Uhm...I suspect QByteArray = operator does an address assignment, not a deep copy. That would compromise the original data array.
I'd stick with the append call, which guarantee a data copy.

plugz:
Fair enough.

Also, from https://doc.qt.io/archives/qt-4.8/qbytearray.html#mid : "If len is -1 (the default), or pos + len >= size(), returns a byte array containing all bytes starting at position pos until the end of the byte array."

So qMin(...) is not necessary here.

mcallegari:
> Also, from https://doc.qt.io/archives/qt-4.8/qbytearray.html#mid : "If len is -1 (the default), or pos + len >= size(), returns a byte array containing all bytes starting at position pos until the end of the byte array."
So qMin(...) is not necessary here.

I know that, but if for some reason length is 50 but data has 100 bytes (after 125 header), we would provide 100 bytes.
Obviously it would be a bad formatted packet, but still I think length should prevail over data.size

plugz:
`data.mid(126, length - 1)`
If for some reason length is 50 but data has 100 bytes, then then length of 50 will be taken into account.

mcallegari:
> If for some reason length is 50 but data has 100 bytes, then then length of 50 will be taken into account.

Yeah, that would probably be enough.
Can you please adjust the PR with that change ?

PR #1025, opened 2018-02-04
Create Chauvet-Rotosphere-Q3.qxf

DetectiveEric:
Adding new fixture definition for Chauvet Rotosphere Q3


mcallegari:
Doesn't even pass the fixture validator: http://www.qlcplus.org/fixture_validator.php
Please fix

DetectiveEric:
Thanks, I wasn't aware of the fixture validator; changed and passes now.

I'm only committing the 19 channel mode as it's the most full featured versus the 3 and 9 channel modes; leaving those for another dev if they want to add.

mcallegari:
Merging, thanks.
Next time though, please submit something for the community, not just yourself.

PR #1027, opened 2018-02-21
Improve BCF+LC2412 tutorial + update make_docs.rb

janosvitok:
The ...a.svg file has text converted to curves.


mcallegari:
While this is OK on a web browser, have you checked that SVG pictures can be rendered correctly by QTextBrowser on Windows and macOS ?
I suspect QtSvg or imageformats/qsvg plugin are needed (in fact I bundle them for QLC+ 5)

janosvitok:
I've added PNG file that is included in html and deployed, and SVG source file for the case something needs to be changed and/or SVG format gets supported. I've even included two SVG versions: one original (bcf2000-lc2412.svg), and one with texts converted to curves (bcf2000-lc2412a.svg), so that custom fonts are not needed.

Maybe the svg files should go to new "source" subdir for now.

mcallegari:
the docs folder needs to be restructured anyway.
See http://www.qlcplus.org/forum/viewtopic.php?f=12&t=11914

mcallegari:
What is this crap ?

image

janosvitok:
Can you be more specific?

mcallegari:
That's how I see the SVG source version.
I deleted it, cause most likely you used a font that is not present in Linux.

janosvitok:
It's Latin Modern Sans Demi Cond, `apt install lmodern`. The idea was that it's easier to change the font to anything that to draw the whole thing again.

In the other version I converted the texts to curves, but then the texts are not editable anymore.

mcallegari:
Yeah, I got the meaning of it, but we should use generic fonts present in every OS (like sans, times, courier, etc)
In any case let's not complicate things. All this stuff gets bundled in the official sources, and I'm not sure if it makes sense to deliver it.

PR #1031, opened 2018-02-25
translate jp new in webaccess

magcho:
I translated the  webaccess

### DIFF
--- a/webaccess/src/webaccess_ja_JP.ts
+++ b/webaccess/src/webaccess_ja_JP.ts
@@ -106,7 +106,7 @@
     <message>
         <location filename="webaccessconfiguration.cpp" line="136"/>
         <source>Passthrough</source>
-        <translation>Passthrough</translation>
+        <translation>パススルー</translation>
     </message>
     <message>
         <location filename="webaccessconfiguration.cpp" line="210"/>
@@ -197,7 +197,7 @@
     <message>
         <location filename="webaccessconfiguration.cpp" line="325"/>
         <source>System</source>
-        <translation>System</translation>
+        <translation>システム</translation>
     </message>
     <message>
         <location filename="webaccessconfiguration.cpp" line="334"/>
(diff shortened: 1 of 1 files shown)


webaccess/src/webaccess_ja_JP.ts
```diff
@@ -0,0 +1,345 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="fr_FR">
+<context>
+    <name>WebAccess</name>
+    <message>
+        <location filename="webaccess.cpp" line="169"/>
+        <source>Loading project...</source>
+        <translation>プロジェクトを開いています...</translation>
```
mcallegari: This and other strings ending with . are warning for Qt Linguist.
However, I have no idea if in Japanese you use punctuation at the end of sentences.
Up to you.

webaccess/src/webaccess_ja_JP.ts
```diff
@@ -0,0 +1,345 @@
...
+    </message>
+    <message>
+        <location filename="webaccesssimpledesk.cpp" line="65"/>
+        <source>Next page</source>
+        <translation>次のページ</translation>
+    </message>
+    <message>
+        <location filename="webaccesssimpledesk.cpp" line="68"/>
+        <source>Reset universe</source>
+        <translation>Reset universe</translation>
```
mcallegari: Untranslated

webaccess/src/webaccess_ja_JP.ts
```diff
@@ -0,0 +1,345 @@
...
+    </message>
+    <message>
+        <location filename="webaccesssimpledesk.cpp" line="70"/>
+        <source>Universe</source>
+        <translation>Univers</translation>
```
mcallegari: Untranslated

webaccess/src/webaccess_ja_JP.ts
magcho: No problem.
Three dot mean please wait in japanese.
It is not correct grammar. But we are useing it in everyday.

mcallegari:
Thanks a lot ! :+1: 
Would be great if you can have a look also at qlcplus_ja_JP.ts in ui/src

PR #1045, opened 2018-04-03
Fix .sh codacy issues

janosvitok:
- replace backticks with $() (see https://mywiki.wooledge.org/BashFAQ/082)
- remove SOURCES from create_rpm.sh (unused since 2.6.2010)


mcallegari:
Thanks :+1:

PR #1053, opened 2018-04-12
Fixes and enhancements in WebAccess

SylvainLaugie:
This PR is about fixing and enhancing some WebAccess items:

- Buttons: JS events `onmousedown` and `onmouseup` replace `onclick` to be able to use Flash mode
- AudioTriggers: new signal `captureEnabled(bool)` in VCAudioTriggers allows WebAccess AudioTriggers state to be updated when changed in QLC+ directly
- Frames: possibility to use circular scrolling of pages
- Cuelist: fixes which prevented Cuelist to work correctly

More generally, I refactored the JS code to be more simple:
- move style update code from websocket.js to virtualconsole.js
- style updates are done only when receiving messages from the WS, not when clicking on the elements of the Web UI

This prevents style updates (ie. a button going green) when the element isn't actually activated in QLC+ (ie. if QLC+ is not in operate mode). Web UI only sends UI events to QLC+, then QLC+ tells back the Web UI what to update.  
This also simplifies websocket.js which doesn't have style update code anymore, only calls to functions in virtualconsole.js  
Finally, this removes some duplication of code between websocket.js and virtualconsole.js

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -16,6 +16,9 @@ qlcplus (4.11.2) stable; urgency=low
   * Virtual Console/Cue List: allow to select a step with next/previous buttons during pause
   * Virtual Console/Cue List: go to the right chaser step after a pause (thanks to Krzysztof Walo)
   * Virtual Console/Frame: fix regression preventing to send the disable feedback
+  * Web Access/VirtualConsole: possibility to use Buttons in Flash mode (Sylvain Laugié)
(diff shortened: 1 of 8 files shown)


mcallegari:
I'm gonna merge this cause I need to release 4.11.2.
I'll make a couple of changes that I found while testing this.
Thanks a lot for this PR. Well done.

Issue #1070, opened 2018-06-17
Infinite Pan/Tilt with the new fixture definitions


fxedel:
I'm wondering how to properly describe continuous (infinite, endless) pan/tilt movement with the new QLC+ fixtures.

There are some channel presets for pan/tilt speed, but even though they are used in [American DJ XS 600](https://github.com/mcallegari/qlcplus/blob/0c1e1fe48a618f057d43b521e7309ef9f49079ec/resources/fixtures/American_DJ/American-DJ-XS-600.qxf) to represent continuous pan/tilt speed, I recognize them more like a general motor (reaction) speed used in combination with PanPosition or TiltPosition:
```
SpeedPanSlowFast
SpeedPanFastSlow
SpeedTiltSlowFast
SpeedTiltFastSlow
SpeedPanTiltSlowFast
SpeedPanTiltFastSlow
```

As of capability presets, there's only the (never-used) `TiltContinuous` preset. Why not `PanContinous`? And what about directions and speed? I'd suggest these capability presets instead:
```
PanContinuous
PanContinuousSlowToFast
PanContinuousFastToSlow
PanContinuousStop
PanContinuousCounterClockwise
PanContinuousCounterClockwiseSlowToFast
PanContinuousCounterClockwiseFastToSlow
TiltContinuous
TiltContinuousSlowToFast
TiltContinuousFastToSlow
TiltContinuousStop
TiltContinuousCounterClockwise
TiltContinuousCounterClockwiseSlowToFast
TiltContinuousCounterClockwiseFastToSlow
```

Sure, these are many presets. But as both direction and speed is included, QLC+ could visualize continuous pan / tilt properly (or at least there's the future opportunity to do that). See the [Elation ACL 360 Roller](https://open-fixture-library.org/elation/acl-360-roller) which would immediately benefit from these new presets!

I don't think any new channel presets are needed, as continuous rotation typically requires more than one capability.

One last thing about physical data: We're currently about to allow "infinite" as possible value for `panMax`/`tiltMax` in OpenLightingProject/open-fixture-library#515. Maybe that'll also be something for QLC+?

cc @FloEdelmann

mcallegari:
The idea of capability presets is to combine them with the channel type.
If the channel is "Pan" type, saying "ContinuousSlowToFast" is enough to tell it refers to the pan motor.

The problem can eventually comes when the channel is called "Pan & Tilt speed" and there are multiple capabilities. I have no idea if there are fixtures with such issue though.

[...]

PR #1073, opened 2018-07-05
WIP: Infinite Pan/Tilt presets

fxedel:
This addresses #1070 

There's still a bug: Most fixtures have this channel order:
1. Pan
2. Tilt
3. Pan Continuous Rotation
4. Tilt Continuous Rotation
5. ...

The EFX function sets channel 3 and 4 then (which is incorrect, as it should access the angle by setting 1 and 2). I don't know how to fix this – maybe by changing something in [qlcfixturehead.cpp#L171](https://github.com/fxedel/qlcplus/blob/d089bbe38e429d9a7ba5ba9cf11623b8817d1480/engine/src/qlcfixturehead.cpp#L171) or [efxfixture.cpp#L98](https://github.com/fxedel/qlcplus/blob/d089bbe38e429d9a7ba5ba9cf11623b8817d1480/engine/src/efxfixture.cpp#L98)?


mcallegari:
Ok, I'm gonna merge this one and fix caching precedence.

mcallegari:
@fxedel  I have just tried ADJ Asteroid in Basic and Standard mode and EFX controls the right channels (1 and 2)
Where did you see the wrong behaviour ?

PR #1074, opened 2018-07-06
Update qlcplus_de_DE.ts

mitscherdinger:
Changed a wrong translation from "Messungen" to "Meter"


mcallegari:
See, that was easy ! :+1:

PR #1080, opened 2018-07-08
Add beat to Mixxx inputprofile

markusb:
Add the 'beat' signal to the inputprofile for Mixxx 'Mixxx-MIDI.qxi'.

The inputprofile for Mixxx features all signals that Mixxx can emit, except 'beat' which I find the most important. This patch adds this signal to the list.


mcallegari:
Thanks :+1:

PR #1082, opened 2018-07-10
qmlui: Catalan translation

sbenejam:
qmlui: Catalan translation

### DIFF
--- a/qmlui/qlcplus_ca_ES.ts
+++ b/qmlui/qlcplus_ca_ES.ts
@@ -1,0 +1,4152 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE TS>
+<TS version="2.1" language="ca_ES">
+<context>
+    <name>ActionsMenu</name>
+    <message>
+        <location filename="qml/ActionsMenu.qml" line="45"/>
+        <source>Open a project</source>
+        <translation>Obrir un projecte</translation>
+    </message>
+    <message>
+        <location filename="qml/ActionsMenu.qml" line="47"/>
+        <location filename="qml/ActionsMenu.qml" line="64"/>
+        <location filename="qml/ActionsMenu.qml" line="79"/>
+        <source>Project files</source>
+        <translation>Fitxers de Projecte</translation>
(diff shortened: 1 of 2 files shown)


mcallegari:
Awesome job Santiago ! Thanks a lot :+1:

sbenejam:
Hi Massimo I have not translated the days of the week.
The Catalan days of the week and their abbreviations are:
     dilluns: dl.
     dimarts: dt.
     dimecres: dc.
     dijous: dj.
     divendres: dv.
     dissabte: ds.
     diumenge: dg.
I did not change the days of the week I left them in English. Maybe it
would be good to reconsider how to translate these items in other languages.
What do you think about it?

Regards

El dt., 10 jul. 2018, 22:24, Massimo Callegari <notifications@github.com>
va escriure:

mcallegari:
@sbenejam The idea is to use the first letter of the days. At least in all the other languages we did that.
Would it work in Catalan as well ?

sbenejam:
I think it will not work for Catalan, info from wikipedia.

En català les abreviatures que s'utilitzen per als dies de la setmana
sempre són de dues lletres i han de portar un punt final, comencen amb la
lletra *d* que va seguida per la consonant que hi ha després de la *i*,
excepte en el cas del dimarts, el dimecres i el diumenge on s'utilitza una
*t* pel dimarts, una *c* pel dimecres i una *g* pel diumenge:

In Catalan the abbreviations that are used for the days of the week always
are of two letters and must take a final point, begin with letter d that is
followed by the consonant that there is after the i, except in the
case Tuesday,
Wednesday and Sunday where a t is used on Tuesdays, a c on Wednesdays and a
g on Sundays:

   - dilluns: dl.
   - dimarts: dt.
   - dimecres: dc.
   - dijous: dj.
   - divendres: dv.
   - dissabte: ds.
   - diumenge: dg.




2018-07-11 19:54 GMT+02:00 Massimo Callegari <notifications@github.com>:

mcallegari:
Yeah, sorry, I realized that too late and edited my original message. Please check it out.

PR #1083, opened 2018-07-11
qmlui Spanish translation

tornel:


mcallegari:
Thanks a lot Luis ! :+1: 
Days of the week are missing, but I can add them :wink:

PR #1088, opened 2018-07-22
Use new path for AppStream metadata as a variable

jeromelebleu:
It defines the AppStream metadata path as a common variable and makes use of it in each project file where it is also defined. Finally, it uses the new path for the application metainfo - see [the relevant documentation](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location) - as the legacy `/usr/share/appdata` path will be dropped in future.


mcallegari:
Wait, what ?
Seems like a very dangerous change to me, especially since I think you didn't test it on all the major Linux distro, did you ?

jeromelebleu:
This is just to follow the current AppStream's specification as described in its documentation - see [here](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#spec-component-location). It seems to be the standard path since at least [v0.9.4](https://github.com/ximion/appstream/blob/master/NEWS#L949) - which has been released on 2016-04-18 - so nothing to deal with me and my mood. Moreover, it is said:

> AppStream tools scan the /usr/share/appdata/ path for legacy compatibility as well. It should not be used anymore by new software though, even on older Linux distributions (like RHEL 7 and Ubuntu 16.04 LTS) the metainfo path is well supported. Support for the legacy path might be dropped completely in future.

I indeed not test that on all the major Linux distributions, *only* Debian. I believe that the others follow the AppStream specification. In fact, I'm working on the Debian packaging in order to include this great software in the official repository, and this one of the errors reported by lintian. I thought it would be better to submit this fix upstream.

mcallegari:
Alright, I think I misread the initial commit and thought you wanted to change the whole /usr/share location !
Looks OK to me, but I checked my /usr/share/metainfo folder in a Ubuntu 16.04 derivate (KDE Neon) and there is only KDE stuff.
Moreover, all the XML files in there start with "org.kde.*". So eventually QLC+ should use org.qlcplus.* ?
Anyway I don't understand what's the deal here. Looks like nobody adopted it (yet!?) but lintian fails !
I'm confused

jeromelebleu:
No problem, glad to know that there were a confusion! I was a bit surprised of your reaction and was wondering if I missed something...

Honestly, I don't know a lot about AppStream - I've most likely discovered it with this lintian error! In my side, I have both `<appname>.appdata.xml` and `org.*.appname.xml` - generally `org.gnome.*` - in my `/usr/share/mateinfo` directory. And effectively, the [`id`](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic) documentation tells that something `org.qlcplus.*` should be used as I understand it.

mcallegari:
@jeromelebleu can you please rebase ?
Also I think it makes more sense to rename all appdata files to org.qlcplus.*
Thanks

jeromelebleu:
@mcallegari I have rebased and renames both files and `<id>` tags to starts with `org.qlcplus.QLCPlus`. I have tried to follow specifications and AppStream metadata of other applications.

I have by the way also added `<provides>` tag when relevant - it was reported as a warning by lintian, see [appstream-metadata-missing-modalias-provide](https://lintian.debian.org/tags/appstream-metadata-missing-modalias-provide.html) for details.

Is it good for you?

mcallegari:
Hi, thanks. I'm not sure if "QLCPlus" is what we want. I checked again my metainfo folder and all the names are lowercase. Examples:
```
org.kde.dolphin.appdata.xml
org.kde.plasma.fuzzyclock.appdata.xml
```
So, I would say this is it:
```
org.qlcplus.qlcplus
org.qlcplus.qlcplus-fixtureeditor
org.qlcplus.artnet-plugin
```
I'm not sure about the "extends" tag either. Does it refer to the actual .desktop file or to another metainfo file ? In the first case, you should not rename .desktop entries.

jeromelebleu:
I've chosen camel case since it is used in official specification - see [`<id/>`](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic) - and generally in all Gnome applications. [This report](https://appstream.debian.org/sid/main/metainfo/index.html) from Debian gives an overview of the variety! Honestly, I don't care, it's as you want. It is also said that hyphens are not recommended to *ensures the greatest possible compatibility of an AppStream ID*.

However, as I understand `org.qlcplus.artnet-plugin` is not "valid" since it's part of QLC+. It should be prefixed by the "product" id, such as `org.qlcplus.QLCPlus.artnet` or `org.qlcplus.qlcplus.artnet`, or even `org.qlcplus.qlcplus.plugin.artnet` or `org.qlcplus.QLCPlus.Plugin.ArtNet`. gnome-software uses for example last one, see [here](https://appstream.debian.org/sid/main/metainfo/gnome-software.html).

Regarding `.desktop` suffix, as it's generally used for the other applications metadata, I'll keep it for qlcplus and qlcplus-fixtureeditor IDs. It will become then `<provides>org.qlcplus.QLCPlus.desktop</provides>` - or in lowercase - for each plugin metadata.

peternewman:
@jeromelebleu see the HTTPS discussion here:
https://github.com/mcallegari/qlcplus/pull/1143

FloEdelmann:
Maybe those blog posts by a KDE developer could help in this discussion:

* https://pointieststick.wordpress.com/2018/01/14/how-you-can-help-drive-flatpak-adoption/
* https://pointieststick.wordpress.com/2018/01/27/how-to-make-an-app-look-good-in-discover/

mcallegari:
Merged! Sorry for letting this pending for so long.

PR #1097, opened 2018-07-24
Improve one UI string

kripton:
Found during translation: I'm not sure what is meant there but the current string sounded a bit odd


mcallegari:
Oh damn ! I should have been "has requested to access".
This means to review all the translations as well

kripton:
Yes, "has requested" is the simpler change. But if a client is connecting and can still be granted or denied access I prefer the formulation with "is requesting access to". Your choice ;)

mcallegari:
Makes sense. Your proposal then ! Thanks for spotting

PR #1096, opened 2018-07-24
Qmlui translation de

kripton:
Completes the german translation of the QML UI
Based on the nice work of @fxedel, thanks!

PR #1098, opened 2018-07-24
Update german translations from current master

kripton:
#1096 wasn't up-to-date with latest master so here we go :)


mcallegari:
Great, thanks ! :+1:

PR #1101, opened 2018-07-25
Updated qmlui catalan translation

sbenejam:
Updated new and modified strings in catalan translation.


mcallegari:
Thanks Santiago ! :+1:

PR #1104, opened 2018-07-26
Remove spaces before punctuation

FloEdelmann:
This is basically #1102 again, but this time also for exclamation marks and the commits are split better to help ease reviewing.

* :us: :gb: English: No space before question mark (?) or exclamation mark (!) [<sup>Source</sup>](https://english.stackexchange.com/q/4645)
* :czech_republic: Czech: No space [<sup>Source: janosvitok</sup>](https://github.com/mcallegari/qlcplus/pull/1102#issuecomment-407924137)
* :de: German: No space <sup>Source: How I learned it in school ;)</sup>
* :it: Italian: No space [<sup>Source: mcallegari</sup>](https://github.com/mcallegari/qlcplus/pull/1102#issuecomment-407909930)
* :netherlands: Dutch: No space [<sup>Source: Netflix subtitle translator guidelines, section 14</sup>](https://partnerhelp.netflixstudios.com/hc/en-us/articles/215350158-Dutch-Timed-Text-Style-Guide)
* :es: Spanish: No space, but upside-down question/exclamation mark in front of the (sub)sentence [<sup>Source</sup>](https://www.thoughtco.com/upside-down-punctuation-in-spanish-3080317)
* :fr: French: Spaces mandatory; in software, use non-breaking spaces [<sup>Source: Duolingo</sup>](https://www.duolingo.com/comment/7344/Spaces-before-punctuation-marks)

I've included the updated code comments again; maybe someone else can review them quickly (shouldn't be too difficult in this case). If you prefer, we can revert the respective commits again.

PR #1105, opened 2018-07-26
Remove space before punctuation in user-facing strings

FloEdelmann:
See #1104.


mcallegari:
Alright, thanks !
Let's cross all the fingers we have and hope this doesn't break anything :laughing:

PR #1115, opened 2018-07-28
Translate qmlui japanese

magcho:


mcallegari:
Thanks a lot Dai :+1: 
I'm gonna update the ja_JP file cause it's behind by 25 new strings

PR #1118, opened 2018-07-31
Added dutch translation

ieperlingetje:
I saw that the dutch translation was removed in the newest version of QLC+, so I added a translation file.


mcallegari:
@ieperlingetje thanks a lot for this ! :+1: 
I just found some mismatching tags in the XML so I fixed them. If you used a text editor to edit it, please consider using Qt Linguist, which is a nice tool dedicated to Qt translations.

ieperlingetje:
I based my translation off of the French translation as a guidance for context when translating strings, which use 'appareil' as the equivalent for fixture.  I'll use Qt linguist in the future. I did use GitHub's built-in editor, which may explain why there was some tag mismatching.

PR #1120, opened 2018-08-05
Tidy up channel presets

FloEdelmann:
Some inconsistencies in channel preset lists I noticed during the development of Open Fixture Library's QLC+ 4.12.0 plugin (see [OpenLightingProject/open-fixture-library#547](https://github.com/OpenLightingProject/open-fixture-library/pull/547)).

Also, there is the `PositionPanCounterClockwise` channel preset, but no matching clockwise or Tilt presets. Is this a mistake?

When adding / removing / renaming a preset, one always needs to do so in various different places:

* XSD schema
* fixture tool
* qlcchannel.h
* qlcchannel.cpp 2 times
* qlccapability.h
* qlccapability.cpp
* maybe more "fixed" places
* and of course everywhere it is used, i.e. fixtures and code like strobe frequency preview

Would it be a good idea to add some kind of a checklist somewhere to reduce the risk of missing something? (We have this for OFL, too; see https://github.com/OpenLightingProject/open-fixture-library/blob/master/docs/capability-types.md#how-to-add-new-capability-types--type-specific-properties)

### DIFF
--- a/engine/src/qlccapability.cpp
+++ b/engine/src/qlccapability.cpp
@@ -92,6 +92,9 @@ void QLCCapability::setPreset(QLCCapability::Preset preset)
     m_preset = preset;
 }
 
+/* please see
+https://github.com/mcallegari/qlcplus/wiki/Fixture-definition-presets
+when changing this function */
 QLCCapability::PresetType QLCCapability::presetType() const
 {
     switch (m_preset)
@@ -119,6 +122,9 @@ QLCCapability::PresetType QLCCapability::presetType() const
     }
 }
 
+/* please see
+https://github.com/mcallegari/qlcplus/wiki/Fixture-definition-presets
+when changing this function */
 QString QLCCapability::presetUnits() const
 {
     switch (m_preset)
--- a/engine/src/qlccapability.h
+++ b/engine/src/qlccapability.h
@@ -95,7 +95,11 @@ public:
     bool operator<(const QLCCapability& capability) const;
 
     /********************************************************************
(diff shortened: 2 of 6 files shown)


mcallegari:
PositionPanCounterClockwise made sense when I first added it.
Then I worked on Monitor properties and now there is a flag to invert pan/tilt there. It was Jano's suggestion to cope with channels inverted from the product menu.
Now PositionPanCounterClockwise can be removed and eventually users will invert from channel properties. The important thing is that there is a way to tell previews how to handle pan/tilt channels.

As for checklist, I agree, but at the moment I have zero time. I am going to release 5.0.0 Alpha3 today.
The proper place could be GitHub Wiki.

FloEdelmann:
I also did the checklist (and a bit of documentation), but I don't have write access to the QLC+ GitHub Wiki, so I created a gist: https://gist.github.com/FloEdelmann/1ad10d637fc932e633f989f4c50835bf. Feel free to modify it and copy it to the Wiki, I will then delete my gist again.

I'd add a link to the Wiki page (as soon as it exists) as code comments in all the places listed above, so please don't immediately merge this PR.

mcallegari:
@FloEdelmann Thanks a lot!
I've added the page to the wiki and linked in the main page (developers corner)

FloEdelmann:
Okay, I've added the comments, so now I think this PR is finished :)

mcallegari:
Thanks @FloEdelmann ! :+1:

PR #1132, opened 2018-08-12
Update DMX-USB Plugin Catalan translation

sbenejam:
Catalan translation, updated new strings DMX-USB Plugin


mcallegari:
Thanks :+1:

PR #1137, opened 2018-08-19
plugins/ola: enable pkg-config on macx platform

delta512:
I get errors when building **qlcplus** on macOS out of the box because I've installed **ola** from source.

This change to **plugins/ola/ola.pro** enables pkg-config queries on those macs where it's available, while behaving exactly as it did before otherwise (ola installed through MacPorts).

It reuses commented instructions, choosing between the two methods based on a qmake test for pkg-config existence.

### DIFF
--- a/plugins/ola/ola.pro
+++ b/plugins/ola/ola.pro
@@ -12,13 +12,19 @@ QTPLUGIN  =
 INCLUDEPATH += ../interfaces
 
 macx: {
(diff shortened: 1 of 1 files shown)


mcallegari:
Have you followed at least this bit of instruction here:
https://github.com/mcallegari/qlcplus/wiki/OSX-build-Qt5#dependenciesports
(see OLA build in bold)

You should install OLA in /opt/local

PR #1140, opened 2018-09-06
Additional Fixture: Varytec Street Bar

danthedeckie:
https://www.thomann.de/gb/varytec_led_street_bar_16x3w_rgb_ip65.htm


mcallegari:
Thanks but this fails validation so cannot be accepted.
You need to use the Fixture Editor of a [QLC+ 4.12.0 test build](https://www.qlcplus.org/forum/viewtopic.php?f=12&t=3135).
Also, please use [the new syntax](https://www.qlcplus.org/forum/viewtopic.php?f=17&t=12495).
Otherwise, post this in the forum and I'll converti it when I have some time.

PR #1143, opened 2018-09-18
Fix some URLs, make others HTTPS

peternewman:


FloEdelmann:
Links to qlcplus.org in the README can also be changed to https.

peternewman:
I'm assuming the Japanese one should have had all the same links as English so I added the missing ones too.

I've done the README, but as you say, plenty more left to httpsify:
https://github.com/mcallegari/qlcplus/search?q=www.qlcplus.org&type=Code

mcallegari:
Thanks! :+1:

PR #1148, opened 2018-10-07
minor translation fix for german

danfai:
The change contains two translation strings that probably were mixed up by creating.


mcallegari:
@fxedel @FloEdelmann @kripton please review ?

PR #1151, opened 2018-10-24
Added flash button support at webaccess for Mobile Devices

marianeitor:
I noticed some time ago that flash buttons where not working at all in the web access mode if you are using a mobile device, and I tought it will be nice to have them working because that would make the web access even more powerful when using mobile devices.

I tested it and it works great.

Also 
```
make check
~/qlcplus_fork/qlcplus
Unit tests passed.
```

### DIFF
--- a/webaccess/res/virtualconsole.js
+++ b/webaccess/res/virtualconsole.js
@@ -19,21 +19,7 @@
 
 /* VCButton */
 
-window.addEventListener("load",() => {
-    var buttons = document.getElementsByClassName("vcbutton");
 
-    for (var btn of buttons) {
-        btn.addEventListener("touchstart", (event) => {
-                 event.preventDefault();
-                buttonPress(this.id)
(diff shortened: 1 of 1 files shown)


enbyted:
Looks good, tested on touch and non-touch input on Firefox, Chrome and Edge on Windows. Only touch tested in android Browser, Chrome and Chrome beta on Android.

mcallegari:
Thanks for this. I'll test as well when I'm on a mobile device

enbyted:
Oh upon further testing with network sniffer it seems this only appears to be working.
`buttonPress(this.id);` should be `buttonPress(event.target.id);`
and similarly for `buttonRealese`

Without it the id passed is `undefined`. Weird that I haven't spotted this before and that it's been working.
Maybe I didn't save a file or something.

mcallegari:
Fixed, thanks

PR #1152, opened 2018-11-01
Redraw some Chauvet PNG gobos as SVG

FloEdelmann:


mcallegari:
Awesome! Thanks a lot @FloEdelmann !

PR #1154, opened 2018-11-10
added ADJ-5P-Hex

exmatrikulator:
Manual: https://cdb.s3-us-west-1.amazonaws.com/ItemRelatedFiles/8658/5p_hex_5p_hex_pearl.pdf


mcallegari:
Rejected. It's in the wrong location !
Should be in `resources/fixtures/American_DJ`.
The manufacturer is "American DJ", not "ADJ".
Also this doesn't use the new format. See this: https://www.qlcplus.org/forum/viewtopic.php?f=17&t=12495

PR #1155, opened 2018-11-12
Add bloom effect

Erkaman:
Adds a bloom effect that affects RGBPanels: 

capture


qmlui/qml/fixturesfunctions/3DView/3DView.qml
```diff
@@ -198,7 +267,7 @@ Rectangle
 
                         sgNode = component.createObject(frameGraph.myCameraSelector,
                         {
-                            "fixtureItem": headEntity,
```
mcallegari: This should be headEntity

Erkaman: Fixed

PR #1163, opened 2018-11-26
Add French translation for new updates

nils-van-zuijlen:
Add French translation for the new 4.12.0 and 5alpha4 updates. (A bit late, but finally, here they are)

@jeromelebleu


mcallegari:
Thanks a lot! :+1:

PR #1164, opened 2018-12-03
Migrate 9 more gobos to SVG

FloEdelmann:


mcallegari:
Awesome stuff! Thanks a lot :+1: 
Out of curiosity, did you find them in some manual or traced it from bitmap ?
I searched for GLP gobos a lot and couldn't find anything usable.

FloEdelmann:
I actually traced them by hand. A lot of work, but those SVG gobos look so much nicer! (Besides being useful for 3D preview, of course :wink:)

Where the current resolution was too low, I searched the PDF manuals and the download sections of the fixtures using the gobo for higher resolution or even vector images. Unfortunately, none of them actually had vector images :/

mcallegari:
Well, in case you don't know it already, Inkscape is an excellent tool for this work.
I used it for all the gobos and icons. It can extract vector resources from PDF and it also includes a bitmap tracer (based on potrace)
Then at the end I apply [SVGO](https://github.com/svg/svgo), a NodeJS based tool for optimizing SVGs.
It strips them to hell, but double check it doesn't mess them up.

FloEdelmann:
I do know, and I always like to use it, too :sweat_smile: 

I just saved the images as *Optimized SVG* directly from Inkscape instead of running SVGO afterwards. Maybe Inkscape uses SVGO internally? The resulting SVG code looks quite similar, at least.

Issue #1169, opened 2019-01-21
Martin MAC Encore Performance Gobos


FloEdelmann:
The [Martin MAC Encore Performance](https://open-fixture-library.org/martin/mac-encore-performance) is not yet included in QLC+ (but it could easily with the OFL export).

Its manual includes 4 nice vector gobos (and one glass gobo image) that seem to be not included in QLC+ yet:

![screenshot_2019-01-21_15-54-15](https://user-images.githubusercontent.com/202916/51481790-f8ea9800-1d94-11e9-9f3e-2cecf34ce93e.png)

I cleaned up the vector gobos (everything is one path now) and cropped the glass gobo image, but I don't know where to put them, because there is no Martin gobo folder. I'll just append them here; you'll surely know where to put them best :blush: 

[mac-encore-performance-gobos.zip](https://github.com/mcallegari/qlcplus/files/2779436/mac-encore-performance-gobos.zip)

mcallegari:
Sorry for the delay. These go to "others".
If you can, please send a PR with the definition and the gobos. Thanks!

P.S. broken marble is 387kb ! :sob:

PR #1172, opened 2019-02-16
Add fixture Martin MAC Encore Performance

FloEdelmann:
See #1169.

I've fixed the large gobo file size by simplifying the path in Inkscape, which doesn't decrease the visual quality too much.


mcallegari:
hmmm doesn't pass schema validation

FloEdelmann:
I know, I'm already working on a fix.

Global `<Physical>` is always required, even if all modes have a `<Physical>` tag themselves. Since the schema can't enforce that either a global physical or all mode physicals are set, I have to write that in the `fixture-tool` validation.

mcallegari:
Why only physical per mode (2 identical if I see it right) and no global at all? In the end I took the global physical idea from OFL ! :sweat_smile: 
This definition should have 1 single global or 1 global + 1 in mode. Not 2 in modes.
Am I missing something?

FloEdelmann:
In this fixture we technically misuse the modes: *CLD* and *WRM* are two different variants of the fixture, but they only differ in the used bulb and color temperature correction filter (channel 7), so merging them into one fixture seems okay. However, setting one bulb in the global physical definition and overwriting it in only one mode seems wrong to me.

We could of course instead change the QLC+ export plugin so that it would use the first mode's physical for the whole fixture and remove the first mode's local one (in cases like this).

mcallegari:
Thanks, sorry for the delay

PR #1173, opened 2019-02-16
Schema updates

FloEdelmann:
Fixes the schema for #1172 (and others like it).

I also added `3-pin IP65` and `5-pin IP65` waterproof DMX connectors, like we have in OFL. Feel free to revert that commit if you don't want them in QLC+.

### DIFF
--- a/resources/schemas/fixture.xsd
+++ b/resources/schemas/fixture.xsd
@@ -41,7 +41,7 @@ elementFormDefault="qualified"
             <xs:element name="Type" type="typeType"/>
             <xs:element name="Channel" type="channelType" maxOccurs="unbounded"/>
             <xs:element name="Mode" type="modeType" maxOccurs="unbounded"/>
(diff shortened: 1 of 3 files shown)


resources/schemas/fixture.xsd
```diff
@@ -359,6 +359,8 @@ when changing this list -->
         <xs:enumeration value="3-pin"/>
         <xs:enumeration value="5-pin"/>
         <xs:enumeration value="3-pin and 5-pin"/>
+        <xs:enumeration value="3-pin IP65"/>
+        <xs:enumeration value="5-pin IP65"/>
```
mcallegari: Technically, if you add entries here, you should add the same to the fixture editor, somewhere around here
https://github.com/mcallegari/qlcplus/blob/master/fixtureeditor/editphysical.ui#L213

not a big deal though

PR #1175, opened 2019-03-25
Adaptations for Qt 5.13

chrisderock:
This patch replaces "qSort" with "std::sort" and "QFontMetrics::width" with "QFontMetrics::horizontalAdvance" if compiled with QT 5.13.

### DIFF
--- a/engine/src/rgbtext.cpp
+++ b/engine/src/rgbtext.cpp
@@ -165,7 +165,7 @@ int RGBText::scrollingTextStepCount() const
 #if (QT_VERSION < QT_VERSION_CHECK(5, 13, 0))
         return fm.width(m_text);
 #else
-	return fm.horizontalAdvance(m_text);
+        return fm.horizontalAdvance(m_text);
 #endif
     }
 }
--- a/engine/src/rgbtext.cpp
+++ b/engine/src/rgbtext.cpp
@@ -161,8 +161,13 @@ int RGBText::scrollingTextStepCount() const
     QFontMetrics fm(m_font);
     if (animationStyle() == Vertical)
         return m_text.length() * fm.ascent();
-    else
+    else{
+#if (QT_VERSION < QT_VERSION_CHECK(5, 13, 0))
         return fm.width(m_text);
+#else
+	return fm.horizontalAdvance(m_text);
+#endif
+    }
 }
 
 RGBMap RGBText::renderScrollingText(const QSize& size, uint rgb, int step) const
--- a/engine/src/sequence.cpp
+++ b/engine/src/sequence.cpp
@@ -172,7 +172,7 @@ bool Sequence::loadXML(QXmlStreamReader &root)
 #if (QT_VERSION < QT_VERSION_CHECK(5, 13, 0))
         qSort(sceneValues.begin(), sceneValues.end());
 #else
-	(sceneValues.begin(), sceneValues.end());
+	std::sort(sceneValues.begin(), sceneValues.end());
 #endif
         m_needFixup = false;
     }
--- a/engine/src/sequence.cpp
+++ b/engine/src/sequence.cpp
@@ -169,7 +169,7 @@ bool Sequence::loadXML(QXmlStreamReader &root)
     if (scene != NULL)
     {
         sceneValues = scene->values();
-	std::sort(sceneValues.begin(), sceneValues.end());
+        std::sort(sceneValues.begin(), sceneValues.end());
         m_needFixup = false;
     }
 
@@ -251,7 +251,7 @@ void Sequence::postLoad()
             return;
         }
 
-	std::sort(sceneValues.begin(), sceneValues.end());
+        std::sort(sceneValues.begin(), sceneValues.end());
     }
 
     int stepIndex = 0;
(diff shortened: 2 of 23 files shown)


mcallegari:
Thanks, but I don't like how you did the sort change.
I'd rather put a define in a common .h file (e.g. engine/src/utils.h), like:
```
#if (QT_VERSION >= QT_VERSION_CHECK(5,13,0))
#define qSort(b, e, c)  std::sort(b, e, c)
#endif
```

engine/src/sequence.cpp
```diff
@@ -169,7 +169,11 @@ bool Sequence::loadXML(QXmlStreamReader &root)
     if (scene != NULL)
     {
         sceneValues = scene->values();
+#if (QT_VERSION < QT_VERSION_CHECK(5, 13, 0))
         qSort(sceneValues.begin(), sceneValues.end());
+#else
+	(sceneValues.begin(), sceneValues.end());
```
mcallegari: this won't even build

chrisderock:
i think that would be useful for this special situation. but how to definde this for qSort(container) vs. std::sort(container.begin(), container.end())?

engine/src/sequence.cpp
chrisderock: sorry, my fault. i forgot the function call...

mcallegari:
> i think that would be useful for this special situation. but how to definde this for qSort(container) vs. std::sort(container.begin(), container.end())?

You just change `container` to `container.begin(), container.end()`

chrisderock:
i can change the modifications as you sugested.. but in what for i file should i use? this covers the whole source tree?

mcallegari:
...or we can just use std:sort eveywhere.
I think qSort was just a wrapper anyway, and std:sort could have been used since the beginning, even during Qt4 times.
My only concern is toolchain compatibility, but apparently std:sort if C++98, so should be fine

chrisderock:
i removed the version check but in rgbtext.cpp and rgbtext_test.cpp it is needed (imho)

engine/src/rgbtext.cpp
```diff
@@ -161,8 +161,13 @@ int RGBText::scrollingTextStepCount() const
     QFontMetrics fm(m_font);
     if (animationStyle() == Vertical)
         return m_text.length() * fm.ascent();
-    else
+    else{
+#if (QT_VERSION < QT_VERSION_CHECK(5, 13, 0))
         return fm.width(m_text);
+#else
+	return fm.horizontalAdvance(m_text);
```
mcallegari: Indentation please

qmlui/importmanager.cpp
```diff
@@ -439,7 +439,7 @@ void ImportManager::importFunctionID(quint32 funcID)
 
             if (removeList.isEmpty() == false)
             {
-                qSort(removeList.begin(), removeList.end());
+		std::sort(removeList.begin(), removeList.end());
```
mcallegari: Indentation

qmlui/virtualconsole/vcslider.cpp
```diff
@@ -606,8 +606,8 @@ void VCSlider::slotTreeDataChanged(TreeModelItem *item, int role, const QVariant
     else
     {
         addLevelChannel(fixtureID, chIndex);
-        qSort(m_levelChannels.begin(), m_levelChannels.end());
-    }
+	std::sort(m_levelChannels.begin(), m_levelChannels.end());
```
mcallegari: Indentation

qmlui/virtualconsole/vcslider.cpp
```diff
@@ -1381,9 +1381,9 @@ bool VCSlider::loadXMLLevel(QXmlStreamReader &level_root)
         } while (level_root.readNextStartElement());
     }
 
-    if (m_levelChannels.count())
-        qSort(m_levelChannels.begin(), m_levelChannels.end());
-
+    if (m_levelChannels.count()){
+	std::sort(m_levelChannels.begin(), m_levelChannels.end());
```
mcallegari: Indentation. No parenthesis

ui/src/rgbmatrixeditor.cpp
```diff
@@ -1067,7 +1067,7 @@ void RGBMatrixEditor::slotSaveToSequenceClicked()
             }
             // !! Important !! matrix's heads can be displaced randomly but in a sequence
             // we absolutely need ordered values. So do it now !
-            qSort(step.values.begin(), step.values.end());
+	    std::sort(step.values.begin(), step.values.end());
```
mcallegari: Indentation

ui/src/virtualconsole/vcslider.cpp
```diff
@@ -549,7 +549,7 @@ void VCSlider::addLevelChannel(quint32 fixture, quint32 channel)
     if (m_levelChannels.contains(lch) == false)
     {
         m_levelChannels.append(lch);
-        qSort(m_levelChannels.begin(), m_levelChannels.end());
+	std::sort(m_levelChannels.begin(), m_levelChannels.end());
```
mcallegari: Indentation

engine/src/rgbtext.cpp
chrisderock: Can you add a comment about the indentation style maybe to the README.md?

mcallegari:
Looks ok now, thanks

PR #1176, opened 2019-03-28
Rgbscripts strobe

chrisderock:
a small improvement for strobe.js to set a frequency for the strobe. with this you cannot only blink every second step. you can define "blink on every Xth step". Also changed the big if-else-block to a simple one-liner

### DIFF
--- a/resources/rgbscripts/strobe.js
+++ b/resources/rgbscripts/strobe.js
@@ -28,13 +28,13 @@ var testAlgo;
       algo.author = "Rob Nieuwenhuizen";
       algo.properties = [];
       algo.acceptColors = 1;
-      algo.properties.push('name:freq|type:range|display:Frequency|values:2,10|write:setFreq|read:getFreq');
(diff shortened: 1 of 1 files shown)


mcallegari:
Please have a look at the other scripts with parameters.
Frequency should be a member of algo.
Example: https://github.com/mcallegari/qlcplus/blob/master/resources/rgbscripts/circles.js#L32

chrisderock:
The first try was with algo.frequency but codacity mentioned errors. can you give me a hint how to avoid errors like "foo[key] = falue; is not allowed"?

chrisderock:
seems that i found a way to avoid map[key] ;)

mcallegari:
I wouldn't bother Codacy that much in this case.
We're not dealing with pages of a public website and Codacy JS linter is made for that.

chrisderock:
i know. but if i can avoid such an error i do that ;)

PR #1178, opened 2019-04-08
rgbscript: clone properties

plugz:
Hi,

A small patch to fix the fact that when cloning an rgbscript, the properties are not cloned and are set to default in the new rgbscript.


mcallegari:
Thanks! :+1:

PR #1179, opened 2019-04-14
Update control

mitscherdinger:
Added libqt5multimedia5-plugins, so the video-function has a codec to work with…


mcallegari:
Thanks :+1:

PR #1180, opened 2019-05-18
Add generic chinese moving head fixture.

daedeloth:
Fixture definition and gobos for these lights:
https://nl.aliexpress.com/item/8-pieces-lot-60w-led-moving-head-spot-mini-LED-stage-light-mini-Led-Follow/32725955226.html

(I'm sure they are generic so probably also sold under other names)


mcallegari:
Thanks but no thanks.
Chinese products are not accepted upstream.
You can submit it here: https://github.com/mcallegari/qlcplus-extras

daedeloth:
But the extra is not included in the installation?

Op zo 19 mei 2019 12:16 schreef Massimo Callegari <notifications@github.com

mcallegari:
> But the extra is not included in the installation?
No, it's not.

daedeloth:
> > But the extra is not included in the installation?
> > No, it's not.

Then I see no reason to include it there.
I'll just have to make sure to keep my own fixtures in sync.

I don't really understand your reasoning but I do understand that discussing it is pointless :-)

mcallegari:
Of course I've got a reason but I was too lazy to explain again. ;-)
There are a gazillion of Chinese products out there. Most of them are clones of the clone of a clone.
Including (all) of them would mean to add "crap" to the fixture library, since most likely they will be useful to very few users.
The QLC+ library is aiming to be useful to the most.
All the rest is considered custom fixture.
Hope this explains.

mcallegari:
By the way your commit didn't comply to standards.
Product directory included spaces and gobos are not allowed if licensed.
Plus, gobos should be svg now.
Next time please use the dedicated area in the forum so we can discuss there

PR #1181, opened 2019-05-22
Implement a 'setFunctionStatus' websocket api method

sgdc3:
I needed a way to start a function (a show in my case) without having to create a button widget in the virtual console.
I think that such method might be useful to others so I opened this PR.
Let me know if something doesn't look right.


mcallegari:
Looks OK, thanks.
[Test web page](https://github.com/mcallegari/qlcplus/blob/master/webaccess/res/Test_Web_API.html) should be updated too so people will know this method exists.

PR #1182, opened 2019-06-02
Introducing Open RX : a way to read DMX signals via simple RSR232 FTDI chips

manuco:
This pull request adds another plugin designed for Enttec like USB DMX adapter based on FTDI RS232R chips.

Called "Open RX", it is based on the Enttec Open TX plugin code, but reads data and emits them in the software instead of writing them.

FTDI RS232 chips can send and receive signals. As a DMX widget, they are usually built with a female DMX socket, since it's the main usage.
    
Using a male/male adaptator and plugging in the widget in a DMX source, it can receive the signal (instead of transmitting it) and forward it to the host computer.
    
It has been tested with another DMX Open TX FTDI device ran by QLC. It works well and is very stable (if the computer hardware is too).
    
It has also been tested with a Stairville DDC-6 LCD very cheap ligthing controller. This controller vaguely speaks DMX, with 250 Hz frame rate frequency (!) and frames of 29 channels. This plugin handles this case, trying to manage bogus frames.
    
As a result, a user can drive 6 sliders (or other virtual widgets) with a cheap DMX Controller using only another cheap DMX USB device.

I didn't have enough devices to test more configuration, but given the fact that the plugin is an optionnal feature, I hope this requests will be accepted.

Please feel free to comment and suggest.

### DIFF
--- a/plugins/dmxusb/src/dmxusb.cpp
+++ b/plugins/dmxusb/src/dmxusb.cpp
@@ -269,6 +269,8 @@ QString DMXUSB::inputInfo(quint32 input)
         str += QString("<H3>%1</H3>").arg(inputs()[input]);
         str += QString("<P>");
         str += tr("Device is operating correctly.");
+        str += QString("<BR>");
+        str += tr("Driver in use: %1").arg(m_inputs[input]->interfaceTypeString());
         str += QString("</P>");
         QString add = m_inputs[input]->additionalInfo();
         if (add.isEmpty() == false)
--- a/plugins/dmxusb/src/dmxusb.cpp
+++ b/plugins/dmxusb/src/dmxusb.cpp
@@ -155,10 +155,10 @@ QString DMXUSB::outputInfo(quint32 output)
         if (m_outputs.size() == 0)
         {
             str += QString("<BR><B>%1</B>").arg(tr("No output support available."));
-            str += QString("<P>");
+            /*str += QString("<P>");
             str += tr("Make sure that you have your hardware firmly plugged in. "
                       "NOTE: FTDI VCP interface is not supported by this plugin.");
-            str += QString("</P>");
+            str += QString("</P>");*/
         }
     }
     else if (output < quint32(m_outputs.size()))
@@ -201,6 +201,7 @@ bool DMXUSB::openInput(quint32 input, quint32 universe)
         DMXUSBWidget *widget = m_inputs.at(input);
         if (widget->type() == DMXUSBWidget::ProRXTX ||
             widget->type() == DMXUSBWidget::ProMk2 ||
+            widget->type() == DMXUSBWidget::OpenRX ||
             widget->type() == DMXUSBWidget::UltraPro)
         {
             EnttecDMXUSBPro *pro = static_cast<EnttecDMXUSBPro*>(widget);
--- a/plugins/dmxusb/src/dmxusb.cpp
+++ b/plugins/dmxusb/src/dmxusb.cpp
@@ -155,10 +155,10 @@ QString DMXUSB::outputInfo(quint32 output)
         if (m_outputs.size() == 0)
         {
             str += QString("<BR><B>%1</B>").arg(tr("No output support available."));
-            /*str += QString("<P>");
+            str += QString("<P>");
             str += tr("Make sure that you have your hardware firmly plugged in. "
                       "NOTE: FTDI VCP interface is not supported by this plugin.");
-            str += QString("</P>");*/
+            str += QString("</P>");
         }
     }
     else if (output < quint32(m_outputs.size()))
--- a/plugins/dmxusb/src/dmxusbopenrx.cpp
+++ b/plugins/dmxusb/src/dmxusbopenrx.cpp
@@ -54,22 +54,6 @@ DMXUSBOpenRx::DMXUSBOpenRx(DMXInterface *interface,
     m_inputLines[0].m_universeData = QByteArray();
     m_inputLines[0].m_compareData = QByteArray();
 
-    /*QSettings settings;
-    QVariant var = settings.value(SETTINGS_CHANNELS);
-    if (var.isValid() == true)
-    {
-        int channels = var.toInt();
-        if (channels > DMX_CHANNELS || channels <= 0)
-            channels = DMX_CHANNELS;
-        // channels + 1 Because the first byte is always zero
-        // to break a full DMX universe transmission
-        m_outputLines[0].m_universeData = QByteArray(channels + 1, 0);
-    }
-    else
-    {
-        m_outputLines[0].m_universeData = QByteArray(DMX_CHANNELS + 1, 0);
-    }*/
-
 // on macOS, QtSerialPort cannot handle an OpenDMX device
 // so, unfortunately, we need to switch back to libftdi
 #if defined(Q_OS_OSX) && defined(QTSERIAL) && (defined(LIBFTDI1) || defined(LIBFTDI))
--- a/plugins/dmxusb/src/dmxusbopenrx.cpp
+++ b/plugins/dmxusb/src/dmxusbopenrx.cpp
@@ -1,0 +1,314 @@
+/*
+  Q Light Controller
+  dmxusbopenrx.cpp
+
+  Copyright (C) Heikki Junnila
+                Christopher Staite
+                Emmanuel Coirier
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0.txt
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+#include <QSettings>
+#include <QDebug>
+#include <math.h>
+#include <QTime>
+
+#include "dmxusbopenrx.h"
+#include "qlcmacros.h"
+
+#define DMX_MAB 16
+#define DMX_BREAK 110
+#define DMX_CHANNELS 512
+#define DEFAULT_OPEN_DMX_FREQUENCY    30  // crap
+#define SETTINGS_CHANNELS "dmxusbopenrx/channels"
+
+/****************************************************************************
+ * Initialization
+ ****************************************************************************/
+
+DMXUSBOpenRx::DMXUSBOpenRx(DMXInterface *interface,
+                                   quint32 inputLine, QObject* parent)
+    : QThread(parent)
+    , DMXUSBWidget(interface, 0, DEFAULT_OPEN_DMX_FREQUENCY)
+    , m_running(false)
+    , m_granularity(Unknown)
+    , m_reader_state(Idling)
+{
+    qDebug() << "Open RX contructor, line" << inputLine;
+
+    m_inputBaseLine = inputLine;
+    setOutputsNumber(0);
+    setInputsNumber(1);
+
+    m_inputLines[0].m_universeData = QByteArray();
+    m_inputLines[0].m_compareData = QByteArray();
+
+    /*QSettings settings;
+    QVariant var = settings.value(SETTINGS_CHANNELS);
+    if (var.isValid() == true)
+    {
+        int channels = var.toInt();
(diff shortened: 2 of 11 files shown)


enbyted:
Tested with my custom FTDI based adapter.
Appears to be working fine, I was actually looking for functionality like that few month ago.

plugins/dmxusb/src/dmxusb.cpp
```diff
@@ -155,10 +155,10 @@ QString DMXUSB::outputInfo(quint32 output)
         if (m_outputs.size() == 0)
         {
             str += QString("<BR><B>%1</B>").arg(tr("No output support available."));
-            str += QString("<P>");
+            /*str += QString("<P>");
```
mcallegari: Why?
Please uncomment.

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,298 @@
+/*
+  Q Light Controller
```
mcallegari: Q Light Controller Plus.
And I doubt Heikki has anything to do with a new file. Write my name instead cause sooner or later I'll edit this file.

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,298 @@
...
+    return DMXUSBWidget::OpenRX;
+}
+
+/****************************************************************************
+ * Open & Close
+ ****************************************************************************/
+
+bool DMXUSBOpenRx::open(quint32 line, bool input)
+{
+    if (input == false) {
```
mcallegari: There's a whole mix of parenthesis indentation in this file
Please use only one, same as the rest of the code. Parenthesis on new line. Thanks.

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,298 @@
...
+bool DMXUSBOpenRx::writeUniverse(quint32 universe, quint32 output, const QByteArray& data)
+{
+    Q_UNUSED(universe)
+    Q_UNUSED(output)
+    Q_UNUSED(data)
+    // not used at runtime but needed for compilation
+    return true;
+}
+
+
```
mcallegari: extra newline. remove

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,298 @@
...
+    QByteArray payload;
+    QByteArray& last_payload = m_inputLines[0].m_compareData;
+    QByteArray& current_payload = m_inputLines[0].m_universeData;
+
+    quint32 missed_frames = 0;
+    quint32 erroneous_frames = 0;
+
+    m_frameTimeUs = 0;
+
```
mcallegari: extra newline.

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,298 @@
...
+            if (current_payload.length() != last_payload.length() && erroneous_frames < 5) {
+                qDebug() << "Bogus frame" << current_payload.length() << "bytes instead of" << last_payload.length();
+                current_payload.clear();
+                erroneous_frames += 1;
+                continue;
+            }
+
+            // a frame has been received
+
```
mcallegari: extra newline

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,298 @@
...
+            current_payload.clear();
+        }
+
+        if (missed_frames == 300) {
+            m_reader_state = Idling;
+            qDebug() << "Idling";
+        } else if (missed_frames == UINT_MAX) {
+            missed_frames = 300;
+        }
+
```
mcallegari: extra newline

plugins/dmxusb/src/dmxusbopenrx.h
```diff
@@ -0,0 +1,100 @@
+/*
+  Q Light Controller
+  dmxusbopenrx.h
+
+  Copyright (C) Heikki Junnila
```
mcallegari: same as before.

plugins/dmxusb/src/libftdi-interface.cpp
```diff
@@ -242,6 +242,10 @@ bool LibFTDIInterface::open()
     }
     else
     {
+        if (ftdi_set_latency_timer(&m_handle, 1)) {
```
mcallegari: This block...just no.
Changing the main FTDI interface class is VERY risky. I experimented it at my expenses.
It may work in some case and it may break some other.
Please move this in your subclass of introduce a new interface API.

plugins/dmxusb/src/dmxusb.cpp
manuco: The purpose of the comment was to not mislead the user with this message, since this message will always be displayed even if all hardawre is correctly plugged.

I'll remove the comment, but perhaps a reword of this notice will be needed.

plugins/dmxusb/src/libftdi-interface.cpp
manuco: I'm ok with the fact that it should be optional.

The problem is that this command is specific with libftdi, and is clearly a hack needed for some devices. As a hack, it should be isolated, but this isolation shouldn't be hacky.

I have no clue on how QtSerialInterface or ftd2xx libraries would work with or without tuning, but I'm pretty sure that the answer won't be the same for all interfaces.

So I suggest this :
- add a new API called "set_low_latency(boolean)" that will enable wathever it needs to let the device be more reactive, as a virtual method
- add a common implementation in DmxInterface that will do nothing
- move my code in a reimplementation method in LibFTDIInterface class
- call this API in my plugin

This will let some kind of placeholder in others implementations if they need some kind of modifications too, while not disturbing anything else for others plugins.

manuco: Let say it's yes. :-)

manuco:
I Massimo,

You can test this feature with any DMX controller if you want. Even a grandMa 😊 I think any cheap USB DMX interface will do the job. 

Can you review my new commits ?

plugins/dmxusb/src/dmxusbopenrx.cpp
```diff
@@ -0,0 +1,338 @@
...
+            if (current_payload.length() != last_payload.length() && erroneous_frames < 5)
+            {
+                qDebug() << interface()->serial() << "Bogus frame" << current_payload.length() << "bytes instead of" << last_payload.length();
+                current_payload.clear();
+                erroneous_frames += 1;
+                continue;
+            }
+
+            // a frame has been received
+            if (missed_frames > 300) // only to emit the debug message once, not at each frame
```
mcallegari: Are you sure this logic is correct ?
missed_frames is reset to 0 two lines below. To me it looks like missed_frames will never reach 300.
Also 
- maybe "missed_frames" is not a proper name since the comment is "a frame has been received"
- "300", "600" and other magic numbers should be placed as defines at the beginning of the cpp file so we can track what they are and easily change them if needed

manuco:
Implementing setLowLatency() in ftd2xx-interface is perhaps a thing to do, but I don't have a compilation stack on my windows setup for now. I see that I can retrieve the Appveyor's dev build. I'll test various configuration to see what can be done on Windows.

plugins/dmxusb/src/dmxusbopenrx.cpp
manuco: Yes, I'm pretty sure of the logic. But it needs more explanation:
- when we reach the comment "a frame has been received", it means that all the checks have succedded, so the current_payload contains a valid frame
- but if missed_frames is > 300, it means that there was more than 300 empty reads in the previous cycles
- so we are receiving a correct frame after a long break (probably because the attached controller was plugged or switched on)
- so we display this fact on the debug logger. The whole purpose of this condition is to display on the console the fact that the widget now receives frames, without cluttering the console with a message on each frame received.

You can see that counters have been reset just after that.

I'll add this as a comment at this place to clarify this thing.

manuco: For the magic numbers, you are right.

I'll move them, and explain their semantics.

manuco:
I'll plan to make these modifications next Monday evening.

mcallegari:
Thanks for the explanations. Makes sense now.
I can merge this as it is and implement D2XX latency myself.
Then you can send a further PR with more changes. Sounds ok ?

manuco:
Ok for the merge now. I'll do a new PR later. Thanks for all  :-)

aussie-katie:
Dumb question, does this mean Open RX was merged into QLC+ and now I can use any FTDI DMX adapter as an input? Or an I misunderstanding the purpose of this?

manuco:
Open rx is merged. But there are other things to do before using this feature. 

Some adapters (i'm not sure that all can do that) have the capability to read signal from the DMX bus. In order for this to work, either you have the right socket on it, or you can use a female/female adapter to plug your adapter into an out port of the DMX line. 

Once this hack is made, you can use this plugin. In the driver window, you will see some information about the signal. If you don't see the frequency of the signal, it is probably because your device is not able to read anything. Another issue is that the signal is too poorly sequenced on the line to be read accurately by the software. 

Open rx is a hacky pluggin designed to use very cheap devices. It's not a first class solution.

DarkMagicSource:
Currently I'm testing with an _Enttec USB Assembled Widget_, and plan to grab a M/M XLR5 adapter from Enttec, more long term if this works I plan to either buy another USB/DMX adapter or build one, as my Enttec one is my primary output one.

omichel:
I am trying to make this work, to no avail so far. I left the frequency to the default value (30Hz). Here is what I get in the driver window (assuming this is what you mean by driver window):
image
Moving the sliders on my DMX console doesn't seem to do anything (I am trying to send values on channel 1 using slider 1 on scanner 1). How can I debug this?

Here is my hardware: FTDI FT232RL CHIP USB TO RS485 DMX512
image
image
image

manuco:
Hi @omichel 

Given the datasheet of your device, you should be able to read something from the bus. The adapter I'm using was able to do this and was built with two leds which helped me diagnose its modes. But the vendor who sold me those units told me he no longer does that. Yours doesn't seem to have control leds, so it won't be easy for you to see if it's reading something from the bus, or if it's reading something but not sending it to the host computer. I've tried mine with two setups: a cheap 6-slider console and another USB/DMX device controlled by QLC+. Maybe you could try another device to see if you're more successful?

Idling means that no data has been received from the adapter for the last sampling windows. The computer host sees your adapter, but the adapter doesn't send any data (read from the bus).

In receive mode, the frequency is read from the remote device and you can't change it from the computer side. It's your console that will write data at a specific rate.

omichel:
Hi @manuco, thank you for the quick reply.

I am on Windows 11. Is it supposed to work there? It does work well in "Open TX" mode and I can control my PAR LEDs without any problem, but the "Open RX" mode with my DMX console doesn't seem to work, although the DMX console works well when connected to the fixtures directly. The "Receiver state:" remains to "Idling" whatever I do... I tried to monitor the virtual COM port corresponding to the device, but I got no information, so I assume the interface doesn't use the VPC, but the FTD2XX.DLL interface instead. Not sure how to debug this...

omichel:
Could it be due to [my XLR male/male adapter](https://www.aliexpress.com/item/1005002388251734.html) as they say:

> This adaptor is designed to change an XLR input into an XLR output.

I checked how it is wired and I noticed the left and right pin are inverted:

```
input        output
1   2        2   1
o   o        o   o
  o            o
  3            3
```

Should I rewire it like that:?
```
input        output
1   2        1   2
o   o        o   o
  o            o
  3            3
```

omichel:
I received it today, but sadly, it doesn't work (exact same results as before).

manuco:
I'm sorry, but I still have no idea. I don't think I can help you without some deep debugging information. And if I remember correctly, I developed this plugin on a Linux machine. Doing hadware debugging on Windows won't be the easiest thing.

The "idling" message indicates that no data is being received from the driver.

Have you tried plugging a TX device into an RX device to see if something different happens?

AlfaWita:
Hello everyone, I hope you're all safe and sound. The same problem persists with Windows 11 and Enttec DMX USB Pro. No matter what I do, setting Pro Tx/RX or just Pro Rx, changing the packet size, there's no reception, and Idle is displayed. I only wanted to receive DMX and pass it through to Artnet, but it's not possible. Sending and everything else works. Best regards to the developers for their great work. Just solve the one thing with the reception. That would be great. Best regards, Andres.

PR #1183, opened 2019-06-29
Sorting of ArtNet interfaces implemented.

ambraspace:
QNetworkInterfaces::allInterfaces() does not guarantee the order of interfaces, which can cause problems if you have many interfaces. QXW file specifies only the "line number" of the selected interface, so sometimes it can be some other interface selected different from what you set up.
For example, I have a Raspberry Pi with 6 interfaces: localhost, ethernet, wlan (hostapd, so that the customer can connect and change the scene), 2 interfaces for 3G connection, and tun (OpenVPN connection). ArtNet device is on ethernet. But, sometimes, if I restart the RPi some other interface can be selected. If that happens with wlan or 3g QLCplus floods that interface with ArtNet packet rendering that interface completely unable to respond (no 3g connection or customer not able to connect to WiFi AP). Yes, flooding of the interface can be prevented with iptables, but the problem of not sending ArtNet packets to proper interface stays.

### DIFF
--- a/plugins/artnet/src/artnetplugin.cpp
+++ b/plugins/artnet/src/artnetplugin.cpp
@@ -27,7 +27,7 @@
 
 bool addressCompare(const ArtNetIO &v1, const ArtNetIO &v2)
 {
-	return v2.address.ip().toString() < v1.address.ip().toString();
+	return v1.address.ip().toString() < v2.address.ip().toString();
 }
 
 
--- a/plugins/artnet/src/artnetplugin.cpp
+++ b/plugins/artnet/src/artnetplugin.cpp
@@ -24,6 +24,13 @@
 
 #define MAX_INIT_RETRY  10
 
+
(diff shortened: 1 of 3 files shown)


plugins/artnet/src/artnetplugin.cpp
```diff
@@ -24,6 +24,13 @@
 
 #define MAX_INIT_RETRY  10
 
+
+bool addressCompare(const ArtNetIO &v1, const ArtNetIO &v2)
+{
+	return v1.address.ip().toString() < v2.address.ip().toString();
```
mcallegari: 4 spaces tab please

mcallegari:
Thanks for this patch.
I think this can be applied to OSC and E1.31 plugin too, since they share the same interface listing code.

mcallegari:
Thanks a lot! :+1:

DanielGibbsNZ:
Looking forward to this, any ideas for when it will be released?

PR #1186, opened 2019-07-10
Don't crash when the RGB script to be parsed is empty. Fixes #1185

### LINKED ISSUE #1185: QMLUI crashes when loading the 3D demo project
QLC+ 5 crashes when loading the 3D demo project found in the forum. This happens because QLC+ tries to parse an emptry RGBScript. I haven't found the root cause (why the script is empty in the first place), but I will send a PR fixing at least the crash

kripton:
See description in #1185


mcallegari:
Thanks :+1:

PR #1187, opened 2019-07-16
Add a way to hide the disclaimer in order to ease development

kripton:
With this change, the disclaimer can be hidden. This should only be done by developers in order to not have to click the disclaimer away when testing out code changes.
I wouldn't document this feature much, developers who want to use this feature will know how to activate it. All others will have to live with the disclaimer ;)


mcallegari:
No thanks.
Just comment out PopupDisclaimer. That's all.

kripton:
allright, commenting out that line should also do it for me

PR #1190, opened 2019-08-25
Update license header of hidapi library

jeromelebleu:
This update the license header of the hidapi library from upstream which is much clearer. It is needed for the upcoming Debian package (see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=694077). Thanks in advance!

### DIFF
--- a/plugins/hid/hidapi.h
+++ b/plugins/hid/hidapi.h
@@ -1,15 +1,24 @@
-/*
- * HIDAPI - Multi-Platform library for
- * communication with HID devices.
- *
- * Copyright (c) 2010 Alan Ott - Signal 11 Software
- *
- * 2010-07-03
- *
- * This software may be used by anyone for any reason so
- * long as the copyright notice in the source files
- * remains intact.
- */
+/*******************************************************
+ HIDAPI - Multi-Platform library for
+ communication with HID devices.
+
+ Alan Ott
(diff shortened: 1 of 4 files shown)


mcallegari:
Look OK to me, thanks :+1: 

Thanks also for the huge work you're doing on Debian.
Do you want me to open an issue here so that we can discuss and you can ask questions?
I already know there are a few things to change to make the debian package cross architecture. (e.g. it's not ok on arm out of the box)

jeromelebleu:
> Thanks also for the huge work you're doing on Debian.

It is a pleasure to do that, especially regarding all the huge work you are doing on this great software! It is also thanks to Lisandro who is reviewing and sponsoring this package.

> Do you want me to open an issue here so that we can discuss and you can ask questions?
> I already know there are a few things to change to make the debian package cross architecture. (e.g. it's not ok on arm out of the box)

As you want, it could be indeed an easy way to track the progress of the upcoming Debian package and discuss possible build issues. With this one fixed, I think Lisandro should soon upload a first release to unstable! By the way, when do you plan to release 4.12.2 please? Is this patch going to be included?

jeromelebleu:
Also, as you are free to choose, are you okay to license those files under the terms of the GNU General Public License v3?

PR #1196, opened 2019-10-07
Improve RGB scripts devtool

fxedel:
* Properly support primary and secondary colors (and hide input fields, depending on `algo.acceptColors`)
* Fixed issue when "Start test" has been clicked multiple times
* Fixed missing `#` in `cell.style.backgroundColor` which lead to sometimes uncolered cells
* Add title to cells so you can easily check a specific cell's position and color
* Use modern HTML form element features (like input type number, min/max)
* Refactor JS logic, e.g. by using global variables instead of always reading input element values

I'm also thinking about adding support for API version 2 parameters – do you want to wait for this first or should I open a new PR then?

### DIFF
--- a/resources/rgbscripts/devtool.html
+++ b/resources/rgbscripts/devtool.html
@@ -1,7 +1,7 @@
 <HTML>
 <HEAD>
 <TITLE>Q Light Controller Plus - RGBScript Development Tool</TITLE>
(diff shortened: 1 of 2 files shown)


mcallegari:
Thanks for this! :+1: 
It definitely needed some fresh air. Let's merge this for now.
If you want to add parameters support, you're very welcome.

PR #1201, opened 2019-11-08
Set E1.31 multicast address from universe number

sparkyb:
Right now when you set an E1.31 output to multicast, it only lets you edit the last octet of the multicast address. The E1.31 multicast address is supposed to match the universe number. By only allowing editing of the last octet, this limits QLC+ to using only the first 255 E1.31 universes, and not universes 256-63999. Also, by requiring the multicast address to be entered manually, it requires type the universe number twice to have it correctly match the universe number, and allows likely erroneous mismatches of the address and universe number. This patch makes the multicast address uneditable and sets it automatically to make the universe number, correctly handling all 64,000 universes.

### DIFF
--- a/plugins/E1.31/configuree131.cpp
+++ b/plugins/E1.31/configuree131.cpp
@@ -112,8 +112,7 @@ void ConfigureE131::fillMappingTree()
                 if (info->inputMulticast)
                 {
                     multicastCb->setChecked(true);
-                    m_uniMapTree->setItemWidget(item, KMapColumnIPAddress,
-                            createMcastIPWidget(info->inputMcastAddress.toString()));
+                    item->setText(KMapColumnIPAddress, info->inputMcastAddress.toString());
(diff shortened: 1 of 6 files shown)


plugz:
Hi,

Yes, being able to send to mcast addresses for universes more than 255 would be nice.
However, I like being able to manually change the multicast address.
I have a weird use case, but I actually use this feature (target devices listen to 1 mcast address, but handle more than 1 universe)
What about adding a "guess from universe number" button in the E1.31 config menu ? When clicking it, the address would be automagically changed.

sparkyb:
@plugz Do you only need this for output or also input? Because for output you can actually just uncheck "multicast" and enter a multicast address as the unicast address and there is no difference. That trick doesn't work for input though.

mcallegari:
See cb3cc67540f15f84391e2666e440c905928070f0

PR #1202, opened 2019-11-08
Hide disconnected network interfaces

sparkyb:
Hides network interfaces that are not connected from the ArtNet, E1.31, and OSC interface choices.
Also disables the E1.31 multicast option for interfaces that don't support multicast.

### DIFF
--- a/plugins/E1.31/configuree131.cpp
+++ b/plugins/E1.31/configuree131.cpp
@@ -125,6 +125,8 @@ void ConfigureE131::fillMappingTree()
                     portSpin->setValue(info->inputUcastPort);
                     m_uniMapTree->setItemWidget(item, KMapColumnPort, portSpin);
                 }
+                if (!controller->canMulticast())
+                    multicastCb->setEnabled(false);
                 connect(multicastCb, SIGNAL(clicked()), this, SLOT(slotMulticastCheckboxClicked()));
                 m_uniMapTree->setItemWidget(item, KMapColumnMulticast, multicastCb);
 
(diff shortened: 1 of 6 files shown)


mcallegari:
Hi, thanks for the patch.
I thought about this and I don't think this is a good idea, even though sometimes many inactive interfaces might be confusing or invasive.
The case is:
- open a QLC+ project with some Artnet lines previously working
- for some reason, a network switch is powered off
- in QLC+ you don't see the interface because it's disconnected
- you power on the switch
- you still don't see the interface in QLC+

Bad user experience.
Moreover, if you don't realize this in time and make changes to the project and save, you risk to completely lose the output settings, just because the network lines are not available.

Makes sense?

sparkyb:
Yeah, this makes sense. Perhaps there is a way to monitor for interfaces becoming connected so that after you power on the switch in your example above, the interface would show up, but even if I did that, I think your point still stands. Might be nice to gray disconnected ones out or something (I have 8, several of which will NEVER be connected to anything and it makes it hard to easily find the correct ones). But ultimately this doesn't really address my real problem anyway which is very much related to your last point about risking losing the output settings. I lose output settings all the time as it is because interfaces appear and disappear or change order so the wrong one ends up getting selected and you can't change which one is checked without losing the settings. The fix I really want is something that saves the settings per-plugin, not per-line, regardless of which line is selected (or even if none are selected). I started working on a patch that does that, but haven't gotten very far yet.

yestalgia:
Even if it was something as simple as a "Network Status" icon to indicate for each ArtNet Output which interfaces were active. This was an interesting idea though.

netmindz:
@sparkyb - it might be an idea to actually disable the network interfaces you don't need at an OS level.

netmindz:
Personally, if there were to be any changes to the way network interfaces were handled, I would consider the relationship between physical interface and IP address/subnet and what users would expect to see.
 
For example, i don't really care about my source ip, but more being on the right network. For example, one day my laptop might pick up 192.168.1.20 from my wifi and 10.0.10.13 from my wired connection and it's the 10.0.10.0/24 network that my lighting fixtures are on. If the next day i get .25 and .15 from the to networks, it's still the 10.0.0.10.0/24 network I want to send my data out on.

However, for other users, they might want to always use their usb 2 ethernet adapter for their output, no matter what the ip range is at that location, so can't simply select output based on subnet

mcallegari:
Closing then

PR #1205, opened 2020-01-20
a better widget delete dialog if a widget has no label

sammuell:


mcallegari:
First, it's a good practice to explain a change, not just throw it.
At least reference a forum discussion or a github ticket.
Second, you haven't followed the code convention used in other thousands lines of code.

This is not the way to contribute.

sammuell:
For the reference, here's the associated thread in the forum: https://www.qlcplus.org/forum/viewtopic.php?f=35&t=14070

mcallegari:
Thanks

mcallegari:
@sammuell Actually, the guy is right
https://www.qlcplus.org/forum/viewtopic.php?f=18&t=14138

I think it's better to separate with "<br>" (not sure if "\n" will work)

PR #1207, opened 2020-01-23
preserve aspect ratio of background image in virtual console

sammuell:
Forum thread: https://www.qlcplus.org/forum/viewtopic.php?f=35&t=14140


mcallegari:
Ok, thanks

PR #1208, opened 2020-01-25
vc: fix mininimum slider width

sammuell:
https://www.qlcplus.org/forum/viewtopic.php?f=35&t=14072


mcallegari:
Rejected, because you insist of wanting to change the code without studying it first.
QML is not the right place to change the default widget height/width.
https://github.com/mcallegari/qlcplus/blob/master/qmlui/virtualconsole/vcframe.cpp#L283

PR #1209, opened 2020-01-25
vc: add a context menu to widgets (copy/paste/delete)

sammuell:
Adding a basic context menu to the vc. www.qlcplus.org/forum/viewtopic.php?f=35&t=14071

### DIFF
--- a/engine/src/mastertimer-unix.cpp
+++ b/engine/src/mastertimer-unix.cpp
@@ -64,14 +64,14 @@ int MasterTimerPrivate::compareTime(struct timespec *time1, struct timespec *tim
 {
     if (time1->tv_sec < time2->tv_sec)
     {
-        qDebug() << "Time is late by" << (time2->tv_sec - time1->tv_sec) << "seconds";
+        //qDebug() << "Time is late by" << (time2->tv_sec - time1->tv_sec) << "seconds";
         return -1;
     }
     else if (time1->tv_sec > time2->tv_sec)
         return 1;
     else if (time1->tv_nsec < time2->tv_nsec)
     {
-        qDebug() << "Time is late by" << (time2->tv_nsec - time1->tv_nsec) << "nanoseconds";
+        //qDebug() << "Time is late by" << (time2->tv_nsec - time1->tv_nsec) << "nanoseconds";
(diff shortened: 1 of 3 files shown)


mcallegari:
The idea is not to use right click menus. So I don't support this change.
v5 is designed to work on mobiles (e.g touchscreens) where you can't possibly right click something!

mcallegari:
I'm not sure what's your plan here, but maybe you should ask before spending your time on these changes.

mcallegari:
And you don't touch the engine.

Lichtschalter-5000:
> The idea is not to use right click menus. So I don't support this change.
> v5 is designed to work on mobiles (e.g touchscreens) where you can't possibly right click something!

Why though? I get your point - you want to preserve usability on all possible devices.
But thereby you're chiseling the possibility to work fast and convenient out of all the users (and I guess this ist the biggest part) with the ability to perform a RMB! See[ this forum thread](https://www.qlcplus.org/forum/viewtopic.php?t=12968)...
Would you state your opinion to that topic, please?

mcallegari:
> Would you state your opinion to that topic, please?

No, I won't, cause you guys are speculating on something unfinished, and I don't have time to waste on pointless discussions.
Has anyone ever asked me what are my plans? No.
If someone did (about context menus) I would have replied: key combinations on desktop, buttons on touchscreens. Where key management is something I completely haven't started to handle yet.
Is it really difficult to ask instead of drawing pointless conclusions?

Lichtschalter-5000:
Sorry, in my opinion you discarded any other suggestions right away with the argument, they couldn't be used with all interfaces. 
So is your statement that alternative methods are still WIP?
My apologies to proceed with the discussion, but in my humble opinion, I did neither draw pointless conclusion nor did I not ask.
- I posted the thread in the issue-reports-thread. In my eyes this equals a question, at least it's a suggestion that something is wrong. A simple "WIP" would have resolved the conflict.
- I did not conclude anything. Based of what you said yourself, I pointed out what I thus thought you would be doing. It seems like I didn't understand what you wanted to make clear or you failed to make it clear in the first place, probably something of both.

PR #1212, opened 2020-02-18
FFT: Fixed mono and stereo audio capture triggers based on frequency spectrum

zamaudio:
As per this thread https://www.qlcplus.org/forum/viewtopic.php?f=24&t=10301&start=10 

I have also observed an issue with the FFT part that analyses the frequencies and attempts to plot the power of the signal for different bands and trigger various functions based on the audio signal levels.  Indeed the levels are appearing too low, and for good reason.  The key problem I have identified is that the power of the signal overall should be the sum of squares of the frequency spectra magnitudes, but currently the power is estimated by finding the sum of the absolute raw signal values.

I would like to start a conversation about this part of the code, so I have changed the bits that I think are critical, but may have made mistakes regarding the bar graph as I don't yet understand the scaling and how the data is plotted/stored in QT.  Please bear with me and we can fix this problem together.

### DIFF
--- a/engine/audio/src/audiocapture.cpp
+++ b/engine/audio/src/audiocapture.cpp
@@ -30,9 +30,7 @@
 
 #define USE_HANNING
 #define CLEAR_FFT_NOISE
-
 #define M_2PI       6.28318530718           /* 2*pi */
-#define SQRT_32768  181.0193
 
 AudioCapture::AudioCapture (QObject* parent)
     : QThread (parent)
@@ -162,7 +160,7 @@ double AudioCapture::fillBandsData(int number)
 
     for (int b = 0; b < number; b++)
     {
-        quint64 magnitudeSum = 0;
+        double magnitudeSum = 0.;
         for (int s = 0; s < subBandWidth; s++, i++)
         {
             if (i == bufferSize)
@@ -170,7 +168,7 @@ double AudioCapture::fillBandsData(int number)
             magnitudeSum += qSqrt((((fftw_complex*)m_fftOutputBuffer)[i][0] * ((fftw_complex*)m_fftOutputBuffer)[i][0]) +
                                   (((fftw_complex*)m_fftOutputBuffer)[i][1] * ((fftw_complex*)m_fftOutputBuffer)[i][1]));
         }
-        double bandMagnitude = (magnitudeSum / subBandWidth) * SQRT_32768;
+        double bandMagnitude = (magnitudeSum / (subBandWidth * M_2PI));
         m_fftMagnitudeMap[number].m_fftMagnitudeBuffer[b] = bandMagnitude;
         if (maxMagnitude < bandMagnitude)
             maxMagnitude = bandMagnitude;
@@ -245,7 +243,7 @@ void AudioCapture::processData()
         {
             pwrSum += m_fftMagnitudeMap[barsNumber].m_fftMagnitudeBuffer[n];
         }
-        m_signalPower = pwrSum * qSqrt(M_2PI) * SQRT_32768 / barsNumber;
+        m_signalPower = 32768 * pwrSum * qSqrt(M_2PI) / (double)barsNumber;
         emit dataProcessed(m_fftMagnitudeMap[barsNumber].m_fftMagnitudeBuffer.data(),
                            m_fftMagnitudeMap[barsNumber].m_fftMagnitudeBuffer.size(),
                            maxMagnitude, m_signalPower);
--- a/engine/audio/src/audiocapture.cpp
+++ b/engine/audio/src/audiocapture.cpp
@@ -42,10 +42,11 @@ AudioCapture::AudioCapture (QObject* parent)
     , m_sampleRate(0)
     , m_channels(0)
     , m_audioBuffer(NULL)
+    , m_audioMixdown(NULL)
     , m_fftInputBuffer(NULL)
     , m_fftOutputBuffer(NULL)
 {
-    int bufferSize = AUDIO_DEFAULT_BUFFER_SIZE;
+    bufferSize = AUDIO_DEFAULT_BUFFER_SIZE;
     m_sampleRate = AUDIO_DEFAULT_SAMPLE_RATE;
     m_channels = AUDIO_DEFAULT_CHANNELS;
 
@@ -65,9 +66,10 @@ AudioCapture::AudioCapture (QObject* parent)
     m_captureSize = bufferSize * m_channels;
 
     m_audioBuffer = new int16_t[m_captureSize];
-    m_fftInputBuffer = new double[m_captureSize];
+    m_audioMixdown = new int16_t[bufferSize];
+    m_fftInputBuffer = new double[bufferSize];
 #ifdef HAS_FFTW3
-    m_fftOutputBuffer = fftw_malloc(sizeof(fftw_complex) * m_captureSize);
+    m_fftOutputBuffer = fftw_malloc(sizeof(fftw_complex) * bufferSize);
 #endif
 }
 
@@ -77,6 +79,7 @@ AudioCapture::~AudioCapture()
     Q_ASSERT(!this->isRunning());
 
     delete[] m_audioBuffer;
+    delete[] m_audioMixdown;
     delete[] m_fftInputBuffer;
 #ifdef HAS_FFTW3
     if (m_fftOutputBuffer)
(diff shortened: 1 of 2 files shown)


engine/audio/src/audiocapture.cpp
```diff
@@ -151,10 +153,10 @@ double AudioCapture::fillBandsData(int number)
     // representing all the frequencies from 0 to m_sampleRate Hz.
     // I will just consider 0 to 5000Hz and will calculate average magnitude
     // for the number of desired bands.
-    double maxMagnitude = 0;
+    double bandMagnitude = 0;
 #ifdef HAS_FFTW3
-    unsigned int i = 0;
-    int subBandWidth = ((m_captureSize * SPECTRUM_MAX_FREQUENCY) / m_sampleRate) / number;
+    unsigned int i = 1; // skip DC bin
```
zamaudio: This should definitely be 1 because the zero bin contains the DC offset and there is no point finding its power.

engine/audio/src/audiocapture.cpp
```diff
@@ -18,6 +18,8 @@
   limitations under the License.
 */
 
+#include <stdio.h>
```
zamaudio: I added this temporarily so I could use printf for hacky debug

engine/audio/src/audiocapture.cpp
```diff
@@ -151,10 +153,10 @@ double AudioCapture::fillBandsData(int number)
     // representing all the frequencies from 0 to m_sampleRate Hz.
```
zamaudio: I believe this function is supposed to fill the data for only a single band numbered by `number`

engine/audio/src/audiocapture.cpp
```diff
@@ -151,10 +153,10 @@ double AudioCapture::fillBandsData(int number)
...
     // for the number of desired bands.
-    double maxMagnitude = 0;
+    double bandMagnitude = 0;
 #ifdef HAS_FFTW3
-    unsigned int i = 0;
-    int subBandWidth = ((m_captureSize * SPECTRUM_MAX_FREQUENCY) / m_sampleRate) / number;
+    unsigned int i = 1; // skip DC bin
+    int subBandWidth = ((m_captureSize * SPECTRUM_MAX_FREQUENCY) / m_sampleRate) / defaultBarsNumber();
 
     for (int b = 0; b < number; b++)
```
zamaudio: I don't understand why this loop is here, surely you don't mean to compute the magnitude for all bands from zero up to the `number`eth band ?

engine/audio/src/audiocapture.cpp
```diff
@@ -151,10 +153,10 @@ double AudioCapture::fillBandsData(int number)
     // representing all the frequencies from 0 to m_sampleRate Hz.
     // I will just consider 0 to 5000Hz and will calculate average magnitude
     // for the number of desired bands.
-    double maxMagnitude = 0;
+    double bandMagnitude = 0;
 #ifdef HAS_FFTW3
-    unsigned int i = 0;
-    int subBandWidth = ((m_captureSize * SPECTRUM_MAX_FREQUENCY) / m_sampleRate) / number;
+    unsigned int i = 1; // skip DC bin
+    int subBandWidth = ((m_captureSize * SPECTRUM_MAX_FREQUENCY) / m_sampleRate) / defaultBarsNumber();
```
zamaudio: I don't know how many bands there are in total, but this should be divided by the total user bands, not by the current band you are computing the magnitude for.

engine/audio/src/audiocapture.cpp
```diff
@@ -166,22 +168,21 @@ double AudioCapture::fillBandsData(int number)
             magnitudeSum += qSqrt((((fftw_complex*)m_fftOutputBuffer)[i][0] * ((fftw_complex*)m_fftOutputBuffer)[i][0]) +
                                   (((fftw_complex*)m_fftOutputBuffer)[i][1] * ((fftw_complex*)m_fftOutputBuffer)[i][1]));
         }
-        double bandMagnitude = (magnitudeSum / subBandWidth);
+        bandMagnitude = (magnitudeSum / subBandWidth);
         m_fftMagnitudeMap[number].m_fftMagnitudeBuffer[b] = bandMagnitude;
-        if (maxMagnitude < bandMagnitude)
-            maxMagnitude = bandMagnitude;
```
zamaudio: I removed the maxMagnitude computation so we can compute it later and keep the result of the current band instead in the return value.

engine/audio/src/audiocapture.cpp
```diff
@@ -166,22 +168,21 @@ double AudioCapture::fillBandsData(int number)
...
 #endif
-    return maxMagnitude;
+    return bandMagnitude;
 }
 
 void AudioCapture::processData()
 {
 #ifdef HAS_FFTW3
     unsigned int i;
-    quint64 pwrSum = 0;
```
zamaudio: I think this needs to be a floating point value, not an integer (?)

engine/audio/src/audiocapture.cpp
```diff
@@ -192,11 +193,6 @@ void AudioCapture::processData()
 
     for (i = 0; i < m_captureSize; i++)
     {
-        if(m_audioBuffer[i] < 0)
-            pwrSum += -1 * m_audioBuffer[i];
-        else
-            pwrSum += m_audioBuffer[i];
-
```
zamaudio: This block is the critical part I believe is just wrong.

engine/audio/src/audiocapture.cpp
```diff
@@ -227,12 +223,20 @@ void AudioCapture::processData()
...
 
     // 5 ********* Calculate the average signal power
-    m_signalPower = pwrSum / m_captureSize;
+    foreach(int barsNumber, m_fftMagnitudeMap.keys())
+    {
+        double bandMagnitude = fillBandsData(barsNumber);
+        pwrSum += bandMagnitude;
+        if (maxMagnitude < bandMagnitude)
+            maxMagnitude = bandMagnitude;
+    }
```
zamaudio: This block computes the total power in all bands (pwrSum) and also picks out the band with maximum magnitude.

engine/audio/src/audiocapture.cpp
```diff
@@ -227,12 +223,20 @@ void AudioCapture::processData()
...
+
+    m_signalPower = 15 * pwrSum;
+
+    //printf("XXX pwrSum=%f signalpower=%d maxMagn=%f\n", pwrSum, m_signalPower, maxMagnitude);
```
zamaudio: I wasn't sure how to scale the result and started dumping values to play with it

engine/audio/src/audiocapture.cpp
mcallegari: No need. You can use this way to print on console with Qt:
qDebug() << "Something" << var;

engine/audio/src/audiocapture.cpp
mcallegari: IIRC `number` is the amount of bars wanted by the user. The FFT is always performed on the whole spectrum but the signal strength visualization is based on sub-bands.
You can have up to 32. Default is 16.
So this is a mistake: you need to use `number`.

engine/audio/src/audiocapture.cpp
mcallegari: Sum of the power is used to calculate an average, so it's a very big number. I needed a 64 bit variable. I think float is not needed at this stage since `m_audioBuffer[i]` contains `int16_t` values.

engine/audio/src/audiocapture.cpp
mcallegari: 15 looks like a magic number to me (as well as 10, mentioned in the forums)
I believe, but I'm not sure, the level issue here is all about mic vs line input.
It's probably better to expose a parameter (maybe named 'gain'?) up to the UI, so users can compensate depending on their audio card/input.
Makes sense?

zamaudio: I disagree, I think the graph needs to be calibrated such that the maximum level that any sound card can output is 0dbFS (which corresponds to +- 32768 in 16 bit samples).  Can you tell me what the value is for the maximum of the graph, then I can work out what the magic number should be.

zamaudio:
I have force pushed a new patch that works well for mono capture signals.  I believe the samples are interleaved for stereo, but the FFT treats it as a flat buffer which is why it is computing wrong for stereo signals.

engine/audio/src/audiocapture.cpp
zamaudio: No, it seems to be computing all the bands for one whole bar graph simultaneously

engine/audio/src/audiocapture.cpp
zamaudio: Ok I see now

engine/audio/src/audiocapture.cpp
zamaudio: I realised that the FFT is supposed to be on values -1.0 to 1.0 for the signal, so I rescaled the input values.

engine/audio/src/audiocapture.cpp
```diff
@@ -227,12 +224,15 @@ void AudioCapture::processData()
...
     foreach(int barsNumber, m_fftMagnitudeMap.keys())
     {
-        double maxMagnitude = fillBandsData(barsNumber);
+        maxMagnitude = fillBandsData(barsNumber);
+        pwrSum = 0.;
+        for (int n = 0; n < barsNumber; n++)
+        {
+            pwrSum += m_fftMagnitudeMap[barsNumber].m_fftMagnitudeBuffer[n];
+        }
+        m_signalPower = pwrSum * 500. / barsNumber;
```
zamaudio: The 500 here is a magic value for calibrating the maximum value of the graph to the power values.  However I still don't know the scale of the graph so I am purely guessing here.

zamaudio:
I fixed stereo as well now, by mixing down the multiple interleaved channels to a mono signal, and performing FFT on that.

mcallegari:
Hi, thanks for this. Looks like it's in good shape to be merged now.
I "only" need to find some time and test it on all platforms (Windows, macOS, Linux)
Please be patient

zamaudio:
I tested it with a reference sweeping sine signal generator and played with it for a while, and I noticed a coarse rounding error, please see latest commit for fix below, will push now.

mcallegari:
Looks good on Windows and Linux. Thanks
Actually, on Windows it even seems more sensitive than Windows itself.
Never gotten to understand where Qt actually reads data on Windows...

PR #1214, opened 2020-02-26
move qlcplus to portaudio

Tuxist:
use professional api like asio or jack with portaudio

advantages
-low latency
-input works also under linux

future plans
-fine tunnig for qlc portaudio backend
-use ffmpeg instead qmultimedia for audio/video playback
-use cmake instead qmake

Video:
https://tuxist.de/files/qlcplus.mp4


mcallegari:

Tuxist:
Ok cmake the cmake thing i think will be finished this week i think adicional buildsystem will be ok but for the portaudio and ffmpeg is to discuss. I can  put the cmake things if it finished in seperate branch

mcallegari:
I received 27 emails from this pull request.
If you just need to experiment, please do it in your branch or your computer.

For now, my comments on this are:

- why did you delete audio renderer files? If I did the same you did, you wouldn't even have found the portaudio code, which was parked there since years. Please let me decide what to keep and what to delete
- since you changed a multimedia part, have you tested this change on Linux, Windows, macOS, QLC+ 5 and on the Raspberry Pi?
- please separate cmake from this. And why do we need cmake by the way?
- I posted here the code guidelines a couple of days ago. Have you read it? The 'Volume' variable you added makes me think you didn't

Tuxist:
i will do it in extra branch with cmake
cmake is QT6 and have direct support for visualstudio also i thing it's better to have seperate builddir

Tuxist:
i will fix volume support

PR #1215, opened 2020-02-27
RGB Matrix and Audio Trigger fixed to respond to submaster

ambraspace:
I noticed that RGB Matrix and Audio Trigger didn't respond to submaster value. While Audio Trigger didn't consider VCWidget's intensity() at all, RGB Matrix had very strange behavior. The slider would not get the position of mouse click, but some scaled down position. I figured, that the culprit was slotFunctionAttributeChanged() method firing back to the slider with scaled down value. So I deleted the whole method. This fixed odd behavior of the slider when submaster value was less then maximum.
After that, I had to override VCWidget's adjustIntensity() method to properly take into account submaster value.
This fixes the problem I have with one installation where customer wants blackout function for some of the fixtures.


ambraspace:
You can download QXW file and generated sound file (for Audio Trigger test) [here](https://nextcloud.ambraspace.com/index.php/s/wgTNFBTcdrqDWAE).

mcallegari:
Looks good to me, thanks! 👍

PR #1220, opened 2020-03-30
Fix wrong index in m_audioDevCombo. 

sbenejam:
Assuming that Default device has index = 0 in m_audioDevCombo, discovered devices are added now starting on index = 1.


mcallegari:
A genuine bug. Thanks!

mcallegari:
@sbenejam 
On a second thought, I think selIdx = 1 is not right though.
If only the default device is present (which is unlikely...but still), selIdx should be 0.
What do you think?

sbenejam:
Hi Massimo, I'm not an expert programmer as you are. But I think in the
case that only Default device is present selIdx should be 0.


Missatge de Massimo Callegari <notifications@github.com> del dia dt., 31 de
març 2020 a les 10:04:

PR #1222, opened 2020-05-06
Maintenance presets

kluszon:
Added presets for capability in maintenance and gobo groups in some lamps.

We have improved our review process and our git management. This is a small pull request to check if the improvements have been effective.


mcallegari:
Looks OK to me. Thanks

PR #1223, opened 2020-06-10
Qt5 5.15.0 deprecation fixes

edogawa23:
Everything inside Qt5 version checks, so this shouldn't break backwards compatibility.
I hope I'm right assuming (from reading similar fixes I found on the net) that with the the new generator seeding isn't needed anymore in doc.cpp. Otherwise please instruct me regarding a proper way to handle that.

### DIFF
--- a/engine/src/script.cpp
+++ b/engine/src/script.cpp
@@ -143,11 +143,7 @@ bool Script::setData(const QString& str)
     if (m_data.isEmpty() == false)
     {
         int i = 1;
-#if QT_VERSION < QT_VERSION_CHECK( 5, 14, 0 )
-        QStringList lines = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), QString::KeepEmptyParts);
-#else
-        QStringList lines = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), Qt::KeepEmptyParts);
-#endif
+        QStringList lines = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"));
         foreach (QString line, lines)
         {
             bool ok = false;
@@ -191,11 +187,7 @@ QString Script::data() const
 
 QStringList Script::dataLines() const
 {
-#if QT_VERSION < QT_VERSION_CHECK( 5, 14, 0 )
-    QStringList result = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), QString::KeepEmptyParts);
-#else
-    QStringList result = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), Qt::KeepEmptyParts);
-#endif
+    QStringList result = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"));
     while (result.count() && result.last().isEmpty())
         result.takeLast();
 
--- a/engine/src/script.cpp
+++ b/engine/src/script.cpp
@@ -143,7 +143,7 @@ bool Script::setData(const QString& str)
     if (m_data.isEmpty() == false)
     {
         int i = 1;
-#if QT_VERSION <= QT_VERSION_CHECK( 5, 14, 0 )
+#if QT_VERSION < QT_VERSION_CHECK( 5, 14, 0 )
         QStringList lines = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), QString::KeepEmptyParts);
 #else
         QStringList lines = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), Qt::KeepEmptyParts);
@@ -191,7 +191,7 @@ QString Script::data() const
 
 QStringList Script::dataLines() const
 {
-#if QT_VERSION <= QT_VERSION_CHECK( 5, 14, 0 )
+#if QT_VERSION < QT_VERSION_CHECK( 5, 14, 0 )
     QStringList result = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), QString::KeepEmptyParts);
 #else
     QStringList result = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), Qt::KeepEmptyParts);
--- a/engine/src/script.cpp
+++ b/engine/src/script.cpp
@@ -25,7 +25,10 @@
 #endif
 #include <QDebug>
 #include <QUrl>
-
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 10, 0 )
+#include <QRandomGenerator>
+#endif
+ 
 #include "genericfader.h"
 #include "fadechannel.h"
 #include "mastertimer.h"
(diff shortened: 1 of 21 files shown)


engine/src/script.cpp
```diff
@@ -140,7 +143,11 @@ bool Script::setData(const QString& str)
     if (m_data.isEmpty() == false)
     {
         int i = 1;
+#if QT_VERSION < QT_VERSION_CHECK( 5, 14, 0 )
         QStringList lines = m_data.split(QRegExp("(\r\n|\n\r|\r|\n)"), QString::KeepEmptyParts);
```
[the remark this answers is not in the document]
edogawa23: So this last commit got pretty big, pretty boring monkey work (look up the matching Qt enum constants and use these instead of plain numbers (mostly 0). This was not possible in two occasions, as the corresponding enum  [Qt::Orientations](https://doc.qt.io/qt-5/qt.html#Orientation-enum) doesn't contain a value 0. I haven't found out how to inline the matching code from my commit here, so I name the files/line numbers:

ui/src/flowlayout.cpp:117
ui/src/monitor/monitorlayout.cpp:120

Both methods don't seem to be used anywhere in the code currently. So I set the return value to Qt::Vertical for now without understanding what the functions are intended to do - please correct me if this is wrong.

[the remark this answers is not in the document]
edogawa23: I'm sorry, there are still missing bits as I see now via travis and appveyor failing, there need to be more version checks for using Qt::endl
I will keep committing stuff until the tests succeed...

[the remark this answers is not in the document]
edogawa23: So travis still fails, with this:
FAIL!  : QLCInputProfile_Test::save() 'ip.saveXML(path) == false' returned FALSE. ()

   Loc: [qlcinputprofile_test.cpp(390)]

I've tried to alter this call locally but can't seem to get it right.
I have no clue which permissions to put there for the test to pass, and ORing more permission bits leads to ther more obscure errors here...

This is a bit beyond my understanding I have to admit.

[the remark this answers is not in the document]
edogawa23: Thanks, i should have been able to try that myself...

Looks good now.

ui/src/ctkrangeslider.cpp
```diff
@@ -769,7 +769,7 @@ void ctkRangeSlider::mouseReleaseEvent(QMouseEvent* mouseEvent)
   this->QSlider::mouseReleaseEvent(mouseEvent);
 
   setSliderDown(false);
-  d->m_SelectedHandles = 0;
+  d->m_SelectedHandles = ctkRangeSliderPrivate::NoHandle;
```
mcallegari, line 772: Where does this come from? I can't find it.

fixtureeditor/main.cpp
```diff
@@ -52,13 +52,23 @@ void printVersion()
 {
     QTextStream cout(stdout, QIODevice::WriteOnly);
 
+#if QT_VERSION >= QT_VERSION_CHECK( 5, 15, 0 )
+    cout << Qt::endl;
+    cout << App::longName() << " " << App::version() << Qt::endl;
```
mcallegari, line 57: I don't like this solution.
There has to be another way. Like "\n" at each line and a single flush at the end.

ui/src/ctkrangeslider.cpp
edogawa23: it's defined in this same file, line 42

fixtureeditor/main.cpp
edogawa23: are you kidding? I just replaced endl by Qt;;endl, what's so bad about that, the rest is your own code

ui/src/ctkrangeslider.cpp
mcallegari: Alright, was looking in the .h file

fixtureeditor/main.cpp
mcallegari: No, I'm not kidding. I don't like the code duplication just to change a constant.
If we go that way, every time we add an option, we need to remember to add it twice, which is ugly.
Right now I don't have the chance to look into this myself, but I will pursue a unified solution I proposed before.

edogawa23: ok i got you wrong, sorry
I see what you mean now, and agree it's ugly. I simply followed the suggestion from the error message:
`error: ‘QTextStream& QTextStreamFunctions::endl(QTextStream&)’ is deprecated: Use Qt::endl [-Werror=deprecated-declarations]`

I just tried with `<< '\n' << flush;` but that leads to the same situation, it wants me to use Qt::flush instead. which like Qt::endl (I think, can't find earlier references) has been introduced with 5.15. 

Also I don't really understand why in the beginning of my porting effort I didn't get these errors at all, there were only a handful regarding the random stuff etc. from the first couple of commits. And just when I thought I was done, all these enum constant errors and finally this stuff popped up... and why the other guy in the forum didn't have to do all that... ?

mcallegari: >  and why the other guy in the forum didn't have to do all that... ?

I think he builds on macOS with CLANG. Most likely it has different build flags enabled than GCC

edogawa23: Here's an idea to avoid duplication, not sure if you will lke this:

`#if QT_VERSION >= QT_VERSION_CHECK( 5, 15, 0 )
#define endl Qt::endl
#endif

// ... printVersion() and printUsage() definitions

#if QT_VERSION >= QT_VERSION_CHECK( 5, 15, 0 )
#undef endl
#endif
`

enbyted: Maybe something along these lines:
```
#if QT_VERSION >= QT_VERSION_CHECK( 5, 15, 0 )
using line_end = Qt::endl;
#else
using line_end = endl;
#endif
...
cout << "..." << line_end;
```
Or maybe just:
```
#if QT_VERSION >= QT_VERSION_CHECK( 5, 15, 0 )
using endl = Qt::endl;
#endif
```

we're using c++, so preprocesor is not necessary for that.
I'm not sure where does `endl` without namespace come from, so I'm not sure if it will work. Is there somewhere `using namespace` that disappears in newer version of Qt?

BTW. I didn't look through all the code here, just this one bit got my interest when I was reading through the emails.

edogawa23: endl (and flush) without namespace comes from QTextStreamFunctions according to the error I quoted above, and that has been deprecated in Qt 5.15...

mcallegari: Alright, I checked this.
@edogawa23  Please proceed with
```
#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
#define endl Qt::endl
#endif
```
Apparently undef is not required.
Please also remove spaces from QT_VERSION_CHECK usage, like above, so it looks like the existing code.
Thanks

edogawa23:
Yes, meanwhile I also realized the #undef is unnecessary.
I hope all is fine now for a merge.

mcallegari:
Merged. Thanks a lot for your patience

PR #1224, opened 2020-06-16
Maintenance presets

kluszon:
Added presets for capability in maintenance and gobo groups in next parts of lamps.


mcallegari:
Sorry but I won't review a pull request of 329 commits and 158 files changed.
I already did this mistake and lots of errors went upstream and I had to fix them.
Plus, I see now some more errors were in (SmallToBig instead of ShutterOpen)

Please do these things:
- squash all these commits into 1 commit per pull request
- send a pull request with maximum 30 files changes each
- send a separate pull request for mistakes (SmallToBig instead of ShutterOpen)

Thanks

kluszon:
Ok, it'll be doing as you want.

PR #1231, opened 2020-07-09
Add APIv2 support to RGBscript devtool.html

hjtappe:
I updated the RGBscripts devtools to support the API v2 parameters and support the script developer in different ways:
- Add display of parameters from API v2
- Use local storage to let parameters survive a reload after a script update
- Update event handling to update matrix updates when changing e.g. speed
- Start timer on reload to make debugging more efficient
- Update visual appearance to let matrix better visible
- make HTML file more conforming to HTML standards to suppress systematic error messages and improve debugging of scripts

May this be useful to the project.
Hans-Jürgen

### DIFF
--- a/resources/rgbscripts/devtool.html
+++ b/resources/rgbscripts/devtool.html
@@ -1,136 +1,164 @@
-<HTML>
-<HEAD>
-<TITLE>Q Light Controller Plus - RGBScript Development Tool</TITLE>
+<!DOCTYPE html>
+<html>
+<head>
+<title>Q Light Controller Plus - RGBScript Development Tool</title>
+<style>
+ .readonly {
+  background-color: lightgrey;
+ }
+ body {
+  background: #373737;
+  color: #f9f8f0;
+ }
+</style>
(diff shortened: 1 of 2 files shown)


mcallegari:
Hi, thanks for this.
Please see my changes.
I don't like so much the forms layout, but it's a developer tool....so....
JS code check is a bit bizarre and it's not possible to fix all the issues.
If you're OK with all, I can merge this

hjtappe:
Thanks for your prompt response. Give me a few more days to look into the forms layout. So far I updated only what seemed a stronger necessity to make things distribute one the same screen.
The JS checks seem useful ( do they appear automatically on GitHub or did I miss to activate them?) and I see you tried to fix some of them. During development I hoped Eclipse and the web developer tools in my browsers would be as strict but they weren't. I think I can tackle at least some of them.

mcallegari:
I've made some more look&feel changes. How about this?

image

hjtappe:
I like the style very much. Thank you. Using mor margins and some indication of the information blocks would have also been my attemt to make it look better. :-)

PR #1232, opened 2020-07-09
RGB script with alternating lines and rectangles

hjtappe:
Currently, I had to make up multiple "evenodd" matrix scenes and combine them in a collection. For the simple use-case of alternating lights, I was looking to improve the efficient configuration of such an effect.
Here is an RGB script to make alternating lines and rectangles easily accessible.
I oriented myself on the plasma-style selection of colors. The script could instead also use the RGB color selector (or both standard color selectors) for the first color. Let me know if this would be better usable.

### DIFF
--- a/resources/rgbscripts/alternate.js
+++ b/resources/rgbscripts/alternate.js
@@ -1,0 +1,272 @@
+/*
+  Q Light Controller Plus
+  plasmacolors.js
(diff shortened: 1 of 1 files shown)


mcallegari:
Looks nice. Thanks

PR #1234, opened 2020-07-23
devtool.js coding corrections

hjtappe:
Improvements and some error corrections:
- Clear the table of properties when loading a new algorithm before the APIv2 check
- Use DOM objects rather than writing the innerHTML directly.
- Define variables in their block scope


hjtappe:
This status now finally fiyes as many Codacy as I could fix in this code. The remaining messages are due to the fact that the script must handle variables and functions of unknown name.

mcallegari:
OK, thanks.

PR #1237, opened 2020-08-04
Added new gobo preset and ActsOn in mode's channel

kluszon:
1. Added new preset GoboRotatingMacro for group capability. 
2. Added "ActsOn" in mode's channel for setting the impact of between channels.

Modifications discussed on [forum](https://www.qlcplus.org/forum/viewtopic.php?f=35&t=14353)


kluszon:
`QLCFixtureMode *mode = (QLCFixtureMode*) item->data(MODE_COL_NAME, PROP_PTR).toULongLong();`

The same casting is originally in line 1018. Should I fix it?

kluszon:
@mcallegari Can you review it?

mcallegari:
How can a channel acts on multiple channels? Example please.
Plus I don't like that clicking any channel always open the "acts on" dialog.

I expected to find an extra column (this is correct) but with a combo box, to select a single channel.
At least, that is how I would do it

kluszon:
> How can a channel acts on multiple channels? Example, please.

Try it one more time, it's impossible in this implementation. I secured it.

> Plus I don't like that clicking any channel always open the "acts on" dialog. What if I simply want to select a channel to remove it?

From my point of view, it's simple. One-click is row selecting, double click is editing channel settings and in this case, it is "acts on" dialogue. If you want I can limit click area to "Acts on" column.

> I expected to find an extra column (this is correct) but with a combo box, to select a single channel.
> At least, that is how I would do it

I can do it in this way if you really like it, although, in my opinion, the current implementation isn't bad and it'll be clear for users because it's similar to other areas in the application.

mcallegari:
I think you didn't get/understand my comments:

1. You implemented "acts on" as a list of channels. This means you had in mind the case where a channel affects (acts on) multiple channels. To be honest I've never encounter that case and I don't think we need it. I would say a channels can act on another single channel

2. In the fixture mode editing dialog, a single click on Linux+KDE always open the acts on dialog. Maybe on Windows or macOS it doesn't, but since the reference platform is Linux, we need to limit this action to the "acts on" column only.

kluszon:
1. I understood you correctly.  

> This means you had in mind the case where a channel affects (acts on) multiple channels.

No, I hadn't :) I tried to do it in a similar way than in other places in the application (e.g. adding channel to device's mode). That is why I block the option to add more than one channel to acts on the column.

2. 

> Maybe on Windows or macOS it doesn't, but since the reference platform is Linux...

Good point! You are right, it can be different on other systems.

**Summary:**

Should I limit "acts on" click action to column only or I must change it to your vision with combo box?

mcallegari:
I prefer a combo box. It's straight forward and requires less clicks.
On the other hand, it is tricky to keep them updated while populating a channel mode.
Perhaps we can adopt the same strategy we implemented in the VC Speed Dial properties.
Values are displayed as simple labels, but when you double click on them, you get a combo box to select something.

Something around this:
https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vcspeeddialproperties.cpp#L68

kluszon:
@mcallegari It was fixed as you expected. Please review it now.

fixtureeditor/editmode.cpp
```diff
@@ -31,6 +31,7 @@
 #include <QDebug>
 #include <QSize>
 #include <QAction>
+#include <QtGlobal>
```
mcallegari, line 34: Please include the proper header(s)
When you include QtGobal, it means you have no clue of what is needed in this class

resources/icons/svg/svgicons.qrc
```diff
@@ -1,169 +1,169 @@
 <RCC>
```
mcallegari, line 1: Please remove this file from this PR.
It has nothing do to with the changes

mcallegari (changes requested):

fixtureeditor/editmode.cpp
kluszon: It was because I used QOverload but build on the server failed, so I must change it...

resources/icons/svg/svgicons.qrc
kluszon: The file reverted to the previous version.

mcallegari (approved):

engine/src/qlcfixturemode.cpp
```diff
@@ -32,14 +32,17 @@
 
 QLCFixtureMode::QLCFixtureMode(QLCFixtureDef* fixtureDef)
     : m_fixtureDef(fixtureDef)
+    , m_actsOnChannelsList({})
```
mcallegari, line 35: No need to put it here. It's inizialied empty anyway

engine/src/qlcfixturemode.cpp
```diff
@@ -365,7 +382,20 @@ bool QLCFixtureMode::loadXML(QXmlStreamReader &doc)
             /* Channel */
             Q_ASSERT(m_fixtureDef != NULL);
             str = doc.attributes().value(KXMLQLCFixtureModeChannelNumber).toString();
-            insertChannel(m_fixtureDef->channel(doc.readElementText()),
+
+            int actsOnChannelIndex = -1;
+
+            if (doc.attributes().hasAttribute(KXMLQLCFixtureModeChannelActsOn)){
```
mcallegari, line 388: remove parenthesis

engine/src/qlcfixturemode.cpp
```diff
@@ -389,6 +419,19 @@ bool QLCFixtureMode::loadXML(QXmlStreamReader &doc)
         }
     }
 
+    // Set acts on channels
+
+    foreach(ChannelActsOnData channelSctsOnData, listChannelsWithActsOnIndex) {
```
mcallegari, line 424: Please respect the indentation of the rest if the code.
Parenthesis on a new line.
Space after conditionals (if, for, while, etc)

engine/src/qlcfixturemode.cpp
```diff
@@ -410,11 +453,22 @@ bool QLCFixtureMode::saveXML(QXmlStreamWriter *doc)
 
     /* Channels */
     QVectorIterator <QLCChannel*> it(m_channels);
-    while (it.hasNext() == true)
+    while(it.hasNext() == true)
```
mcallegari, line 456: Please don't remove spaces.

engine/src/qlcfixturemode.cpp
```diff
@@ -410,11 +453,22 @@ bool QLCFixtureMode::saveXML(QXmlStreamWriter *doc)
...
     {
+        QLCChannel* channel = it.next();
+
         doc->writeStartElement(KXMLQLCFixtureModeChannel);
         doc->writeAttribute(KXMLQLCFixtureModeChannelNumber, QString::number(i++));
-        doc->writeCharacters(it.next()->name());
+
+        if(m_actsOnChannelsList.contains(channel))
```
mcallegari, line 463: Space after if.
No need for parenthesis for a single line if

engine/src/qlcfixturemode.cpp
```diff
@@ -410,11 +453,22 @@ bool QLCFixtureMode::saveXML(QXmlStreamWriter *doc)
...
+        {
+            QLCChannel *ChannelActsOn = m_actsOnChannelsList.value(channel);
+            if(ChannelActsOn != nullptr){
```
mcallegari, line 466: Use NULL. It's backward compatible with older GCCs

engine/src/qlcfixturemode.h
```diff
@@ -206,9 +207,31 @@ class QLCFixtureMode
...
     quint32 masterIntensityChannel() const;
 
+    /*!
+     * \brief The ChannelActsOnData struct
+     *
+     * Contains channel pointer and acts on channel index.
+     *
+     */
+
+    struct ChannelActsOnData{
```
mcallegari, line 217: Parenthesis on new line

engine/src/qlcfixturemode.h
```diff
@@ -206,9 +207,31 @@ class QLCFixtureMode
...
+        QLCChannel *channel;
+        int actsOnIndex;
+
+        ChannelActsOnData(QLCChannel *newChannel, int newAcsOnIndex) :
```
mcallegari, line 221: I'm not sure this will work with older GCCs.
Might need to move it in the cpp file

fixtureeditor/addchannelsdialog.cpp
```diff
@@ -20,6 +20,7 @@
 #include "qlcchannel.h"
 #include "addchannelsdialog.h"
 #include "ui_addchannelsdialog.h"
+#include <QDebug>
```
mcallegari, line 23: No longer needed?

fixtureeditor/addchannelsdialog.h
```diff
@@ -21,6 +21,7 @@
 #define ADDCHANNELSDIALOG_H
 
 #include <QDialog>
+#include <QComboBox>
```
mcallegari, line 24: Not needed? Or move it?

fixtureeditor/editmode.cpp
```diff
@@ -233,13 +246,53 @@ void EditMode::refreshChannelList()
...
         item->setText(COL_NUM, str.asprintf("%.3d", (i + 1)));
         item->setText(COL_NAME, ch->name());
         item->setIcon(COL_NAME, ch->getIcon());
         item->setData(COL_NAME, PROP_PTR, (qulonglong) ch);
+
+        QStringList comboList;
+
+        comboList << "-";
+
+        for(auto index = 0; index < m_mode->channels().size(); index++)
```
mcallegari, line 265: Please don't use auto. It's C++11.
Modern syntax is used only in the qmlui

fixtureeditor/editmode.cpp
```diff
@@ -233,13 +246,53 @@ void EditMode::refreshChannelList()
...
+        {
+            QLCChannel *currentChannel = m_mode->channels().at(index);
+            comboList << QString::number(index) + " - " + currentChannel->name();
+        }
+
+        QComboBox *comboBox = new QComboBox(this);
+        comboBox->addItems(comboList);
+
+        if(actsOnChannelIndex >= 0){
```
mcallegari, line 274: Remove parenthesis

fixtureeditor/editmode.cpp
```diff
@@ -233,13 +246,53 @@ void EditMode::refreshChannelList()
...
+            comboBox->setCurrentIndex(actsOnChannelIndex + 1);
+        }else{
+            comboBox->setCurrentIndex(0);
+        }
+
+        m_channelList->setItemWidget(item, COL_ACTS_ON, comboBox);
+
+        connect(comboBox, static_cast<void(QComboBox::*)(int)>(&QComboBox::currentIndexChanged), comboBox,
```
mcallegari, line 282: Looks like C++11 too.
Please use backward compatible syntax

fixtureeditor/fixtureeditor.cpp
```diff
@@ -512,6 +512,21 @@ void QLCFixtureEditor::slotRemoveChannel()
 
         m_channelList->setCurrentItem(next);
         setModified();
+
+        // Remove channel from modes and set nullptr for acts on channels
+
+        for( int i = 0; i < m_modeList->topLevelItemCount(); ++i )
```
mcallegari, line 518: space after for and no space before/after parenthesis

mcallegari (changes requested):

engine/src/qlcfixturemode.cpp
kluszon: You've got 859 of nullptr in your code...

engine/src/qlcfixturemode.cpp
```diff
@@ -34,12 +34,14 @@ QLCFixtureMode::QLCFixtureMode(QLCFixtureDef* fixtureDef)
     : m_fixtureDef(fixtureDef)
     , m_masterIntensityChannel(QLCChannel::invalid())
     , m_useGlobalPhysical(true)
+
```
mcallegari, line 37: remove empty line

fixtureeditor/editmode.cpp
```diff
@@ -233,13 +246,41 @@ void EditMode::refreshChannelList()
...
+            comboList << QString::number(index) + " - " + currentChannel->name();
+        }
+
+        QComboBox *comboBox = new QComboBox(this);
+        comboBox->addItems(comboList);
+
+        if (actsOnChannelIndex >= 0)
+        {
+            comboBox->setCurrentIndex(actsOnChannelIndex + 1);
+        } else {
```
mcallegari, line 277: parenthesis on new line or remove them all (from if and else)

fixtureeditor/fixtureeditor.ui
```diff
@@ -36,7 +36,7 @@
    <item row="0" column="0">
     <widget class="QTabWidget" name="m_tab">
      <property name="currentIndex">
-      <number>0</number>
+      <number>2</number>
```
mcallegari, line 39: Remove this change.
Default tab is 0, which is "General"

ui/src/qlcui.qrc
```diff
@@ -1,5 +1,6 @@
 <RCC>
     <qresource prefix="/">
+        <file alias="acts-on.png">../../resources/icons/png/acts-on.png</file>
```
mcallegari, line 3: Remove this change and the icon file.
It's not used anywhere

mcallegari (changes requested):

mcallegari (approved):

PR #1238, opened 2020-09-09
Fix some issues in AppStream MetaInfo

jeromelebleu:
This tries to fix some issues reported by [AppStream Report for Debian](https://appstream.debian.org/sid/main/issues/qlcplus.html):
* **asv-url-not-secure**: use HTTPS in homepage URL of MetaInfo
* **missing-desktop-file**: define `<launchable>` tag in desktop-application MetaInfo to associate it with the .desktop file
* **metainfo-no-summary**: I guess that it would be fixed with the `<launchable>` tag since it is said that either .desktop file should have a `Comment=` field set or the component MetaInfo file should have a `summary` tag - which is the case.

### DIFF
--- a/platforms/linux/org.qlcplus.QLCPlus.appdata.xml
+++ b/platforms/linux/org.qlcplus.QLCPlus.appdata.xml
@@ -81,6 +81,7 @@
  </screenshots>
  <url type="homepage">https://www.qlcplus.org/</url>
  <url type="bugtracker">https://github.com/mcallegari/qlcplus/issues</url>
+ <launchable type="desktop-id">qlcplus.desktop</launchable>
  <provides>
   <binary>qlcplus</binary>
  </provides>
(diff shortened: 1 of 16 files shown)


platforms/linux/org.qlcplus.QLCPlus.appdata.xml
```diff
@@ -81,6 +81,7 @@
  </screenshots>
  <url type="homepage">http://www.qlcplus.org/</url>
```
mcallegari, line 82: https

platforms/linux/org.qlcplus.QLCPlusFixtureEditor.appdata.xml
```diff
@@ -43,6 +43,7 @@
  </screenshots>
  <url type="homepage">http://www.qlcplus.org/</url>
```
mcallegari, line 44: https

mcallegari (changes requested):

jeromelebleu:
@mcallegari Thanks, I missed those files!

mcallegari (approved):

PR #1239, opened 2020-09-11
Let RGBMatrix address RGB, W, A, UV channels.

hjtappe:
Additionally, fix inconsistency of stopcolor being used but neither editable nor reset in additive and substractive blend mode.

This commit enables RGB matrix users to use the full color space of RGB+X fixtures, such as implementing a white LED flashing on top of a set RGB color in the matrix.

This implements feature request https://www.qlcplus.org/forum/viewtopic.php?f=18&t=14207

### DIFF
--- a/engine/src/rgbmatrix.cpp
+++ b/engine/src/rgbmatrix.cpp
@@ -45,6 +45,12 @@
 #define KXMLQLCRGBMatrixPropertyName "Name"
 #define KXMLQLCRGBMatrixPropertyValue "Value"
 
+#define KXMLQLCRGBMatrixColorMode "ColorMode"
+#define KXMLQLCRGBMatrixColorModeRgb "RGB"
+#define KXMLQLCRGBMatrixColorModeAmber "Amber"
+#define KXMLQLCRGBMatrixColorModeWhite "White"
+#define KXMLQLCRGBMatrixColorModeUV "UV"
+
 /****************************************************************************
  * Initialization
  ****************************************************************************/
@@ -62,6 +68,7 @@ RGBMatrix::RGBMatrix(Doc* doc)
     , m_roundTime(new QElapsedTimer())
     , m_stepsCount(0)
     , m_stepBeatDuration(0)
+	, m_colorMode(RGBMatrix::ColorModeRgb)
 {
     setName(tr("New RGB Matrix"));
     setDuration(500);
@@ -392,6 +399,10 @@ bool RGBMatrix::loadXML(QXmlStreamReader &root)
         {
             setEndColor(QColor::fromRgb(QRgb(root.readElementText().toUInt())));
         }
+        else if (root.name() == KXMLQLCRGBMatrixColorMode)
+        {
+            setColorMode(stringToColorMode(root.readElementText()));
+        }
         else if (root.name() == KXMLQLCRGBMatrixProperty)
         {
             QString name = root.attributes().value(KXMLQLCRGBMatrixPropertyName).toString();
@@ -447,6 +458,9 @@ bool RGBMatrix::saveXML(QXmlStreamWriter *doc)
     {
         doc->writeTextElement(KXMLQLCRGBMatrixEndColor, QString::number(endColor().rgb()));
     }
+
+    /* Color Mode */
+    doc->writeTextElement(KXMLQLCRGBMatrixColorMode, RGBMatrix::colorModeToString(m_colorMode));
 
     /* Fixture Group */
     doc->writeTextElement(KXMLQLCRGBMatrixFixtureGroup, QString::number(fixtureGroup()));
@@ -700,6 +714,9 @@ void RGBMatrix::updateMapChannels(const RGBMap& map, const FixtureGroup *grp, QL
 
         QVector <quint32> rgb = head.rgbChannels();
         QVector <quint32> cmy = head.cmyChannels();
+        QVector <quint32> white = head.whiteChannels();
+        QVector <quint32> amber = head.amberChannels();
+        QVector <quint32> uv = head.uvChannels();
 
         quint32 masterDim = fxi->masterIntensityChannel();
         quint32 headDim = head.channelNumber(QLCChannel::Intensity, QLCChannel::MSB);
@@ -732,7 +749,7 @@ void RGBMatrix::updateMapChannels(const RGBMap& map, const FixtureGroup *grp, QL
 
         uint col = map[pt.y()][pt.x()];
 
-        if (rgb.size() == 3)
+        if (m_colorMode == ColorModeRgb && rgb.size() == 3)
         {
             // RGB color mixing
             FadeChannel *fc = getFader(universes, fxi->universe(), grpHead.fxi, rgb.at(0));
@@ -744,7 +761,7 @@ void RGBMatrix::updateMapChannels(const RGBMap& map, const FixtureGroup *grp, QL
             fc = getFader(universes, fxi->universe(), grpHead.fxi, rgb.at(2));
             updateFaderValues(fc, qBlue(col), fadeTime);
         }
-        else if (cmy.size() == 3)
+        else if (m_colorMode == ColorModeRgb && cmy.size() == 3)
         {
             // CMY color mixing
             QColor cmyCol(col);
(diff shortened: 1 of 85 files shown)


mcallegari:
Hi, thanks for this. I'll need some time to analyze and test it.

One thing: since this has been requested multiple times, in my mind I thought of "autodetecting" the white case when R == G == B.
In this way, the white channel is indeed considered and if you select shades of gray only the white channel is used.
it's just an "implicit" way of doing this, instead of creating a new mode.
But again, just give me some time to process it.

hjtappe:
Hi, take your time to process my proposal. :-)

For me, I decided against using the implicit mode for several reasons:
- When fading through a color scheme and going through a "grey" color, this would in many cases result in a flash as the new white-only grey would be generated by 1x3W instead of 3x3W Color LED.
- The exact color to match is in many cases not simply R=G=W but - depending on the quality of the white LED - is only somewhere near, so you would probably see the switch. Many white LEDs have more blue built into them than the usual blue single LED would cover (as far as I see).
- The same is for the Amber - depending on the LED, the amber value is not always the same amber "color", as far as I know there is no standard for it and the switch would not match exactly the right color., unless the fixtured define an exact color scheme, taking the color and intensity matches into account...
- It does not cover the UV case

All in all, I think, using the matrix for RGB effects plus adding the option to address another, non-RGB color seemed to make more sense to me and gives more clarity of what the user gets.

There is one more scenario to be considered, which is mixing RGB, RGBW and Halogen Dimmer channels in one matrix. In this case, I prefer to address the channels more distinctly (only the white channels and the halogen dimmer channels).

Thanks for reviewing my inputs.

mcallegari:
Makes sense.
Since this has been requested too: what about implementing a generic solution that allows to control any channel via RGB Matrix?
Cases could be dimmer channels, shutter, gobos(?), effects, etc..
I'm pretty sure more than one user could find it useful.
Then we would have the "happy problem" that the name "RGB Matrix" wouldn't fit much any longer...but that's another story...

hjtappe:
The dimmer channels are included in the "White" mode I think. For shutters, gobos and effect, I currently have no idea how this would map into the RGB matrix with scripts like plasma, given that the gobos and effects are not normalized and they do not fit sequences in the same way as we have with effects running through fixtures and start/stop colors (channels) moving through values in the sequence (which for RGB / grayscale makes sense, but would address an arbitrary gobo or effect).
The idea is interesting, e.g. to control alternating gobos or effects in a matrix of moving heads, but to be honest, I would see this either as a separate effects matrix (with other scripts in the background) or even better as a wizard for the script function interface (setfixture: ...).

engine/src/rgbmatrix.cpp
```diff
@@ -62,6 +68,7 @@ RGBMatrix::RGBMatrix(Doc* doc)
     , m_roundTime(new QElapsedTimer())
     , m_stepsCount(0)
     , m_stepBeatDuration(0)
+	, m_colorMode(RGBMatrix::ColorModeRgb)
```
mcallegari, line 71: spacing

engine/src/rgbmatrix.cpp
```diff
@@ -780,6 +810,13 @@ void RGBMatrix::updateMapChannels(const RGBMap& map, const FixtureGroup *grp, QL
     }
 }
 
+uchar RGBMatrix::rgbToGrey(uint col)
+{
+    // the weights are taken from
```
mcallegari, line 815: indentation

engine/src/rgbmatrix.cpp
```diff
@@ -819,6 +856,55 @@ void RGBMatrix::setBlendMode(Universe::BlendMode mode)
...
+
+RGBMatrix::ColorMode RGBMatrix::colorMode() const
+{
+	return m_colorMode;
+}
+
+void RGBMatrix::setColorMode(RGBMatrix::ColorMode type)
+{
+	m_colorMode = type;
+    emit changed(id());
```
mcallegari, line 871: indentation

engine/src/rgbmatrix.h
```diff
@@ -265,6 +269,31 @@ class RGBMatrix : public Function
...
+
+    /*************************************************************************
+     * RGB Color Mode
+     *************************************************************************/
+public:
+    /** Color Modes for the RGB Matrix */
+    enum ColorMode
+    {
+        ColorModeRgb = 0,
+		ColorModeAmber,
```
mcallegari, line 281: indentation

ui/src/rgbmatrixeditor.cpp
```diff
@@ -369,14 +374,15 @@ void RGBMatrixEditor::updateExtraOptions()
...
+            if (accColors == 1 || m_blendModeCombo->currentIndex() == Universe::MaskBlend)
             {
                 m_endColorButton->hide();
                 m_resetEndColorButton->hide();
             }
-            else
+            else // accColors > 1
             {
-                m_endColorButton->show();
+            	m_endColorButton->show();
```
mcallegari, line 385: indentation

ui/src/rgbmatrixeditor.cpp
```diff
@@ -385,6 +391,73 @@ void RGBMatrixEditor::updateExtraOptions()
...
 
+void RGBMatrixEditor::updateColors()
+{
+    if (m_matrix->algorithm() != NULL)
+    {
+        int accColors = m_matrix->algorithm()->acceptColors();
+        if (accColors > 0)
+        {
+        	if (m_matrix->blendMode() == Universe::MaskBlend)
+			{
```
mcallegari, line 402: indentation

ui/src/rgbmatrixeditor.cpp
```diff
@@ -635,23 +708,23 @@ void RGBMatrixEditor::slotBlendModeChanged(int index)
...
         m_startColorButton->setEnabled(false);
-        slotRestartTest();
     }
     else
     {
         m_startColorButton->setEnabled(true);
     }
     updateExtraOptions();
+    updateColors();
+	slotRestartTest();
```
mcallegari, line 719: indentation

ui/src/rgbmatrixeditor.cpp
```diff
@@ -635,23 +708,23 @@ void RGBMatrixEditor::slotBlendModeChanged(int index)
...
+}
+
+void RGBMatrixEditor::slotColorModeChanged(int index)
+{
+	RGBMatrix::ColorMode mode = RGBMatrix::ColorMode(index);
+	m_matrix->setColorMode(mode);
+    updateColors();
```
mcallegari, line 726: indentation

mcallegari (changes requested):
Overall it looks OK. I haven't tested it yet but I get the logic.
Please fix indentation and remove the translation files from this PR.
I haven't realized you implemented the change in QLC+ 5 too!
Cool :)

hjtappe:
Hi,
> Please fix indentation and remove the translation files from this PR.

Oh, sorry, yes, indentation and spacing fixed. I had hoped I would "see" that when viewing the files, now I have updated my Eclipse config, so it should no more creep in. Wouldn't there be a codacy check to ensure BSD coding style and 4x spacing per indentation, that could help me? :-)

> I haven't realized you implemented the change in QLC+ 5 too!

Well, yes - but when I did that, I did not yet understand that qmlui would be version 5.x and would not be compiled, so it is a bit of blind programming. I hope it works. :-) See my patches in the forum, so I will be able to build qmlui under Linux.

hjtappe:
I left the German translation files in this pull request because I added the translations for the new code (and very few more missing ones). I hope that is OK. Please contact me if it should be separate.

engine/src/rgbmatrix.cpp
```diff
@@ -700,6 +714,9 @@ void RGBMatrix::updateMapChannels(const RGBMap& map, const FixtureGroup *grp, QL
 
         QVector <quint32> rgb = head.rgbChannels();
         QVector <quint32> cmy = head.cmyChannels();
+        QVector <quint32> white = head.whiteChannels();
```
mcallegari, line 717: Would it be better to perform these chaching only for their specific color mode?
UpdateMapChannels is performed at each RGBMatrix step: with a lot of fixtures or lot of steps this might be CPU expensive.

resources/docs/html_en_EN/rgbmatrixeditor.html
```diff
@@ -146,6 +146,16 @@ <H2 id="controls">Controls</H2>
     blending.
   </TD>
  </TR>
+ <TR>
+  <TD>
+   <B>Pattern Color Mode</B>
+  </TD>
+  <TD>The color mode is the mechanism select which color channels the RGB Matrices applies to.<br>
+    Default setting is RGB mode, addressing RGB channels or YMK channels. If set otherwise, only the
```
mcallegari, line 154: YMK -> CMY channels

ui/src/rgbmatrixeditor.cpp
```diff
@@ -385,6 +391,73 @@ void RGBMatrixEditor::updateExtraOptions()
...
+                    }
+                    else
+                    {
+                        pm.fill(QColor(gray, gray, gray));
+                    }
+                    m_endColorButton->setIcon(QIcon(pm));
+                }
+                m_previewHandler->calculateColorDelta(m_matrix->startColor(), m_matrix->endColor());
+            }
+            else {
```
mcallegari, line 438: Parenthesis on new line please

mcallegari (changes requested):
Please let's keep translation files on a separate PR
Otherwise I could update all the translation files after I merge this PR and then each translator can contribute. Thanks

mcallegari (approved):

mcallegari:
Ok, I have tested this. It works fine but here's some more comments:
- please remove the translation files and I'll merge the PR
- we definitely need a "Dimmer mode" and get rid of the old "dimmer control" parameter. I am thinking of how useful it could be on discharge lamps, where there's not RGB control but only color wheels
- color mode in QML UI doesn't work :) I'll fix it.

engine/src/rgbmatrix.cpp
hjtappe: Yes. I will refactor the method to perform only the necessary calculations. Also, the sanity check for the matrix size (leadting to "continue") will be moved some lines earlier.

resources/docs/html_en_EN/rgbmatrixeditor.html
hjtappe: Thanks for fixing.

hjtappe:
- Remaining translation files have been removed. Let me know if I can help with the German part.
- For the dimmer mode, I am undecided. For now, I would keep it for compatibility of old configurations. If there will never be the scenario to have a mix of e.g. RGB mode and dimmer control, I can assist in modifying the code or adding just the new option, so users can select to use it as a mode. I currently have no fixture type in mind, where I could test this behavior.
- Please let me know if I should implement the re-phrase towards "Control mode" (rather than "Color Mode").
- Sorry for my failed try to implement the QMLUI part. :-) Maybe, you can review the changes to get the qmlui Linux build working, I can then assist to work towards version 5.x.

PR #1246, opened 2020-11-21
Add extraFeedback option for MIDI input channels

mhthies:
This patch introduces a new option for input channels in MIDI input profiles. The option is meant to workaround the behaviour of some MIDI controllers, such as the *Behringer X Touch Compact*:

This controller (and probably others?) updates the state of its button LEDs internally, when the button is pressed or released. In the default configuration (buttons send high value when pressed and low value when rleased), the LED is turned off internally, when the button is released. QLC+ only sends feedback signals, when the state of the assigned virtual button changes, which is (for toggle buttons) only upon a high value from the input. Thus, when using such a button with a button widget in QLC+ in toggle mode, the controller's button reflects the virtual button state correctly while being pressed, but is reset incorrectly when being released. In simpler words: The LED stays off after toggling a virtual button on.

A workaround has been proposed in [this thread in the forum](http://www.qlcplus.org/forum/viewtopic.php?t=10943#p56060): The controller's buttons can be configured to work in "toggle" mode, so they send a high value when pressed and released once and a low value when pressed and released a second time. QLC+ supports this behaviour when using the "extra press" input channel option. However, with this configuration change, the buttons cannot be used for flash buttons anymore, since QLC+ does not receive any event on release of the controller's buttons.

So, this PR introduces the new option "Send extra feedback on release", which allows to update the controller's LED state when releasing a button.

Please let me know if I missed something or if the patch does not comply with the coding guidelines.

In particular, I'm unsure if the new `m_extraFeedbackOnRelease` attribute of the `VCWidget` class is the correct way to store this information for making it available to `VCButton::slotInputValueChanged`.

### DIFF
--- a/engine/src/qlcinputchannel.cpp
+++ b/engine/src/qlcinputchannel.cpp
@@ -35,6 +35,7 @@ QLCInputChannel::QLCInputChannel()
     , m_movementType(Absolute)
     , m_movementSensitivity(20)
     , m_sendExtraPress(false)
+    , m_extraFeedback(false)
     , m_lower(0)
     , m_upper(UCHAR_MAX)
 {
--- a/engine/src/qlcinputchannel.cpp
+++ b/engine/src/qlcinputchannel.cpp
@@ -48,6 +48,7 @@ QLCInputChannel *QLCInputChannel::createCopy()
     copy->setMovementType(this->movementType());
     copy->setMovementSensitivity(this->movementSensitivity());
     copy->setSendExtraPress(this->sendExtraPress());
+    copy->setExtraFeedback(this->extraFeedback());
     copy->setRange(this->lowerValue(), this->upperValue());
 
     return copy;
@@ -208,6 +209,16 @@ bool QLCInputChannel::sendExtraPress() const
     return m_sendExtraPress;
 }
 
+void QLCInputChannel::setExtraFeedback(bool enable)
+{
+    m_extraFeedback = enable;
+}
+
+bool QLCInputChannel::extraFeedback() const
+{
+    return m_extraFeedback;
+}
+
 void QLCInputChannel::setRange(uchar lower, uchar upper)
 {
     m_lower = lower;
@@ -250,6 +261,11 @@ bool QLCInputChannel::loadXML(QXmlStreamReader &root)
         {
             root.readElementText();
             setSendExtraPress(true);
+        }
+        else if (root.name() == KXMLQLCInputChannelExtraFeedback)
+        {
+            root.readElementText();
+            setExtraFeedback(true);
         }
         else if (root.name() == KXMLQLCInputChannelMovement)
         {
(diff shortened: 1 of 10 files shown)


mcallegari:
Hi, thanks for this patch.
Can you please explain a bit better how this device work?
Something like:
Button press - send MIDI value 127 - LED turn on
Button release - send MIDI value 0 - LED turn off

Is that how it works?
Thanks

mhthies:
Yes, correct. In a bit more detail:

Each button on the device can be configured in one of two modes: "Momentary" and "Toggle".


In "Momentary" mode:

* Button press - sends MIDI Note-on with velocity=127, LED on
* Button release - sends MIDI Note-off with velocity=0, LED off

In "Toggle" mode:

* Button press - sends MIDI Note-on with velocity=127, LED on
* Button release - nothing
* Button press - sends MIDI Note-off with velocity=0, LED off
* Button release - nothing


Additionally, in any mode, at any time, the LED is turned on when a MIDI Note-on is received and turned off when a MIDI Note-off (or Note-on with velocity=0) is received.

Unfortunately, changing the mode requires using a Windows-only proprietary software to write a different configuration to the device. Thus, I'd like to keep all butons in "Momentary" mode and do the toggle/flash configuration in QLC+.

mcallegari:
> Unfortunately, changing the mode requires using a Windows-only proprietary software to write a different configuration to the device. Thus, I'd like to keep all butons in "Momentary" mode and do the toggle/flash configuration in QLC+.

Just a comment on this one. If you manage to "sniff" what this software does on the MIDI line, if we're lucky, it sends SYSEX data, which can be then saved into a QLC+ MIDI template.

I'd say toggle mode is the one used by most controllers. I need to double check what Scene flash does for that

mhthies:
> If you manage to "sniff" what this software does on the MIDI line, if we're lucky, it sends SYSEX data, which can be then saved into a QLC+ MIDI template.

Thanks for the hint. I will try to figure out what this configuration software tool does.

> I'd say toggle mode is the one used by most controllers. I need to double check what Scene flash does for that

As far as I understand the code and the options, "Momentary" mode is the one expected by QLC+ with default settings – except for the LED behaviour. This is because, a toggle button in QLC+ only toggles on 'Note on' MIDI messages and ignorse 'Note off' messages. To support the "Toggle" button mode, the "extra press" in the input channel options must be checked, which causes generation of a Note-on-Note-off sequence for each incoming Note on/off message.

siegmund42:
This is basically addressing the same problem (I had on BCF 2000) I solved via #927 
The proposed solution here avoids an extra (and sometimes unnecessary) sending of feedbacks.

In any case, would love to see one of these PRs merged - would help a lot!

mhthies:
@siegmund42: Oh, sorry, I should have looked at the old PRs before digging into this.

And I'm actually not sure if saving the unnecessary feedback updates on other MIDI controllers is worth all the the complexity of the new config option.

mcallegari:
@siegmund42 @mhthies if you're both on the same page with @siegmund42 change, I'd prefer merging his change as less impacting than another new config option (which might create confusion to users)
I also have a BCF2000 that I can test for VC flashing buttons.
My only concern is if it "regress" on some other devices I couldn't think of right now (OSC? loopback?)

mcallegari:
I have merged #927 then

siegmund42:
Awesome, thanks!

PR #1254, opened 2021-01-06
Account for channel value modifed with VC widget page value when using MIDI feedback for OMNI channel.

mmmorris1975:
VC frame pages encode their value in the channel value sent to the plugin, when using the MIDI OMNI
channel, this gets embedded in the bits which define the MIDI channel to send a feedback message on.
Take off the embedded page data to get the correct MIDI channel for sending feedback.

Add some simple test cases to validate the change.

[Forum discussion](https://www.qlcplus.org/forum/viewtopic.php?f=33&t=14709)

### DIFF
--- a/plugins/midi/src/common/midiprotocol.cpp
+++ b/plugins/midi/src/common/midiprotocol.cpp
@@ -127,8 +127,9 @@ bool QLCMIDIProtocol::feedbackToMidi(quint32 channel, uchar value,
                                      uchar* data1, uchar* data2)
 {
     // for OMNI mode, retrieve the original MIDI channel where data was sent
+    // account for cases where channel has been modified with VC widget page info
     if (midiChannel == MAX_MIDI_CHANNELS)
-        midiChannel = channel >> 12;
+        midiChannel = (channel >> 12) % MAX_MIDI_CHANNELS;
 
     // Remove the 4 MSB bits to retrieve the QLC+ channel to be processed
     channel = channel & 0x0FFF;
--- a/plugins/midi/test/midi_test.cpp
+++ b/plugins/midi/test/midi_test.cpp
@@ -57,7 +57,7 @@ void Midi_Test::feedbackToMidi()
 
     QLCMIDIProtocol::feedbackToMidi(channel, value, midiChannel, false, &cmd, &data1, &data2);
 
-    QCOMPARE(cmd, MIDI_PROGRAM_CHANGE | midiChannel);
+    QCOMPARE(cmd, uchar(MIDI_PROGRAM_CHANGE | midiChannel));
     QCOMPARE(data1, uchar(2U));
 }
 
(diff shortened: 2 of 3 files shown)


mcallegari:
Hi, thanks for this, but I need to test it myself. I'm scared of regressions around the corner.
I was used to use a tool named KMidimon in Linux, but it's not possible to build it these days.
How did you test this? Can you share a workspace and suggest a tool? Would save me a lot of time.
Thanks

sbenejam:
Hi Massimo, you can use Protokol to insect Midi and OSC traffic. There are
versions for Linux, Mac, Windows, Android, ...
https://hexler.net/products/protokol

Missatge de Massimo Callegari <notifications@github.com> del dia dc., 6 de
gen. 2021 a les 14:11:

mcallegari:
Thanks @sbenejam ! I'll give it a try

mmmorris1975:
Hi Massimo!  I forgot to upload the workspace I used to test the changes, but it gave me an opportunity to think of a few other things to add to it.  I have uploaded them to the forum, since I can't attach them to this thread.  Hopefully it covers a few more cases now.  The midi-frame-test workspace should work in any version on any platform, since it uses non-paged frames to hold the buttons sending the feedback.  The midi-page-test workspace has those same buttons spread out across multiple pages.  I also created those same sets of buttons around pages which are a multiple of 16 to make sure the behavior holds around the boundary of the calculation.  So pages 1-6 hold buttons, then pages 16-21 have the same set of widgets (but different channel numbers due to the encoding of the page info in the value), and finally pages 32-37 have those buttons again for one extra test.  All pages in between are empty.  The buttons are labeled to correspond to the MIDI data it should send (Ch1 PC1 for MIDI channel 1 Program Change 1, Ch3 CC3 for MIDI channel 3 Controller 3).  For the control change messages, I just let it send the button default values when it is toggled.

Regarding tools, I've been using [midipipe](https://midipipe.en.softonic.com/mac) on Mac, though I've seen it start acting strange if too many "invalid" messages are sent to it.  I'll have to check on that Protokol tool @sbenejam mentioned.  On the raspberrypi I'm using the `gmidimonitor` package, and on my Linux Mint VM I'm using the `midisnoop` package.  It looks like midisnoop package is available in the RPi and Linux Mint repos.

mcallegari:
Sorry for the delay, but I'm not able to reproduce this.
I'm using a APC mini on OMNI channel and everything seems to work correctly across frame pages without your patch.
Is there a MIDI trace or QLC+ debug messages that you can share to show the problem?
Thanks

mmmorris1975:
Thought a video demo might be helpful.  There's no audio, but there's probably little I could add with voice that the clicking through the QLC+ app, and seeing the output in the midisnoop tool demonstrates.  This was on my Linux VM, but if you want, I can do the same directly from the Raspberry Pi, or a Mac, with the same QLC+ version and show file.

https://youtu.be/qRmIr_JI7Cs

mmmorris1975:
I have a new laptop running Linux (Pop OS 20.10 -- a Ubuntu derivative), and with QLC+ 4.12.3 DEB package from the website, I can still reproduce this error.  After applying this patch, and recompiling using the `create-deb.sh` script, and installing the new package over 4.12.3, the expected MIDI messages are emitted.

FYI, I rebased to the HEAD of qlcplus:master to make sure everything still works. (like a fool, I also pushed that, so now there's a whole bunch of extra stuff in the PR).

mcallegari:
Ouch...I was about to merge this PR, but now it doesn't look ok (100 files modified)
Can you please fix it and go back to the tiny change that was there previously?
Thanks

mmmorris1975:
This should be fixed.  A bit of `git reflog` and `git reset` work was able to get PR back in its original state.

mcallegari:
Great, thanks :+1:

PR #1258, opened 2021-01-10
Display snow flakes falling or air bubbles rising.

hjtappe:
This RGB script displays either pixels falling or rising in different Z-levels and intensity. The effect can simulate falling snow or bubbles (e.g. in combination with the ocean effect)
it supports flakes in random colors and can be configured to apply a random windchill which moves the pixel flakes / bubbles to simulate wind gusts (or water waves respectively).
I hope this may this be useful in themes around winter and water.


mcallegari:
Thanks! :+1:

PR #1260, opened 2021-01-10
Simulate fireworks on an RGB Matrix

hjtappe:
Firework is starting near the center of the panel and explodes in a ball effect, flying with speed relative to the distance from the center slightly towards the side.
With a flickering (random noise) effect, the firework exists through the top of the panel.
It supports random colors for creating new firework.
I hope this may this be useful, especially at the end of shows in parallel to a physical firework.


sandinak:
So .. ran this on a new high end i7 with lots of memory .. and for a single matrix I have 64x8 .. and setting it to 100ms it gets pretty abusive to the CPU using the devtool .. however running in in the app it's quite a bit more efficient. Nice add! 

LGTM +1

mcallegari:
I like most of it, but not quite the part when exploded items are carried away from the matrix.
It's a start but maybe there's a little bit more work to do to make this awesome?

Maybe there's something here to get inspired?
http://www.jsfx.com/submitscripts/fireworks/
https://slicker.me/javascript/fireworks.htm
https://codepen.io/whqet/pen/Auzch
https://jsfiddle.net/dtrooper/AceJJ/

hjtappe:
I see what you are looking for. Give me some time to go into that direction.

hjtappe:
As far as I could see, the examples are based on the HTML canvas to extrapolate the moving objecs. I will try to implement a similar visual appearance in the RGB map later this year.

hjtappe:
Hello Massimo, this is now the script with the improved explosion of the rocket into particles.
The examples you showed me are based on a HTML5 canvas, which requires not only the JS engine but also a HTML5 engine. The particles and the dimming of the path in the current rgbScript framework need to be calculated pixelwise in JS loops rather than using the optimized HTML5 canvas capabilities.
It could be a future extension to QLCplus to calculate the more performance-expensive scripts in such a framework (HTML canvas in 2D or even 3D) to open the world of more complex scripts.

mcallegari:
This is so nice! Thank you very much!

PR #1261, opened 2021-01-10
Display Xmas trees and Xmas stars on an RGB matrix

hjtappe:
These scripts display a configurable number of trees or stars on an RGB matrix, moving around and bouncing on edges (and themselves) similar to the balls effect.
I hope this may this be useful, especially on performances around trees, stars or christmas.


sandinak:
I like the color collisions.  Nice work. LGTM +1

mcallegari:
Can you please unify these two and call it "Xmas Shapes" with a "Shape" parameter? 
Perhaps you can create a set of sprites? For now flake and tree, in future a snowman? :wink:

hjtappe:
The script can now be extended with arbitrary objects through the matrix. I implemented the mentioned snowmans (two-balls-flavour) , improved the color shading on the Xmas trees, derived a bell from the trees with another calculation and added UFOs as more flying objects. So was the file renamed. Please review and see if it can be useful.

On the way, I extended the devtools to also save the color settings and the matrix size setting to improve the reloads when the page is reloaded.

mcallegari:
Very nice!
Last comments:
- please rename the file to lowercase like the others: "flyingobjects.js"
- basically the "balls" script(s) are no longer needed?
- I would improve a bit the bells: they're too tall 
- improve also stars with 5 tips: https://previews.123rf.com/images/verdateo/verdateo1207/verdateo120700021/14301846-gold-christmas-star-isolated-on-white.jpg
- what about snowflakes? https://www.macmillandictionary.com/external/slideshow/thumb/emoji_snowflake_thumb.jpg

hjtappe:
Hi,
here we go. When doing this mathmatically correct, i.e. with fading out boundaries, the calculation time increases.
While implementing the requested changes and the snowflake and a better star, I came along many other basic objects, which I promptly added. Let's see if it will be usable.
From my point of view, I would keep the standalone ball.js for backward compatibility.

mcallegari:
@hjtappe WOW!
This is absolutely brilliant! I love it!
Thank you so much this whole lot of effort

mcallegari:
@hjtappe hi, it seems recent Qt versions rejects this script for 2 reasons:
- doesn't like the "let" statement. Replacing with "var" seems to fix it.
- doesn't like "=>" at line 496. Is there a replacement for that?

Thanks

kripton:
Hi Massimo, since I know some JavaScript, I would fix this right now. Unfortunately, I don't have the time to properly test it. The replacement for the so-called "Arrow functions" can be seen here, for example: https://www.w3schools.com/js/js_arrow_function.asp

It's basically an "anonymous"/inline function with shorter syntax. Hope it helps :)

kripton:
And both things Qt complains about (let and the arrow functions) are pretty new. If they worked in previous versions of Qt, they may default to an older standard in the new versions. But I would be surprised if there was no parameter for the JS interpreter to define the standard being used. See also here: https://bugreports.qt.io/browse/QTBUG-47735

hjtappe:
Interesting. At least for the "let", I thought, it was had been introduced already  between 2013 and 2016 (https://262.ecma-international.org/6.0/#sec-let-and-const-declarations and the link you posted).
Is it a QT regression or are these semantics deprecated now?
@kripton, thanks for fixing it. If you don't find the time, ping me and I will give it a try.

mcallegari:
Indeed, the reason for this "regression" is unclear to me too.
What I know is that QLC+ 4 still uses the QtScript lib to parse JS, which is marked as deprecated since many 5.x versions.
QLC+ 5 instead, uses the "official" Qt way, which is QtDeclarative. More modern, more efficient, more everything.
It's a pity to make a step back on this RGB script, but to be honest I don't see a so much big deal here if we use an "old fashion" syntax. Agree?

hjtappe:
Sure, for this single script, it should be no problem to use the pre-2015 semantics. Nevertheless, the ECMAscript-215 semantics had been chosen due to the many functions using the same variable names (for good reason) and the thought that only "let" will explicitly limit their scope (while "var" might cause an implicit shadowing). BTW: The tests on travis did not fail so far with the QT-5 / qmlui build (or did they and we did not see? or does it not test with the "right" QMLUI-build engine). We had activated them with the Linux-build patch a month ago.

I am thinking a bit broader: There might be more locally modified scripts in the wild, so we should well understand the problem and at least add a release note telling our users what to modify and why. It does not seem logical (but only pragmatic) to QLC+-locally deprecate the previously-working ECMAscript semantics which worked before, was there since 2015 and still is not deprecated in 2021 (https://262.ecma-international.org/#sec-let-and-const-declarations).

So, in order to reproduce the problem locally on my side (considering the need for a release note for the locally modified scripts of our users), would you mind to please describe how I can reproduce the error, so I can do a bit of research and we can understand the underlying problem better?


As a side-note, building with Ubuntu-20.04-packaged QT also does not work because the maintainers decided to deactivate the QT build for an "unsafe" QT function which is normally packaged. It is only back in Ubuntu-20.10. I suppose, there might be a similar problem down under this ECMAscript problem (like a compile option mission removed as a regression within QT) and it might be worth digging down into this to understand the root cause.

FloEdelmann:
>  was there since 2015 and still is not deprecated in 2021

In fact, using `var` instead of `let` and `const` actually considered bad practice in web development nowadays.

* https://scotch.io/courses/10-need-to-know-javascript-concepts/declaring-javascript-variables-var-let-and-const#toc-which-to-use
* https://medium.com/swlh/the-difference-of-var-vs-let-vs-const-in-javascript-abe37e214d66#a2d0
* https://hacks.mozilla.org/2015/07/es6-in-depth-let-and-const/

hjtappe:
Hello Massimo,

I created an account with QT and tried to find the the effect with different versions. QT fixed a JS caching issue with 5.15.1. This is the version at which also the UI slightly changes.

Nevertheless, I am sorry, I cannot reproduce the problem in Ubuntu 20.10 in the latest 5.x QT version, which today is 5.15.2:
`make clean; make distclean; QTDIR=~/Qt/5.15.2/gcc_64/ ./create-appimage.sh`

image

I will try to extend the test to really step through the scripts and find error messages therein. So far, I cannot see error messages and the script and especially the snowflake function is showing up well.

Can you please check whether there is a problem (like some leftover QJSengine cache) in your development environment which causes this strange regression? Or are you already preparing to use QT 6.x, which is now the actively maintained version?

hjtappe:
I nevertheless noted that I have to toggle the visibility of the "speed" row in order for the "parameters" row to show the script parameters. I did not yet find the cause of this UI effect. Maybe, you have an idea on that.

PR #1264, opened 2021-01-31
Rpms

kiilerix:
Some fixes for building RPMs on Fedora.

Especially the last 2 commits seem like bad workarounds for something that really should be fixed elsewhere. And it would be nice to have a more elegant way to specify udev location.

### DIFF
--- a/platforms/linux/qlcplus.spec
+++ b/platforms/linux/qlcplus.spec
@@ -41,6 +41,9 @@ sed -ie '/UDEVRULESDIR/s|/etc/udev/rules.d|/usr/lib/udev/rules.d|' variables.pri
 #############################################################################
 
 %build
+# qmake-qt5 will only include existing files in install_translations - create the .qm files first
+./translate.sh ui
+
 qmake-qt5
 make
 
--- a/platforms/linux/qlcplus.spec
+++ b/platforms/linux/qlcplus.spec
@@ -95,6 +95,7 @@ INSTALL_ROOT=$RPM_BUILD_ROOT make install
 %_libdir/qt5/plugins/qlcplus/libe131.so
 %_libdir/qt5/plugins/qlcplus/libspi.so
 %_libdir/qt5/plugins/qlcplus/libloopback.so
+%_mandir/*/*
 %doc /usr/share/qlcplus/documents/*
 /usr/lib/udev/rules.d/z65-dmxusb.rules
 /usr/lib/udev/rules.d/z65-anyma-udmx.rules
--- a/platforms/linux/qlcplus.spec
+++ b/platforms/linux/qlcplus.spec
@@ -64,9 +64,9 @@ INSTALL_ROOT=$RPM_BUILD_ROOT make install
 %files
 %defattr(-,root,root)
 %{_bindir}/*
-%{_libdir}/libqlcplusengine.so.*
-%{_libdir}/libqlcplusui.so.*
-%{_libdir}/libqlcpluswebaccess.so.*
+%{_libdir}/libqlcplusengine.so*
+%{_libdir}/libqlcplusui.so*
+%{_libdir}/libqlcpluswebaccess.so*
 %{_datadir}/qlcplus/translations/*
 %{_datadir}/applications/*
 %{_datadir}/pixmaps/*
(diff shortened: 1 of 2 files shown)


platforms/linux/qlcplus.spec
```diff
@@ -41,6 +41,10 @@ of other lighting control commercial softwares.
 %build
 sed -ie '/UDEVRULESDIR/s|/etc/udev/rules.d|/usr/lib/udev/rules.d|' variables.pri
 
+# qmake-qt5 will only include existing files in install_translations - create the .qm files before *really* running qmake-qt5
+qmake-qt5
+make translate
```
mcallegari, line 47: I think you can replace these 2 lines with
./translate.sh

platforms/linux/qlcplus.spec
```diff
@@ -39,6 +39,7 @@ of other lighting control commercial softwares.
 #############################################################################
 
 %build
+rm .qmake.stash
```
mcallegari, line 42: To be honest I never had issues with qmake.stash
What do you mean with "Keeping it prevents building on other platforms"?
When you build QLC+ you are targeting a single platform and you are supposedly starting from a clean checkout of the repository, which doesn't include qmake.stash.
I would revert this commit.

platforms/linux/qlcplus.spec
kiilerix: Right.
But even more, shouldn't it be able to generate the full Makefile directly after a clean checkout?

platforms/linux/qlcplus.spec
kiilerix: I understand that the current main case for create-rpm.sh is to build locally for the current platform. But currently, it still works by creating a "clean" tar and build the rpm "from scratch" in the lightweight rpmbuild "sandbox".
I think it will save a lot of trouble to do that cleanly.
For example, I tried to verify that the same source rpm could build in a real sandbox on "all" hardware platforms. It failed as https://koji.fedoraproject.org/koji/taskinfo?taskID=60937238 when the source rpm contained an x86_64 .qmake.stash . build.log shows for example
```
g++ -pipe -Werror -Wno-unused-local-typedefs -g -fPIC -Wall -Wextra -dM -E -o moc_predefs.h /../lib/qt5/mkspecs/features/data/dummy.cpp
/../lib/qt5/bin/moc -DQT_GUI_LIB -DQT_CORE_LIB --include /builddir/build/BUILD/qlcplus-4.12.4/hotplugmonitor/src/moc_predefs.h -I/../lib/qt5/mkspecs/linux-g++ -I/builddir/build/BUILD/qlcplus-4.12.4/hotplugmonitor/src -I/usr/include/qt5 -I/usr/include/qt5/QtGui -I/usr/include/qt5/QtCore -I/usr/include/c++/10 -I/usr/include/c++/10/x86_64-redhat-linux -I/usr/include/c++/10/backward -I/usr/lib/gcc/x86_64-redhat-linux/10/include -I/usr/local/include -I/usr/include hpmprivate-udev.h -o moc_hpmprivate-udev.cpp
usr/include/c++/10/bits/stl_relops.:67: Parse error at "std"
make[2]: *** [Makefile:451: moc_hpmprivate-udev.cpp] Error 1
```

platforms/linux/qlcplus.spec
kiilerix: I read the qtmake documentation and googled around. I now understand that there is no way to do this cleanly with qmake.

With qmake not being able to control the whole build process and quitly depending on build artifacts at configuration time, it becomes even more important to consistently do the steps in scripts and documentation. And to have an initial distclean that cleans everything up so qtmake will do exactly the same every time.

platforms/linux/qlcplus.spec
kiilerix: I haven't been able to reproduce or understand how this could happen. I will leave out this change.

mcallegari:
Looks good to me, thanks.
Perhaps, just give the .spec file a check against the one I use for automated builds on OBS:
https://build.opensuse.org/package/view_file/home:mcallegari79/qlcplus-qt5/qlcplus-qt5.spec?expand=1
There might be more details to include

PR #1265, opened 2021-02-01
Make detached contexts apply the Qt style sheet

chille:
When detaching a context the Qt style sheet is not applied to the new window. This is because DetachedContext does not call QMainWindow::setStyleSheet() nor set any parent where the style sheet could be inherited from.

Disclaimer: This patch is only tested on Linux (Ubuntu 20.04.2 LTS). I have no idea if setting the main window as parent might have any other unforeseen consequences. Everything seems to work as usual.


mcallegari:
Nice little patch, thanks a lot!

PR #1267, opened 2021-02-02
Rpms 2nd round

kiilerix:
More RPM cleanups.
Only tested on Fedora 33, but should be safe on all versions and on Suse.


mcallegari:
Looks good, thanks :+1:

PR #1269, opened 2021-02-05
Rpms take 3

kiilerix:
AFAICS, this should make these RPMs pretty much as good as the ones on OBS. Many of the cleanup steps could be applied there too. I think it would be a good idea to use this one in both places.

Are there any differences you are concerned about?


mcallegari:
What's the deal with qmlui files?
To answer your question: the OBS .spec file handle builds also for openSuse so it has a few conditionals specific to OBS

kiilerix:
For some reason, the qmlui simplecrypt files were committed with X bit. I think that popped up as a rpmlint error message or something. That seems unintentional, and like a trivial thing to fix.

I don't see anything in the obs spec that seems specific to obs. I only see conditionals to handle difference between fedora and suse. I reviewed carefully and think everything is handled with these spec changes.

kiilerix:
Any thoughts on how we can make progress on this one?

There might be more work to do before it can be fully aligned with the OBS packaging, but can we agree that these steps make sense? You ar e very welcome to graft the changes you like so we can focus on reworking the rest.

mcallegari:
Sorry, lately I've been busy with other personal life matters.
I'll get back on this as soon as I can

mcallegari:
```
%if %{defined fedora}
%if %{defined suse_version}
```
Are these actually defined in Fedora and Suse? I thought they were OBS specific variables.

kiilerix:
Yes. That howto is (in that regard) summmarizing how distros are different.

```
$ rpm -E '%{defined suse_version}'
0
$ rpm -E '%{defined fedora}'
1
$ rpm -E '%{fedora}'
33

$ mock -r opensuse-leap-15.2-x86_64 --chroot 'rpm -E "%{defined suse_version}"' 2> /dev/null
1
$ mock -r opensuse-leap-15.2-x86_64 --chroot 'rpm -E "%{suse_version}"' 2> /dev/null
1500
$ mock -r opensuse-leap-15.2-x86_64 --chroot 'rpm -E "%{defined fedora}"' 2> /dev/null
0

```

mcallegari:
Alright, then this is good to be merged. Thanks

kiilerix:
Thanks.
It would be interesting to see if this spec builds on OBS and whether they could be consolidated.

PR #1273, opened 2021-03-21
deal with new flush API for buffers in recent libftdi1 1.5

edogawa23:
https://www.qlcplus.org/forum/viewtopic.php?f=12&t=14804

Updated the patch to act conditionally, depending on which libftdi version is found at compile time.


mcallegari:
If this is for libftdi 1.5 only then using LIBFTDI1 is wrong. It involves all 1.x versions

edogawa23:
You're right, I've added a version check and new variable to src.pro

mcallegari:
It looks too complicated to me. Moreover, does it work on macOS?
I would simply use the `ftdi_get_library_version` in the original `LibFTDIInterface::purgeBuffers` method and decide to invoke `ftdi_tcioflush` or `ftdi_usb_purge_buffers`.
That's all. Simple and clean.

edogawa23:
Yeah I just tried and ran into problems, I can't use the deprecated function name from C++.

Not sure how to check macOS behaviour, the only Macbook I have access to is the machine used for editing and running shows at the theater, and not set up for compiling stuff - I rather wouldn't mess with it at this time.

mcallegari:
I think you dragged in unrelated MIDI code in this PR. Please rebase/fix.

Also, please instead of purge/flush as separate methods, make a single change in LibFTDIInterface::purgeBuffers and ifdef the FTDI versions. This way it looks cleaner than multiple ifdefs around.
Thanks

edogawa23:
Sorry for the MIDI stuff, i had merged your master branch - still unsure about best practices in GH...

I thought the method name should reflect what it's doing, but if you prefer this simpler way, I'm totally fine with that :)

Hopefully it's ready now for merging?

mcallegari:
Yeah, thanks. I need to check if this is OK on macOS and then it can be merged

mcallegari:
@edogawa23 I finally checked on macOS and this seems OK 👍 
One last change: can you please rename "NEW_LIBFTDI1" to "LIBFTDI1_5"? I doubt "NEW" will still make sense in 5-10 years ;)
Also, can you simplify the version check like the following? 
Does it really make sense to check all the non-verified conditions and do nothing first?
```
versionAtLeast(ftdi1, 1.5) {
    message("Found libftdi1 version $$ftdi1 (new buffer flush API)")
    DEFINES += LIBFTDI1_5
}
```

mcallegari:
Thanks! :+1:

PR #1274, opened 2021-05-13
Updated qmlui catalan and spanish translations

sbenejam:
Updated qmlui catalan and spanish translations


mcallegari:
Thanks a lot! :+1:

PR #1275, opened 2021-05-22
Add parentheses to `print` in Python script (Python 3 compatibility)

FloEdelmann:
It's now possible to call the script with both `python3` and `python2` (although the latter is no longer maintained).


mcallegari:
Thanks 👍

PR #1276, opened 2021-06-07
fix handling of two deprecated Qt5 enum value names

edogawa23:
Once again I'm getting warnings (treated as errors) breaking compilation for Tumbleweed in OBS, so I've removed the QComboBox::AdjustToMinimumContentsLength property as it seems unnecessary (the QComboBox above it doesn't have this either), and instead added a tooltip even if the meaning of that choice should be totally clear.

The second change of 2 instances from now deprecated Qt::MidButton to Qt::MiddleButton, which seems to be around already since Qt 4.6.0.

### DIFF
--- a/ui/src/addfixture.ui
+++ b/ui/src/addfixture.ui
@@ -89,11 +89,8 @@
       </item>
       <item row="3" column="2">
        <widget class="QComboBox" name="m_universeCombo">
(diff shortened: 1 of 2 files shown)


mcallegari:
Thanks 👍 
Let's see if this is enough also to fix the Arch build on OBS

PR #1278, opened 2021-06-11
integrate peperoni Rodin1 mk3 

Tolmino:
New Version of Rodin1 MK3 is not found.


mcallegari:
Hi, thanks for this. Have you tested the change with the device? Does it work as expected?

Tolmino:
yes, it works.

after commit i have had 2 mail with Jan Menzel and he confirmed that i have done the right thing

PR #1279, opened 2021-06-13
Update build to support building QMLUI / QLC+v5.x for Linux

hjtappe:
Hello Massimo,
I adapted the build of create-appimage.sh and create-deb.sh to be able to build QMLUI / QLC+v5.x for Ubuntu. On the way, I had to fix some tests and some code which was indicated by the tests to fulfill changed test assumptions related to scriptv4.cpp.
Please check the changes (I left one FIXME in the create-appimage.sh for the "chrpath" line of code which does returns an error on Ubuntu, but is left and set to ignore the error).
The unittest.bat adaption is a blind try to reflect the required change in the script call to accept the variant to be built. Please review this
Please see if I can help further to prepare the Linux side of things.

### DIFF
--- a/create-appimage.sh
+++ b/create-appimage.sh
@@ -17,7 +17,12 @@ else
     qmake CONFIG+=appimage CONFIG+=qmlui
 fi
 
-make -j8
+NUM_CPUS=`nproc`
+if [ -z "$NUM_CPUS" ]; then
+    NUM_CPUS=8
+fi
+
+make -j$NUM_CPUS
 make check
 
 mkdir $TARGET_DIR
--- a/create-appimage.sh
+++ b/create-appimage.sh
@@ -5,6 +5,9 @@
 # Export QTDIR before running this, like:
 #   export QTDIR=/home/user/Qt5.9.4/5.9.4/gcc_64
 
+# Exit on error
+set -e
+
 TARGET_DIR=~/qlcplus.AppDir
 
 # Build
@@ -15,6 +18,7 @@ else
 fi
 
 make -j8
+make check
 
 mkdir $TARGET_DIR
 make INSTALL_ROOT=$TARGET_DIR install
@@ -24,27 +28,32 @@ strip $TARGET_DIR/usr/lib/libqlcplusengine.so.1.0.0
 
 chrpath -r "../lib" $TARGET_DIR/usr/bin/qlcplus-qml
 
-cd $TARGET_DIR/usr/bin
+pushd $TARGET_DIR/usr/bin
 find . -name plugins.qmltypes -type f -delete
 find . -name *.qmlc -type f -delete
 rm -rf QtQuick/Extras QtQuick/Particles.2 QtQuick/XmlListModel
 rm -rf QtQuick/Controls.2/designer QtQuick/Controls.2/Material
 rm -rf QtQuick/Controls.2/Universal QtQuick/Controls.2/Fusion
 rm -rf QtQuick/Controls.2/Imagine QtQuick/Controls.2/Scene2D
-cd -
+popd
 
-wget -c https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-x86_64 -O $TARGET_DIR/AppRun
-chmod a+x $TARGET_DIR/AppRun
+if [ ! -f $TARGET_DIR/AppRun]; then
+    wget -c https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-x86_64 -O $TARGET_DIR/AppRun
+    chmod a+x $TARGET_DIR/AppRun
+fi
 
 cp resources/icons/svg/qlcplus.svg $TARGET_DIR
 cp platforms/linux/qlcplus.desktop $TARGET_DIR
 sed -i -e 's/Exec=qlcplus --open %f/Exec=qlcplus-qml/g' $TARGET_DIR/qlcplus.desktop
 
+if [ ! -f /tmp/appimagetool-x86_64.AppImage ]; then
+    wget -c https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage -O /tmp/appimagetool-x86_64.AppImage
+    chmod a+x /tmp/appimagetool-x86_64.AppImage
+fi
 
-rm /tmp/appimagetool-x86_64.AppImage
-wget -c https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage -O /tmp/appimagetool-x86_64.AppImage
-chmod a+x /tmp/appimagetool-x86_64.AppImage
-
-cd $TARGET_DIR/..
+pushd $TARGET_DIR/..
 /tmp/appimagetool-x86_64.AppImage $TARGET_DIR
-cd -
+popd
+
+make clean
+
--- a/create-appimage.sh
+++ b/create-appimage.sh
@@ -17,7 +17,7 @@ else
     qmake CONFIG+=appimage CONFIG+=qmlui
 fi
 
-NUM_CPUS=`nproc`
+NUM_CPUS=`nproc` || true
 if [ -z "$NUM_CPUS" ]; then
     NUM_CPUS=8
 fi
@@ -25,13 +25,17 @@ fi
 make -j$NUM_CPUS
 make check
 
-mkdir $TARGET_DIR
+if [ ! -d "$TARGET_DIR" ]; then
+    mkdir $TARGET_DIR
+fi
 make INSTALL_ROOT=$TARGET_DIR install
 
 strip $TARGET_DIR/usr/bin/qlcplus-qml
-strip $TARGET_DIR/usr/lib/libqlcplusengine.so.1.0.0
+# see variables.pri, where to find the LIBSDIR
+find $TARGET_DIR/usr/lib/ -name libqlcplusengine.so.1.0.0 -exec strip -v {} \;
 
-chrpath -r "../lib" $TARGET_DIR/usr/bin/qlcplus-qml
+# FIXME: no rpath or runpath tag found.
+chrpath -r "../lib" $TARGET_DIR/usr/bin/qlcplus-qml || true
 
 pushd $TARGET_DIR/usr/bin
 find . -name plugins.qmltypes -type f -delete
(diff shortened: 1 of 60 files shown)


hjtappe:
Hello Massimo, can you please take a look into the travis result? I activated the unittest for qmlui and if I am looking for the message in the web, it seems like the different version of QML (or the travis container?) leads to the differences in the rgbtext test result which cannot open the X display. It could be that the travis environment lacks some QT library like the libqt5x11extras5 if that is missing (https://askubuntu.com/questions/308128/failed-to-load-platform-plugin-xcb-while-launching-qt5-app-on-linux-without).

How did you approach the UI tests when setting up the tests for version 4?

hjtappe:
I found the line making the difference. "It's easy if you look at the right place, but if you don't know the place, it's difficult." :-) Just enabling the xvfb service did the trick to provide the test with what it needed.
I think, this pull request is now good to be reviewed. Mind the one FIXME I mentioned in create-appimage.sh which is "useless" in Ubuntu Linux but might be required in a Mac environment (I 'm guessing)?

mcallegari:
Sorry I don't have access to a PC now. I will have a look on Wednesday or so

hjtappe:
No hurry, thanks for all your engagement.

create-appimage.sh
```diff
@@ -14,37 +17,51 @@ else
...
+    chmod a+x $TARGET_DIR/AppRun
+fi
 
-cp resources/icons/svg/qlcplus.svg $TARGET_DIR
-cp platforms/linux/qlcplus.desktop $TARGET_DIR
+cp -v resources/icons/svg/qlcplus.svg $TARGET_DIR
+cp -v platforms/linux/qlcplus.desktop $TARGET_DIR
 sed -i -e 's/Exec=qlcplus --open %f/Exec=qlcplus-qml/g' $TARGET_DIR/qlcplus.desktop
 
+if [ ! -f /tmp/appimagetool-x86_64.AppImage ]; then
```
mcallegari, line 58: I would remove this check.
There might be a new version of the tool available.

create-appimage.sh
```diff
@@ -14,37 +17,51 @@ else
...
 rm -rf QtQuick/Extras QtQuick/Particles.2 QtQuick/XmlListModel
 rm -rf QtQuick/Controls.2/designer QtQuick/Controls.2/Material
 rm -rf QtQuick/Controls.2/Universal QtQuick/Controls.2/Fusion
 rm -rf QtQuick/Controls.2/Imagine QtQuick/Controls.2/Scene2D
-cd -
+popd
 
-wget -c https://github.com/AppImage/AppImageKit/releases/download/continuous/AppRun-x86_64 -O $TARGET_DIR/AppRun
-chmod a+x $TARGET_DIR/AppRun
+if [ ! -f $TARGET_DIR/AppRun]; then
```
mcallegari, line 49: I would remove this check.
There might be a new version of the tool available.

mcallegari:
@hjtappe hi, the scripts part looks good to me. Thanks.
The cpp part needs an explanation though: is there a problem with the test units?
In any case, I would split scripts/cpp into two separate PRs. Does it make sense to you?
Thanks

hjtappe:
Thanks for your review. I will also remove the last FIXME line and just keep the command where it is as it does not spoil the build under Linux (but if you send me a remark what it does, I can add this as a comment, so it becomes more clear why this problem can be ignored but the chrpath needs to remain).
https://github.com/mcallegari/qlcplus/pull/1279/files#diff-57321cfd7179a48888d756bdbb4f5823d00e27dc9b8ab29311e7d7915544d56aR37

I can split the CPP fix from the script, but you will have to first merge the CPP code and then the scrips as the test is broken with the qmlui build. Please give me some time to work on that. :-)

In Script::functionList() const, the parsing fails because the ending parenthesis is not covered as a command boundary. The "startswith" also fixes a potential problem if the systemcommand program contains a keyword which would match already here.

One major difference to the ui build is that the scriptv4 does not store an extra line number in the array of used-by class names, but only stores the class name. Thus the doc.cpp in Doc::getUsage(quint32 fid) has to iterate stepwisein the qmlui build and 2-step-wise in the ui build. In the program, you will most probably be able to trigger a segfault at that place (or an exception from the iterator class) as it accesses an invalid offset in the list. Try to run the doc_test in a qmlui build and you will see that exception.

There could be another direction to fix ist - adding the line number next to the used-by classes in scriptv4.cpp, but that line was (intentionally?) missing there.
Please check that place in the code for advice on how to handle it if this was not intentional.
https://github.com/mcallegari/qlcplus/blob/c2e4405b7a41f9185cb5a9e6ea98a84a02c980a1/engine/src/script.cpp#L210
versus
https://github.com/mcallegari/qlcplus/blob/c2e4405b7a41f9185cb5a9e6ea98a84a02c980a1/engine/src/scriptv4.cpp#L177
leading to
https://github.com/mcallegari/qlcplus/pull/1279/files#diff-f57795248966ecdd153017d55ae4aa6c129a914b13a27a2b8ff0646dd5efd9dbL1157

create-appimage.sh
hjtappe: OK, check is now removed.

create-appimage.sh
hjtappe: OK, check is now removed.

unittest.sh
```diff
@@ -61,13 +68,17 @@ do
         continue
     fi
 
+    # Ignore script folder when in qmlui mode
+    if [ "${test}" == "engine/test/script" -a "$TARGET" == "qmlui" ]; then
```
mcallegari, line 72: If I understand correctly. this is a workaround because the current Script test unit doesn't work with ScriptV4 class?
If so, we should fix the test, not the sh script ;)

hjtappe: Good catch. Thanks for your review, I was chewing on getting the build and the tests for v5 running on Linux for quite a while (on a very low priority) and had forgotten about this deactivation.
Please see https://github.com/mcallegari/qlcplus/pull/1283 as a basis for the fix which (when integrated) I will merge here together with the fix of above lines.

mcallegari: This can be removed now and also in unittest.bat.
Also, this can be changed too:
https://github.com/mcallegari/qlcplus/blob/master/engine/test/test.pro#L42

hjtappe:
Hello Massimo,
the failure revealed by test was presumably something where I got lost with the test when I deactivated it long ago. It does not fail on every test execution and might be a race condition or a contract not tested sufficiently.

Can you please investigate the test and the scriptv4 module?
It seems like an initialization or check is missing like in https://github.com/mcallegari/qlcplus/blob/master/engine/src/rgbscript.cpp#L210 is missing (engine = new QScriptEngine(QCoreApplication::instance());) but I am lost with my little expertise to judge whether it should be in the test or is missing in the code already.

mcallegari:
I think we can try to replace this line 
https://github.com/mcallegari/qlcplus/blob/master/engine/test/script/script_test.cpp#L72
with
QTEST_MAIN(Script_Test)

hjtappe:
I tried - this makes the test fail more reliably with the QMLUI / v5 build. :-)
My wild guess is that it is due to the new scriptv4 design with the scriptrunner, which seems to require a thread to attach to, but somewhere does ensure that it can find it in the QCoreApplication instance. It smells like it should follow a similar concept as the rgbscript implementation and test, but I am not (yet) deep enough in the code and inner workings to understand the new implementation concept and find the spot.

hjtappe:
OK, it seems to fix the problem. I found more sources suggesting this fix, so I tried to see what travis tells us.
The failure I see locally might have to do with my local ccache or the like. I will try to clean my environment.
Nevertheless, the implementation in the rgbscript.cpp re-uses the QJSEngine, which might also be applied in the script runner. You may want to consider this at a later point in time.

mcallegari:
On the contrary, QTEST_APPLESS_MAIN works for me locally, while QTEST_MAIN doesn't :(
I'll try to trigger a few times this last build to see if Travis is consistent with the results...

hjtappe:
Hm, an unstable test is a bit of a nightmare for us developers. ;-)

As the test does not do a lot, should we disable it for the time being? As far as I debugged, the QCoreApplication is there when checked in the test code, but when the test terminates, the application seems to tear down the QCoreApplication before the QJSengine thread is terminating.

I tried to re-implement the Mutex approach from rgbscript.cpp in scriptrunner.cpp, but it did not change the behavior.
I tried to reduce the script to make sure, we are not falling into a wait() trap in the JS snippet but to no avail.
On my virtual machine, APPLESS_MAIN passes in 50% of runs, MAIN passes in only 5%.

I think, we are chasing an underlying problem in the thread architecture in or around scriptrunner.

hjtappe:
In the new thread architecture between svriptv4 and scriptrunner, if the scriptv4 is waiting for the thread to finish, the test completes because the running JSEngine first terminates before the test and thus the QCoreApplication thread is terminated. But please review this with your expert judgement - I cannot judge about potential side effects.

mcallegari:
Interesting!
However, I'd say wait() should go before exit(), so I'd put the wait call at the end of the stop function:
https://github.com/mcallegari/qlcplus/blob/master/engine/src/scriptrunner.cpp#L59
basically like many other threads do:
https://github.com/mcallegari/qlcplus/blob/master/plugins/dmxusb/src/enttecdmxusbopen.cpp#L160
Thanks for your efforts!

hjtappe:
Let me know if I can make any change (such as changing the wait() location or moving the part into another merge request). I fear I cannot provide assistance regarding the question about the optimal place for wait() due to lack of knowledge.
Still, the current location was my first guess because the implementation suggests with the exit() at that place that it assumes, the thread is terminated there (but it is not until we wait() for it).

unittest.bat
```diff
@@ -49,6 +49,7 @@ FOR /D %%G IN ("*") DO (
         exit /B %ERRORLEVEL%
     )
     cd ..
+    :CONTINUE
```
mcallegari, line 52: Is this still needed?

mcallegari:
This can be merged, thanks!
Care to help with #1284 ? I suspect it's a Qt regression, but I couldn't figure out quickly a way to solve it.

unittest.bat
hjtappe: You are right, this is now superfluous. I have removed it.

hjtappe:
I can try to help with the other issue #1284, but the only thing we can hope for is some analytical capability - as I have no personal experience with QT myself. So please don't rely on me. :-)

PR #1282, opened 2021-06-22
Fix script bracket parsing and add line number to usage list.

hjtappe:
Hi, this is part two of the CPP files patches to fix problems when running the unittests in a QMLUI environment under Linux.
This patch addresses the new scriptv4 "functionlist" handling which lacked the second information in the list, which is the line number. The doc.cpp failed with an index out of bound error, which is now leading to a debug message. The keywords and translation from legacy to v4 is now more explicit between the versions.


mcallegari:
Looks good to me, thanks :+1:

PR #1283, opened 2021-07-04
Choose an common test entry point for script.cpp and scriptv4.cpp

hjtappe:
Hello Massimo,

The original Script::executeCommand is not part of scriptv4, so the test needs to use another entry point in order to also work for scriptv4.

Nevertheless, please check the test as it does not seem to implement any useful outcome (it seeds a script with syntax errors, but does not even check the reaction on these), but at least now it compiles.


mcallegari:
I agree, this test as it is, is pretty useless.
So let's merge your patch for now.
I'll think of something more meaningful for this case later.
Thanks

PR #1286, opened 2021-07-16
Change the FlowLayout from vertical to horizontal.

### LINKED ISSUE #1284: [ui] XY Pad doesn't scale correctly
Screenshot attached.
Tried different Qt versions.

Screenshot_20210711_155432

hjtappe:
This fixes #1284 - [ui] XY Pad doesn't scale correctly.

Between 4052a1b (10.06.2020) and 8e24317 (13.06.2020) the value was set
from zero to Qt::Vertical, which caused the xy pad to display incorrectly.

The original return value was zero, the enumeration is:
    enum Orientation {
        Horizontal = 0x1,
        Vertical = 0x2
    };

### DIFF
--- a/ui/src/flowlayout.cpp
+++ b/ui/src/flowlayout.cpp
@@ -114,7 +114,7 @@ QLayoutItem *FlowLayout::takeAt(int index)
 //! [6]
 Qt::Orientations FlowLayout::expandingDirections() const
 {
-    return Qt::Vertical;
+    return Qt::Horizontal;
 }
 //! [6]
 


hjtappe:
@mcallegari, please review carefully. I climbed down through the months to find the start of the regression, then read the diff and analytically checked the assumptions if there could be a regression smell.
This fixes the display, but I have whatsoever no idea of possible side effects, Qt or what the FlowLayout is.
Please judge for yourself if this is what was needed to be done to fix it. :-)

mcallegari:
@hjtappe wow, nice catch!
I am actually puzzled since Qt::Horizontal nor Qt::Vertical are equal to zero (like to old code)
So now I'm wondering how could the original code possibly working. Probably Qt itself didn't recognize 0 as a valid value and fallback to Horizontal.
In any case, this seems to be the genuine fix.
Thanks a lot!

PR #1288, opened 2021-07-31
Add support for FTDI USB-COM485-PLUS2 for DMX USB Interface

digitall:
This has been tested and works fine using port #0 on the unit i.e.
port closest to the USB connector.


mcallegari:
Looks good, thanks! 👍

PR #1289, opened 2021-08-13
Do not load RGBscript files which are not script files

hjtappe:
While searching through the logs for the JS execution problem and errors, I would like to add the following lines to avoid the errors while loading files which are not meant to be rgbscripts:
Do not load RGBscript files which are not script files (Makefiles, backup files, directories, etc.)

### DIFF
--- a/engine/src/rgbscriptscache.cpp
+++ b/engine/src/rgbscriptscache.cpp
@@ -69,6 +69,11 @@ bool RGBScriptsCache::load(const QDir& dir)
 
     foreach (QString file, dir.entryList())
     {
+        if (!file.toLower().endsWith(".js") || file.endsWith("devtool.js"))
+        {
+            qDebug() << "    " << file << " skipped (special file or does not end on *.js)";
+            continue;
+        }
         if (!m_scriptsMap.contains(file))
         {
             RGBScript* script = new RGBScript(m_doc);
--- a/engine/src/rgbscriptscache.cpp
+++ b/engine/src/rgbscriptscache.cpp
@@ -69,7 +69,7 @@ bool RGBScriptsCache::load(const QDir& dir)
 
     foreach (QString file, dir.entryList())
     {
-        if (!file.toLower().endsWith(".js") || file.endsWith("devtool.js"))
+        if (!file.toLower().endsWith(".js"))
         {
             qDebug() << "    " << file << " skipped (special file or does not end on *.js)";
             continue;
--- a/resources/rgbscripts/lines.js
+++ b/resources/rgbscripts/lines.js
@@ -33,7 +33,7 @@ var testAlgo;
     algo.linesAmount = 3;
     algo.properties.push("name:linesAmount|type:range|display:Amount|values:1,200|write:setAmount|read:getAmount");
     algo.linesSize = 10;
-    algo.properties.push("name:linesSize|type:range|display:Size|values:1,32|write:setSize|read:getSize");
+    algo.properties.push("name:linesSize|type:range|display:Size|values:1,32|write:setLinesSize|read:getLinesSize");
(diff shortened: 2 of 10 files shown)


hjtappe:
There might be a means to do it more with QT functions, such as:
`    dir.setFilter(QDir::Files);
    dir.setNameFilters(QStringList() << QString("*.js"));`
but modifying `dir` in this context is not allowed (probably to avoid side effects). Let me know if the patch is good enough or requires refactoring. I think due to the special JS file "devtool.js", we need to filter at this point anyway and there is no performance issue as the cache is loaded only once during initialization, so the filter will apply only once.

engine/src/rgbscriptscache.cpp
```diff
@@ -69,6 +69,11 @@ bool RGBScriptsCache::load(const QDir& dir)
 
     foreach (QString file, dir.entryList())
     {
+        if (!file.toLower().endsWith(".js") || file.endsWith("devtool.js"))
```
mcallegari, line 72: As you can see here
https://github.com/mcallegari/qlcplus/blob/master/resources/rgbscripts/rgbscripts.pro
devtool.js is not installed, so I don't see a reason to check for it in the code.

hjtappe: Hello Massimo, for the application image, this is true. Nevertheless, people customizing their scripts in the user directory, will have a copy of the devtool.js around in the user script directory.
I will nevertheless remove the check as you requested. If you can follow my thoughts, I can add it again as it only runs once when the cache loads the scripts and thus should not pose a performance issue.

hjtappe:
Hello Massimo, on the search for the QJScript problem you mentioned, I was lacking a possibility to reproduce any error or warning message. That lead me to implement a test executing all the scripts, so we will have a reference to look at when we see a problem with the scripts. Also, new scripts will be more thoroughly tested (one of the new scripts had out-of-bound problems, which are now fixed).
Please review my pull request and advise if anything should be changed or extracted into a separate request.

resources/rgbscripts/lines.js
```diff
@@ -33,7 +33,7 @@ var testAlgo;
     algo.linesAmount = 3;
     algo.properties.push("name:linesAmount|type:range|display:Amount|values:1,200|write:setAmount|read:getAmount");
     algo.linesSize = 10;
-    algo.properties.push("name:linesSize|type:range|display:Size|values:1,32|write:setSize|read:getSize");
+    algo.properties.push("name:linesSize|type:range|display:Size|values:1,32|write:setLinesSize|read:getLinesSize");
     algo.linesVariablity = 0;
```
mcallegari, line 37: I think this is a typo and should be fixed everywhere in the script

hjtappe: In the quest to find and fix the bugs in this script, I applied best practice to name the getters by the object they get and to avoid any name clashes for setSize / setSize in any other shadowed namespace.
The variable name in the script was not changed from "linesSize", so the getters and setters are [gs]etLinesSize by intention.
Would you like to revert this change in the getter and setter name or do you refer to the other best practice to use singular for the variable names, i.e. rename linesSize to lineSize?

mcallegari: My comment is about "linesVariablity". Should be "linesVariability"

hjtappe: Ah, thanks for the clarification. Blindness on my side. The script was broken due to this typo, but I updated it by consistently promoting the typo spelling instead of the correct one. I'll fix this.

hjtappe:
I extended the checks for valid definition and use of parameters and was able to detect and remove some other flaws in the scripts.

mcallegari:
Thanks! 👍

PR #1290, opened 2021-09-28
Devtool: Separate localstorage namespace, add pixel size in grid settings

hjtappe:
Hello Massimo, I started to work again on the draft pull request for a fireworks script - to see if we can have something useful at the end of the year :-).

While at it, I saw a need to display a preview grid with smaller pixels - and potential namespace clashes between devtool and the script under development, so I modified it in the devtool:

- use devtool. namespace in devtool local storage, to avoid name clashes with parameters named width, height, size, speed, etc.
- Add a parameter to size the pixels in the preview grid and fit a larger preview grid on the screen.


mcallegari:
Thanks 👍

PR #1295, opened 2021-10-15
Prevent color tool being covered up in fullscreen

SJChannel:
Specify the parent of the QColorDialog when creating it for the
color tool.  This ensures that the dialog remains on top of the
main window.  Without this fix, the dialog can get covered up in
fullscreen mode under Ubuntu MATE and probably other Linux distributions.
Once the dialog is buried, recovery is impossible and the application
must be killed from a terminal window.


SJChannel:
The recipe for reproducing the bug on Ubuntu MATE can be found [here](https://www.qlcplus.org/forum/viewtopic.php?f=21&t=15279).

PR #1299, opened 2021-11-25
Add Support for Nodle R4S HID_DMX Interface

qfulmina:
The Nodle R4S is a ready built version of the already supported DMXControl Projects e.V. Nodle U1.

See https://www.dmxcontrol-projects.org/en/projects/nodle-r4s.html

It differs only in its Product ID, namely 0833 instead of 0830. Hence the addition of that ID to the respective files.

Compiled and tested using Ubuntu 18.04.


mcallegari:
Looks good, thank you very much 👍

PR #1300, opened 2021-11-28
Blinder fix

hjtappe:
Hello Massimo,
While using the blinder effect, I found that there must be logic inversion (line 81/82), leading to just showing a full panel pane without the bulb showing for some dimensions.
While reviewing the code for any more potential optimizations, I moved some code from the loop into the initialization which should help saving system resources.


mcallegari:
Thanks :+1: 
I'll shortly review also your other PR
Sorry for the delay

PR #1303, opened 2021-12-05
Remove SYSFS from udev rules

jeromelebleu:
SYSFS key has been removed a long time ago from udev (in v174), and those rules generate errors that we could get rid of.


mcallegari:
Agreed. Those are very legacy statements to support ancient distros.
Let's move on! 👍

PR #1306, opened 2021-12-26
Improve the tempo detection algorithm

awwright:
The Speed Dial widget features a "Tap" button that allows users to tap along with a beat, when choosing a duration.

Currently, this tap button only computes the time since the last tap. The computed value can be off by +-20ms. When tapping to music, this is enough to put the beat completely out-of-phase with the music in 10 seconds.

This patch records the most recent 20 taps, and computes the tempo that best passes through all of them. Within 10 taps it can compute a tempo that's correct +- 2ms. It's even robust against occasional missed taps.

The first tap uses the old (current) behavior, as does waiting more than 1.5 seconds in between a tap.

### DIFF
--- a/ui/src/speeddial.cpp
+++ b/ui/src/speeddial.cpp
@@ -551,9 +551,53 @@ void SpeedDial::slotTapClicked()
     }
     // Round the elapsed time to the nearest full 10th ms.
     m_value = m_tapTime->elapsed();
-    setSpinValues(m_value);
-
     m_tapTime->restart();
+
+    // If it's been a while since the last tap, reset the history and just use the time since the last tap
+    if (m_value > 1500)
+    {
+        // TODO: shift the tempo slightly upward or downward and re-phase the tempo
+        m_tapHistory.clear();
+        // TODO: If m_value is above the widget's maximum, then ignore this tap altogether
+        setSpinValues(m_value);
+        updateTapTimer();
+        emit tapped();
+        return;
+    }
+
+    // If multiple taps have been input recently,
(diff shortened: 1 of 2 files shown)


mcallegari:
Hi, before I review this, please indent your changes like the rest of the code. Thanks.

awwright:
@mcallegari Maybe you can help me out a bit, I'm not sure what you're referring to. The guidelines don't seem to be specific, but you seem to use 4-space indents, correct?

awwright:
I pushed an update that changed some of the variable names and whitespace around the `if` tokens, hopefully that helps.

Would you be interested in a PR that adds a linter for these things?

ui/src/speeddial.cpp
```diff
@@ -551,10 +551,51 @@ void SpeedDial::slotTapClicked()
...
+        setSpinValues(m_value);
+        return;
+    }
+
+    // If multiple taps have been input recently,
+    // find the tempo that best passes through all of them.
+    m_tapHistory.append(m_value);
+    // This algorithm stabilizes around a tempo very quickly,
+    // so keeping more than a few taps in the history merely complicates tempo changes.
+    while(m_tapHistory.count() > 16) m_tapHistory.removeFirst();
```
mcallegari, line 571: newline here please

mcallegari (changes requested):

mcallegari:
4-space tabs is correct.
No need for a linter right now.
Thanks

mcallegari (approved):

mcallegari:
Great, thanks :+1:

PR #1307, opened 2021-12-26
HID plugin: Worked around a bug in QFile on Windows

Haspamelodica:
HID device paths are device paths on Windows. Those aren't handled correctly by QFile: Calling fileName after setting a path with setFileName returns the path without the `\\?\` prefix, which causes CreateFileA to later fail with ERROR_FILE_NOT_FOUND, when HIDDMXDevice::init is called.


mcallegari:
Hi, nice! Does this fix FX5 not working on Windows? Thanks

Haspamelodica:
Maybe. I'm not sure because I don't own a "real" FX5 (or Digital Englightment) interface, but I'm currently working on reimplementing them on an Arduino: https://github.com/Haspamelodica/dmx-usb-atmega328p-hid.
At least for my own interface, this bugfix makes it work with QLC+.

PR #1317, opened 2022-02-07
change background color of efxpreview area from white to proper theme…

edogawa23:
… "Base" color

(previously, using a dark desktop theme in linux would result in white lines and markers on white background)


mcallegari:
Ah, this is good! Also on macOS there's components to fix with dark theme
Thanks 👍

PR #1318, opened 2022-02-21
ui: fix wrong feedback on vccuelist in stepmode

dandanx:
The vccuelist feedback (Side Fader) works fine in Crossfade-Mode but in Steps-Mode the feedback values goes crazy.


mcallegari:
Makes perfect sense, thank you 👍

PR #1324, opened 2022-04-10
Verticalfall RGBmatrix script fix for out-of-bounds program termination

hjtappe:
This patch implements the re-initialization of the script once the geometry changes to avoid qlcplus crashes induced by out-of-bound array access in the script.

The attached sample project simulates the behavior, without the fix, when changing the matrix from "line" to the larger "matrix", QLC+ terminates unexpectedly (in both v4 and v5)
[test.qxw.txt](https://github.com/mcallegari/qlcplus/files/8459749/test.qxw.txt)

I am not sure if we could either a) handle this more graceful with an exception in the QLC+ code (scriptrunner for v5?) or guard this by a test case which changes the geometry and performs a few steps.


mcallegari:
Thanks :+1:

PR #1325, opened 2022-04-10
Qmake / make targets: run and doxygen

hjtappe:
For development convenience, I would like to suggest adding the "run" and "doxygen" targets to the command line make targets.


mcallegari:
Thanks :+1:

PR #1326, opened 2022-04-14
avoid sending duplicate MIDI notes

edogawa23:
When using MIDI plugin in Note mode to output channel data from QLC, notes are sent twice.
This patch removes a superfluous call of snd_seq_event_output(m_alsa, &ev); that causes this misbehaviour.

### DIFF
--- a/plugins/midi/src/alsa/alsamidioutputdevice.cpp
+++ b/plugins/midi/src/alsa/alsamidioutputdevice.cpp
@@ -153,7 +153,6 @@ void AlsaMidiOutputDevice::writeUniverse(const QByteArray& universe)
                 snd_seq_ev_set_noteoff(&ev, midiChannel(), channel, scaled);
             else
                 snd_seq_ev_set_noteon(&ev, midiChannel(), channel, scaled);
-            snd_seq_event_output(m_alsa, &ev);
         }
         else if (mode() == ProgramChange)
         {
--- a/plugins/midi/src/common/midiprotocol.h
+++ b/plugins/midi/src/common/midiprotocol.h
@@ -91,7 +91,7 @@ namespace QLCMIDIProtocol
 #define MIDI2DMX(x) uchar((x == 127U) ? 255U : x << 1)
 
 /** Convert DMX value to MIDI value */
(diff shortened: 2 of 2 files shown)


edogawa23:
Also, the code to scale DMX to MIDI values doesn't seem to work correctly, I get weird out of range values:

when going to value 128 in simple desk, MIDI value becomes 192, and increasing further ends weirdly at 255 for DMX 255;
in CC mode, MIDI value wraps around to -64 at 128, and continue  being negative values for CC mode). 

I'm looking into this but haven't succeeded so far, and am in a hurry right now.
Hope I have explained well enough, I believe the MIDI2DMX macro is wrong somehow.

edogawa23:
Seems I've found the proper fix for the problem mentioned in my comment above

mcallegari:
Hi, this sounds strange to me.
I haven't checked yet, but are you sure the two calls were because of Note On/Note Off?
Cause if that's the case IIRC that was a wanted behaviour.
How did you test this? With a button? Slider?

edogawa23:
I came across this when setting up to control our lightboard from QLC+ via MIDI. it behaved weirdly, jumping 2 cues whenever I pressed the prepared scene button. I also tested with a slider and in simple desk. then I connected kmidimon to see what is going on, and saw two identical midi notes arriving. To cross-check, I set it to CC mode and it didn't happen there. Today I removed the line in code, recompiled and that duplicate note was gone, and CC mode still working. I am in theater now, and with my modified version of QLC+ the console responds correctly.

sbenejam:
I just tested in Ubuntu and QLC+ 4.12.5-GIT compiled this morning. In CC
mode works as expected with Note mode QLC+ sends Note_On or Note_Off twice.
See attached screen capture.

Missatge de Edgar Aichinger ***@***.***> del dia dj., 14
d’abr. 2022 a les 18:22:

edogawa23:
I'm confident that my change is correct - see line 174 where snd_seq_event_output() finally is called for all if/else branches before draining the buffer.

mcallegari:
Alright, this looks good to me.
Thanks :+1:

PR #1327, opened 2022-04-14
chore: add websocket reconnect when connection is closed

joaoopereira:
This PR implements an automatic reconnection when the websocket is closed.

When using the webaccess in iOS Safari everytime that the iPhone screen is locked the websocket connection is lost and to workaround that, we are forced to refresh the page.


mcallegari:
Quite helpful, thanks :+1:

PR #1328, opened 2022-04-14
Rgbscripts fill.js: Add reverse direction

hjtappe:
In order to use mirrored RGB matrices sides left and right, I propose to add the reverse directions to these rgb scripts.

While the "Horizontal" setting is running left to right, the new "Horizontal Reversed" setting is running right to left, same applies for the Vertical and Vertical Reversed setting to be used in matrices placed above each other.


mcallegari:
Isn't this achievable by playing the RGB matrix backward?

hjtappe:
Yes, for fillunfill and stripes, I agree and I reverted both files. But for fill.js, that is not the case. I also added the reverse mode now to devtools, so it's easier to evaluate.

PR #1330, opened 2022-05-02
Support a uDMX device that is sold as a Lixada

lbt:
This allows a device I bought from eBay that had Lixada branding to be identified as a uDMX device and operate with an ADJ ProfilePanel.

Based on the comment here https://www.qlcplus.org/forum/viewtopic.php?p=52529#p52529

Signed-off-by: David Greaves <david@dgreaves.com>

### DIFF
--- a/plugins/udmx/src/udmxdevice.cpp
+++ b/plugins/udmx/src/udmxdevice.cpp
@@ -36,6 +36,9 @@
 #define UDMX_AVLDIY_D512_CLONE_VENDOR     0x03EB /* Atmel Corp. (Clone VID)*/
 #define UDMX_AVLDIY_D512_CLONE_PRODUCT    0x8888 /* Clone PID */
 
+#define UDMX_LIXADA_D512_CLONE_VENDOR     0x0403 /* Lixada. (Clone VID)*/
+#define UDMX_LIXADA_D512_CLONE_PRODUCT    0x6001 /* Clone PID */
+
 #define UDMX_SET_CHANNEL_RANGE 0x0002 /* Command to set n channel values */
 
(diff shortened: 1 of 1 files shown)


mcallegari:
Sorry but I need to reject this. The same (real) VID/PID belong to FTDI and they are fundamental for the DMX USB plugin.
See here: https://github.com/mcallegari/qlcplus/blob/master/plugins/dmxusb/src/dmxinterface.h#L88

lbt:
Ah - I hadn't seen that.

So this device was initially recognised as a USB DMX but the serial number was garbage and it didn't operate at all - no LEDs on the usb part of the device.
When I applied this patch the serial number is seen and the device operates my panel properly, LEDs flicker etc. All good.

It sounds like there is a vanilla FTDI chip inside but this device is using the uDMX protocol.

When I tried selecting the "Pro RX/TX" mode in the DMX USB settings I got a segfault (which persisted and prevented the app starting- until the config was cleared - but that's a different issue) however, I  then saw your comment in the linked forum post and it made me wonder if it was a uDMX device - so I added the VID/PID and it worked. I didn't explore further.

So how to continue? I'm not familiar with the devices - should I be leaving it as a USB DMX and selecting a different mode or conceptually would you be OK with a patch that allowed UDMXDevice::isUDMXDevice() to return true based on some setting or config (maybe a "uDMX" entry in the settings dropdown)

mcallegari:
Mixing uDMX with DMX USB is not a viable option.
One thing that could be done is to add to the uDMX configuration dialog something like "Use custom USB VID/PID" with 2 text boxes. That would help also future clones to be detected.

mcallegari:
No reaction in a month. Closing.

Issue #1329, opened 2022-04-28
Restore CAPS regular expression


mcallegari:
The "port" to build with Qt6 required to switch QRegExp calls to QRegularExpression, which now supports only Perl-like regexp.
This commit 2bf07b313f3f629eb00478ce61504e1edfeb9b75 disables caps checking on fixture editor, but it needs to be restored with the proper syntax.

Most likely, there are other regexp around that need to be checked

hjtappe:
I could help with fixing regular expressions in 3 weeks if you need help and if there is more to be done.

mcallegari:
@hjtappe that would be a great help, thanks!

mcallegari:
Fixed by #1333

PR #1333, opened 2022-05-11
Re-Enable CAPS check in fixtureeditor and update other RexExp patterns

hjtappe:
Related to https://github.com/mcallegari/qlcplus/issues/1329.

Documentation
https://doc.qt.io/qt-5/qregularexpression.html#introduction

### DIFF
--- a/engine/src/qlcphysical.cpp
+++ b/engine/src/qlcphysical.cpp
@@ -266,7 +266,7 @@ int QLCPhysical::powerConsumption() const
         /* If power consumption value is missing, return bulb watts
          * plus a guesstimate 100W, since there's usually other
           * electronics consuming power as well. */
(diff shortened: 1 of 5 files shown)


hjtappe:
Do I interpret the build correctly that the fixture editor is not currently built on Travis CI? I would expect a QT='qt5' build for the fixtureeditor directory which is not built in companion with the top-level build. Should I try to add it or is it not available intentionally?

hjtappe:
Can you please take a look into the build configurations? The Appveyor build fails on some missing clang32 database and the Travis build fails on the installation on installing the required ppa.

mcallegari:
Hi, thanks for this patch.
Appveyor should be fixed now and Travis is failing because of unstable PPA. Sometimes it works, sometimes it doesn't. I hope they'll fix it.

mcallegari:
Thanks a lot! 👍

PR #1334, opened 2022-05-19
docs: remove "file:" scheme from links

awwright:
I'd like to submit a small series of documentation improvements, to make it easier to maintain, and browse and edit without exporting it first.

This is a straightforward fix, but a large one, that removes the `file:` scheme from links, which is redundant.

---

Adding `file:` in front of the URL reference forces it to be interpreted as a filesystem reference,
but this can be removed so that links will work without changes even over HTTP.


mcallegari:
Hi, I'm not sure about this.
It needs to be tested on Linux, Windows and macOS. Also with Qt 5.x and at this point Qt 6.x.
The documentation is accessible from within QLC+ through a Qt component. I take though it's broken as it is as plain html content and a solution would be nice. 
We added file: because it was needed. Honestly I don't remember the particular reason. Maybe now it can be removed for real.
I need to check myself

awwright:
I figure there's some reason for it, but even if so, I'd like to see if there's a better workaround.

I tested it with wkhtmltopdf 0.12.6 and it seemed to work, although I needed to add some command line arguments that they recently required. I'll file a PR for that separately.

awwright:
I think I found the source of the bug that required adding `file:` to everything, and it seems like a very simple fix. Qt seems to think that relative references are [always external references due to the external link detection](https://github.com/qt/qt/blob/0a2f2382541424726168804be2c90b91381608c6/src/gui/widgets/qtextbrowser.cpp#L217-L228). It is a simple fix by hooking into the `anchorClicked` signal.

I was able to go through all the links and test they work, including:

* Same-document references like `#Function`
* Relative-path references like `fixturedefinitioneditor.html` and `fixturedefinitioneditor.html#physical`
* External links like `http://os2l.org/`

mcallegari:
Very nice analysis and solution. Thank you very much

Issue #1308, opened 2021-12-27
A crash when using the Web interface


awwright:
I was playing around with the Web interface and after pressing lots of buttons, qlcplus crashed (SIGABRT from a failed assertion). It appears to be random, but I was able to eventually reproduce the crash inside gdb:

```
[Thread 0x7fffeb7fe640 (LWP 45970) exited]
ASSERT: "uint(i) < uint(size())" in file /usr/include/x86_64-linux-gnu/qt5/QtCore/qbytearray.h, line 500

Thread 1 "qlcplus" received signal SIGABRT, Aborted.
__pthread_kill_implementation (no_tid=0, signo=6, threadid=140737260009536) at pthread_kill.c:44
44	pthread_kill.c: No such file or directory.
(gdb) backtrace
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737260009536) at pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737260009536) at pthread_kill.c:80
#2  __GI___pthread_kill (threadid=140737260009536, signo=signo@entry=6) at pthread_kill.c:91
#3  0x00007ffff65fc476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff65e27b7 in __GI_abort () at abort.c:79
#5  0x00007ffff6aa3ba3 in  () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#6  0x00007ffff6aa2ff0 in qt_assert_x(char const*, char const*, char const*, int) () at /lib/x86_64-linux-gnu/libQt5Core.so.5
#7  0x00007ffff7e28044 in QByteArray::at(int) const (this=0x7fffffffdf90, i=2) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qbytearray.h:500
#8  0x00007ffff7677a83 in QHttpConnection::webSocketRead(QByteArray) (this=0x7fff18003360, data=...) at qhttpserver/qhttpconnection.cpp:440
#9  0x00007ffff7676419 in QHttpConnection::parseRequest() (this=0x7fff18003360) at qhttpserver/qhttpconnection.cpp:129
#10 0x00007ffff76a2f7f in QHttpConnection::qt_static_metacall(QObject*, QMetaObject::Call, int, void**)
[...]

PR #1335, opened 2022-05-20
bugfix: Rework WebSocket library to be stateful

awwright:
This avoids bugs when the IP or TCP packets arrive fragmented.

---

This needs a lot more work, but making the parser stateful seems to be what's necessary to prevent #1308.


mcallegari:
Just one overall note. QLC+ code indentation is like:
```
if (cond)
{
}
```
Please reindent like that, thanks

awwright:
I would be more confident if there's a linter I could use, is there one you prefer? Would it be possible to add one to the tests?

mcallegari:
Sorry no linter here 😢 
Just best practice:

PR #1336, opened 2022-05-20
Rgbscrips fix acceptColors attributes, formatting, removing unused variables

hjtappe:
Hello @mcallegari,

I took some in-depth time to review the attributes, usage and code of the RGB matrix scripts and have corrected the places which raised my attention. Please review if you agree with my findings.

- Set the acceptColors to a value as used in the script. Some scripts had no value set, which defaults to 2, but never used the second color.
- Set the acceptColors to 1 if the second color can neither be used in matrix object initalization and if the script does not evaluate the current progStep to always apply the same color in the same point in sequence of steps (i.e. it does not distinguish between forward / backward increments / decrements). Color change on the stage must be managed otherwise, e.g. through the virtual console to achieve a synchronous color change.
- Set the rgbMapStepCount to 2 if it is not relevant for the proceedings of the effect, especially to make the acceptColors=2 usable as the second color in instanciating matrix objects (apply the first / second color on an even / uneven step)
- Fix tabs to spaces
- Use the ECMAscript 2015 form and replace Class Functions by a Class with constructor.

### DIFF
--- a/engine/test/rgbscript/rgbscript_test.cpp
+++ b/engine/test/rgbscript/rgbscript_test.cpp
@@ -278,6 +278,13 @@ void RGBScript_Test::runScripts()
                 }
             }
         }
+        // Prepare a reference RGB map
+        bool randomScript = fileName.contains("random", Qt::CaseInsensitive);
+        RGBMap rgbRefMap;
+        if (1 < s.acceptColors() && 2 < steps && ! randomScript) {
+            // When more than 2 colors are accepted, the steps shall be reproducible to allow back and forth color fade.
+            s.rgbMap(mapSizePlus, red, 0, rgbRefMap);
+        }
         // Switch to the larger map and step a few times.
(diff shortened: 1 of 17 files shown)


hjtappe:
Hm, the Travis build is not active? Please check where it has gone, so it will confirm "make check" tests.

mcallegari:
Travis builds are active. It built PR 1334 and 1335, no idea why it didn't build 1336

hjtappe:
@mcallegari, I have reviewed all current scripts and checked if the usage of the second color and the color fade functionality related to the matrix step cound makes sense on stage. So, I reworked those scripts which apply an arbitrary fade to a second color to make better use of the second color and leave the color fade control to the VC matrix widget connected to a leading color channel.

On the way, I also added devtools capability to simulate alternating runs and a test extension to make the effect visible already during development.

I hope this makes sense.

mcallegari:
Merged, thanks! :+1:

PR #1337, opened 2022-05-23
Scene editor: Immediately update tabs in when fixture selection changes

awwright:
Currently, when the view is changed between the Tab View and the All Fixtures View, only the fixtures selected in the General tab are populated. This is somewhat confusing. The purpose appears to be so I can highlight a subset of fixtures, and "Copy current values to all fixtures" will only copy to the _selected_ fixtures.

This adds a signal to update the tabs and consoles immediately when the selection changes.

---

This is submitted as a draft because populating all of the tabs when the selection changes is a little sluggish, there's a few optimizations I'd like to work on.

Alternatively, if this feature was supposed to be removed, let's remove the code for that in the Tab/All channels toggle as well.


mcallegari:
Superseded by #1914

PR #1343, opened 2022-06-23
Rgbscript circular

hjtappe:
Hello Massimo,

I was in need for a script to apply some circular effects for a circular light arrangement. So I went to apply some more trigonometric functions to make up some circular effects, such as radar or circles.

Lights can now be arranged in a matrix like the following figure to apply the effects best.
image


mcallegari:
As usual, very nice job! :+1:

PR #1346, opened 2022-06-27
noop/cosmetic: remove duplicate resource includes in affected .ui files

edogawa23:
By chance I found out that most .ui files in ui/src and subdirs have a duplicated entry for including the common icon resource...

Obviously this didn't harm in any way, but I thought it's better to tidy this up.


mcallegari:
Makes sense, thanks for this! 👍 
I wonder how that could happen...maybe it's Qt creator doing this

PR #1347, opened 2022-06-28
Rgbscript test fix

hjtappe:
Hello @mcallegari, I remember you raised a problem with old ECMAscript syntax like "let", "map", "Class" and inline functions. I couldn't reproduce it at that time, so we forgot about it and continued to develop mainly with v5 in mind.

Now, after the fresh 4.12.5 release, I wanted to test the scripts, but they did not show up, so I started to debug.

Finally, it brought me back to the ancient JS engine still used in v4 and I added a test in the rgbscripts_test.cpp to find the bad pieces. This patch now refactors the scripts, so they should work again in v4.

Sorry with the inconvenience in the surprising but well received v4 release that was just published.


mcallegari:
Hi, thanks for this.
I just checked on v5 with Qt 5.15.2 and all the current scripts load correctly without this patch.
I think the problem was with v5 and Qt 5.14.2 but before your previous patches (maybe #1336 did the trick?)
I will include this patch too as it seems to help some more.

PR #1352, opened 2022-07-18
Rgbscripts build config

hjtappe:
The alternate.js script was missing in the .pro build file. I also extended a test to detect that in the future.


mcallegari:
Thanks 👍

PR #1353, opened 2022-08-14
Add Cameo F2 T PO fixture

hjtappe:
Technical Data Sheet and DMX Control Table are available at
https://www.cameolight.com/en/solutions/rental/static-lighting/theater-lights/20477/f2-t-po


mcallegari:
Hi, thanks for this.
Please remember that a channel cannot act on itself! Acts On is used only when a channel affects another channel (e.g. Gobo Wheel and Gobo Wheel Rotation)

PR #1357, opened 2022-08-19
Fixtures tool in unittests

hjtappe:
Hello @mcallegari, when you asked me to remember the usage of the "ActsOn" in fiture definitions, I thought, I'd like to add me a "reminder" in the check script and tests.

I found different things along the way:
- The fixture_tools.py did not work with python3-lxml, which is now the standard.
- The fixture_tools.py did not validate when used as specified in the header
- The unittests did not execute fixture_tools.py

So I reworked the fixture_tools.py and the unittests chain to retrieve the results. And added the check for the ActsOn usage.

I found two fixtures which failed on validation and corrected them.

I hope it helps to maintain the fixture contributions by already checking things in the pull request build.


hjtappe:
@mcallegari,

- I don't find a way to satisfy Codacy with a method name that is not "too complex" (by taking over the original name and modifying the arguments).
- The travis build is not triggered (even before I added the python-lxml and changed the indentation). I validated the modified travis.yml at https://config.travis-ci.com/explore

Can you please have a brief look if you see what I still should improve to satisfy the build?

mcallegari:
Since one year or so, Travis works with credits. 😢 
Credits are free of charge for open source projects but they give me very few and every build consumes credits.
Right now builds are disabled because I ran out of credits...again.
Lately I try to pack commits before pushing to consume less credits, but I cannot do anything on PRs other than disable CI which I would avoid.

I'm gonna ask for more credits, but please try to reduce the number of pushes if you can. Thanks

hjtappe:
As far as I read, the credits are consumed per build minute.

I usually collect and build all fine grained commits before I push them here, so that should be fine - but when pushing e.g. a fixture the build consumes a lot of credits to build the C++ code instead of running only the validation. The same goes if Codacy errors are to be fixed, thus a new full build is triggered, not just the Travis build.

- Is there anything we could help with, e.g. using github actions to reduce the travis load?
- Is there anything to bill the travis build before the pull request on my fork, thus on my account?
- Is there a way to reduce the load by separating v5 builds and v4 builds into branches or deactivating the coverage build to reduce total build time?

Thanks for organizing more credits.

mcallegari:
@hjtappe sorry for the delay
At this point Github actions would solve a few issues. I had a rough look at them and they should do the job.
Also, it would be interesting if there was a way to "select" the build flavor with a commit. E.g. full build + tests (default), fixtures check, code coverage, etc
Not sure if this possible at all though

mcallegari:
hi @hjtappe is this ready to be merged?

hjtappe:
> hi @hjtappe is this ready to be merged?

@mcallegari, yes, unless you want me to work further to satisfy the function complexity of the fixture check ad advised by Codacy by extracting some further functionality into separate functions. I think, it would not add significant value.

mcallegari:
Don't worry about Codacy.
Sometimes it's too picky even if the code is OK.
I'll merge then, thanks :+1:

mcallegari:
I just did what I usually did and got this:
```
# scripts/fixtures-tool.py --validate --map
Namespace(convert=None, map=True, validate=[])
Traceback (most recent call last):
  File "scripts/fixtures-tool.py", line 752, in <module>
    createFixtureMap()
  File "scripts/fixtures-tool.py", line 691, in createFixtureMap
    xmlFile.write(etree.tostring(root, pretty_print=True, xml_declaration=True, encoding="UTF-8", doctype="<!DOCTYPE FixturesMap>").decode('utf8'))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xb0' in position 33577: ordinal not in range(128)
```

hjtappe:
Sorry for the problem. I was not thinking about python 2.7 being still a thing. :-)
#1363 will support both python versions and distuinguish between the different needs depending on the underlying version.

mcallegari:
Hey @hjtappe are you sure validation is working also on Python 2.7?
This function doesn't seem to load hasPan and hasTilt correctly to me:
https://github.com/mcallegari/qlcplus/blob/master/resources/fixtures/scripts/fixtures-tool.py#L637

If I set pan or tilt degrees to 0 on some moving head, I get no errors at all. This is not expected.
Can you please have a look?
I wanted to add also zoom degrees validation, that's how I discovered this

hjtappe:
You can integrate the patch from #1365. My bad. I should take more time to review and test that, even if I only refactor the complexity.

PR #1360, opened 2022-09-05
Add Vari-Lite VL4000 Spot fixture definition

HON95:
This is the first fixture definition I'm attempting to add, I'm happy to fix any functional or quality problems with it. It follows the official DMX tables for the fixture, minus the color wheel "mid-positions" (between two colors) which aren't described by the tables, but which were also used by the other Vari-Lite fixtures.


mcallegari:
Hi, thanks for this.
Being so huge I think for now it's mostly functional.
It misses a few details (e.g. gobos, channels relationships) and I think I need to add some new presets for framing systems

PR #1361, opened 2022-09-14
RGB Script which alternates exactly between two halfs

DrFreas:
Splits a Matrix in half and alternates between both halfs, vertically or horizontally.

### DIFF
--- a/resources/rgbscripts/splitinhalf.js
+++ b/resources/rgbscripts/splitinhalf.js
@@ -1,0 +1,101 @@
+/*
+  Q Light Controller
+  splitinhalf.js
+
+  Copyright (c) Jan Fries
+
+  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


hjtappe:
Can this already be achieved by setting the blocksize in the alternate.js script? Would it make sense to extend that one or is it reflecting another use-case?

DrFreas:
I also saw the alternate.js script, but the block size is limited to 16, as far as I see it. This script here would simply divide a matrix in two halfs, regardless of size. Also at least I use it for live shows where I choose the effects from the RGB-effect widget thingy and setups change frequently, so I won't have to worry about any parameters I would have to set to adapt.

mcallegari:
To be honest I would extend the alternate.js script with a new parameter like "divide in halves"
Too many scripts with similar purposes would create confusion

mcallegari:
Addressed by #1414

PR #1363, opened 2022-09-26
Fixtures tool - support --map with python 2.7

hjtappe:
@mcallegari, under Python 3.x, the xmlFile.write function requires the .decode('utf8') to be set, under 2.7, it is not supported. The lxml.etree documentation was not explicit but threw an error due to the UTF-8 encoding, which indicated in forums that this "decode" extension would return the same results as the 2.7 output, including the file encoding.


mcallegari:
Thanks

PR #1365, opened 2022-09-30
…Move physicalcheck into validate_fx_channels

hjtappe:
Move physicalcheck into validate_fx_channels as it requires hasPan and hasTilt

Fixes the problem discussed in  #1357 by @mcallegari 
Sorry, that happened in the refactoring to make smaller functions due to Codacy complexity warnings. I should have seen that in my review.


mcallegari:
Fixed by e5e37ac3c4ea0a865293591e7838d45f896159a2

PR #1366, opened 2022-10-10
ui: do not focus function name text when function is selected

pomowunk:
This modifies the behavior in the function manager when a function is selected.
Currently, the nameEdit widget in the function editor is explicitly focused and the text is selected.
Because focus changes to the text, it breaks keyboard navigation and shortcuts (clone, delete...) in the FunctionsTreeWidget.

This pull request does not set the focus, this allows the expected interaction with the tree view, for example:
- traversing the function tree with the arrow keys
- clicking on functions, then using the clone or delete key shortcut

If the function name should be selected for immediate editing for newly created functions, the old behavior could be kept for this special case.

If you agree, I can follow this pull request with an adjustment of the tab order in the function editors, so the selected treeview item leads to the function name and then the function parameters.

### DIFF
--- a/ui/src/audioeditor.cpp
+++ b/ui/src/audioeditor.cpp
@@ -43,7 +43,6 @@ AudioEditor::AudioEditor(QWidget* parent, Audio *audio, Doc* doc)
     setupUi(this);
 
     m_nameEdit->setText(m_audio->name());
-    m_nameEdit->setSelection(0, m_nameEdit->text().length());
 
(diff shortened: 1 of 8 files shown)


mcallegari:
Sorry but no.
This is a breaking change and while it works for you I'm pretty sure it will make many other users unhappy (me included)
I usually hit the clone button and rename right away, which I find pretty convenient.
The real fix here would be to consider global shortcuts even when Function label is focused.
Basically a totally different PR.
Closing for now.

PR #1367, opened 2022-10-21
Fix starfield RGBmatrix script to use the full matrix

hjtappe:
Fixes issue reported at
https://www.qlcplus.org/forum/viewtopic.php?t=15867


mcallegari:
Thanks. Ran out of Travis credits again...we should really consider migrating Linux build to GitHub actions :(

hjtappe:
> Thanks. Ran out of Travis credits again...we should really consider migrating Linux build to GitHub actions :(

I am on my way to define a GH Actions yaml file for it and made some progress to avoid some setup obstacles. It is not yet at the point to start a build but it should be possible. Give me a few more days.  🙂

PR #1368, opened 2022-10-24
Bug Fixture Remapping VCWidgets

Off-IT:
Related to this forum post : https://www.qlcplus.org/forum/viewtopic.php?f=20&t=15864
Somehow some VCWidgets are listed twice (or more ?) with getVCChildren. The remapping happens then more than once, making confusion in fixture IDs at the second remapping.
Just removing duplicate VCWidget from the list using a Set.

### DIFF
--- a/engine/src/doc.cpp
+++ b/engine/src/doc.cpp
@@ -579,6 +579,8 @@ bool Doc::replaceFixtures(QList<Fixture*> newFixturesList)
         }
 
         newFixture->setExcludeFadeChannels(fixture->excludeFadeChannels());
+        newFixture->setForcedHTPChannels(fixture->forcedHTPChannels());
+        newFixture->setForcedLTPChannels(fixture->forcedLTPChannels());
         m_fixtures.insert(id, newFixture);
         m_fixturesListCacheUpToDate = false;
 
--- a/engine/src/doc.cpp
+++ b/engine/src/doc.cpp
@@ -579,6 +579,9 @@ bool Doc::replaceFixtures(QList<Fixture*> newFixturesList)
         }
 
         newFixture->setExcludeFadeChannels(fixture->excludeFadeChannels());
+        newFixture->setForcedHTPChannels(fixture->forcedHTPChannels());
+        newFixture->setForcedLTPChannels(fixture->forcedLTPChannels());
+
         m_fixtures.insert(id, newFixture);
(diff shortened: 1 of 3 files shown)


mcallegari:
How is it possible that a widget is listed twice?

Off-IT:
I haven't debugged that far. I had a conditional break point on the id of the slider widget that I knew was problematic. That break point was hit twice. That's how I found out there were duplicates in the list.

mcallegari:
getVCChildren is recursive. Hitting a breakpoint twice doesn't mean there are duplicates.
Is this patch actually solving the issue you describe in the forum?
Debugging this issue with 65 fixture like in your project is impossible for me.

Off-IT:
Yes, this patch solves the issue I described in the forum.
The break point I set was in the accept function in the for loop after the call to getVCChildren like this :  ( line 831 )
`if (slider->sliderMode() == VCSlider::Level)
            {
                if (widget->id() == 19) {  // ID of the widget I knew was problematic.
                    QDebug() << "19"; // break point here
                }`
At that point the list shouldn't have any duplicates so that's why I thought about using a QSet here.

I also noticed that one to one remapping didn't keep the HTP/LTP settings on the fixture. That's solved the third commit of this pull request ( probably should have addressed this in a new PR ).

mcallegari:
Ok. The HTP/LTP patch is legit.
Right now I'm adding fixture import code as you suggested.
I really would like to understand this ID duplication but I'd like to have a simplified project to reproduce the issue

mcallegari:
I have pushed the import code and your patch for HTP/LTP.
I'm gonna analyze the ID duplication later

Off-IT:
Great thanks.
In the meantime I have come with a simplified project that reproduce the issues. It uses uses 2 fixtures, 2 sliders and 3 frames. Each slider is in its own frame and those two frames are in another frame. Although the remapping ( cloning ) has no issues, the sliders are listed twice. Seems like something's wrong with the nested frames.
As I can't attach a .qxw file here, I'll upload it in the forum post.

mcallegari:
Finally got it! See 5f2bf86868b5c714c5fbbbcf0d2a5c59d2d82be3
Thanks for the simplified project. It helped indeed.
Basically QObject::findChildren is already recursive unless explicitely specified!
Thanks for reporting :+1:

PR #1372, opened 2022-10-30
OSC: Bind to QHostAddress::Any instead of the IP of the interface

kripton:
This fixes OSC broadcast reception on Linux as discussed here: https://www.qlcplus.org/forum/viewtopic.php?f=12&t=15647

Reception and Transmission of OSC packages tested using wireshark.


mcallegari:
Thanks :+1:

PR #1375, opened 2022-11-10
Read a local plugin configuration from plugins.config

awwright:
When I have several different branches checked out, changes to the plugins/plugins.pro file become difficult to manage because you can't rebase unless the working tree is completely clean. And using "git stash" causes conflicts. Moving local configuration to a separate, untracked file resolves this.

For example, my "plugins.config" file looks like:

```
SUBDIRS = artnet loopback osc
```

### DIFF
--- a/plugins/plugins.pro
+++ b/plugins/plugins.pro
@@ -27,3 +27,9 @@ SUBDIRS              += artnet
 SUBDIRS              += E1.31
 SUBDIRS              += loopback
 SUBDIRS              += osc
+
+# Allow a local configuration file (not checked in) to override
+# or add to the defaults specified here
+exists(plugins.config) {
+  include(plugins.config)
+}


mcallegari:
Sorry but no.
I'm not gonna introduce complexity in QLC+ for an issue that can somehow be resolved with git or a slightly different workflow.

awwright:
@mcallegari But this can't be resolved with Git, I explained why. If you think you have a better solution you should explain what that is. But right now I don't think anyone knows what that is.

And adding code is not necessarily the same as adding complexity. If this makes the overall development process simpler then quite the opposite, it will make development easier to think about.

awwright:
@mcallegari I've got a laundry list of things I'm hoping to work on and this is getting in the way. Do you have a better way to work around this problem or not?

mcallegari:
The plugins.pro file changed 3 times in 6 years. The git tool has a plethora of ways to work with branches and merges. You can even create a patch with git patch and apply it later with git apply. Also, instead of git rebase you can try git merge.
You in the first place didn't explain why you need this patch, on which OS you work, why you require less plugins than the ones bundled, how many branches you deal with, what are the conflicts you see, etc.
You just want to change QLC+ because you cannot handle your own changes. For me "my config" doesn't justify a "let's do it for everyone".
You are free to modify QLC+ as you desire, but please don't expect this project to agree on whatever change you need for yourself. 
If every single one who forked QLC+ would send patches like this, can you imagine what this project would turn into in a few years?
Sorry but for me this conversation is over.

awwright:
> You just want to change QLC+ because you cannot handle your own changes.

That's the problem that this is fixing. I shouldn't have to commit my local changes to Git and force it on everyone; there should be a local configuration file.

> You in the first place didn't explain why you need this patch

I explained it in the first paragraph:

When I have several different branches checked out, changes to the plugins/plugins.pro file become difficult to manage because you can't rebase unless the working tree is completely clean. And using "git stash" causes conflicts. Moving local configuration to a separate, untracked file resolves this.

> Also, instead of git rebase you can try git merge

git merge is not a substitute for rebase in multi-branch strategies.

> on which OS you work

This affects all platforms, though I use Ubuntu, Gentoo, macOS, and Windows.

>  If every single one who forked QLC+ would send patches like this, can you imagine what this project would turn into in a few years?

You're being obtuse; if every change that made development easier was merged in, then development become easier, not more complicated.

PR #1376, opened 2022-11-11
update audio preview when a different file is loaded

edogawa23:
See this forum [thread]:(https://www.qlcplus.org/forum/viewtopic.php?f=22&t=15924)

This fixes the waveform display at new file load, for me.

### DIFF
--- a/ui/src/showmanager/audioitem.cpp
+++ b/ui/src/showmanager/audioitem.cpp
@@ -140,6 +140,11 @@ Audio *AudioItem::getAudio()
 
 void AudioItem::slotAudioChanged(quint32)
 {
+    PreviewThread *waveformThread = new PreviewThread;
+    waveformThread->setAudioItem(this);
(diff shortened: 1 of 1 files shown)


mcallegari:
Hi, thanks for this.
Does it make more sense to move this block of code right after the new width calculation?
Or between width calculation and new duration set? (I don't remember who exactly triggers a repaint)

edogawa23:
We are starting a thread to calculate the new image, isn't it reasonable to do that early so it can finish sooner/process in parallel with the other block?

edogawa23:
I had a closer look, and see no explicit manual (re)painting except via update() during PreviewThread::run()

mcallegari:
prepareGeometryChange requests a repaint (so in our case a width change)
https://codebrowser.dev/qt5/qtbase/src/widgets/graphicsview/qgraphicsitem.cpp.html#7698

Anyway, I'll merge this and do some more tests myself. Thanks :+1:

PR #1388, opened 2023-02-02
Minor linux changes while looking at qml mode

kiilerix:
The appimage changes are blind and untested

### DIFF
--- a/platforms/linux/linux.pro
+++ b/platforms/linux/linux.pro
@@ -4,11 +4,17 @@ TEMPLATE = subdirs
 TARGET   = icons
 
 desktop.path   = $$INSTALLROOT/share/applications/
-desktop.files += qlcplus.desktop qlcplus-fixtureeditor.desktop
+desktop.files += qlcplus.desktop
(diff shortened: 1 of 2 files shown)


mcallegari:
I don't think this is correct.
At some point v5 will replace v4, so there's no need to duplicate everything with a different name.
Linux users will just use the appimage until there will be a new qlcplus-5*.deb package

kiilerix:
Ok. The two versions are currently almost side-by-side compatible. If taken all the way, that could perhaps help getting more feedback and contribution to v5.
But it could of course also go the other way and make it a configuration time choice if v4 or v5 should end up in the qlcplus binary. Would you like the name qlcplus-qml to go away?

kiilerix:
> Linux users will just use the appimage until there will be a new qlcplus-5*.deb package

FWIW: As Fedora user I prefer an rpm. I'm not so inclined to test something that isn't "ready". I would rather push forward on getting it ready in all ways.
Appimage is not native and will get less / different test coverage.

Issue #1390, opened 2023-02-12
Concurrent MIDI notes not correctly read on Mac


Timebutt:
**Describe the bug**
A MIDI note that is immediately preceded by a MIDI CC value is not being read by QLC+, causing whatever is listening to that signal to not fire.

I have attached two MIDI files as well as a QLC+ project file to this issue to illustrate the problem. If you open `Test.qxw`, only `Scene 5` and the `Slider` are of note. The idea is that at a certain point in time, my MIDI track sends both MIDI CC to control the fader level (you'll see it bounce around as you play either of the supplied MIDI tracks), as well as at the exact same time send a `Note On` to trigger `Scene 5`.

If you first play `not-working.midi`, you'll see that `Scene 5` does not get triggered, and the MIDI timeline looks like this when viewed with a debugging tool (https://www.midimonitor.com/ is a pretty cool one):

```
16:49 23:567 | Virtual MIDI Bus 1 | Control Change | CC#1 Value 0 | 1
16:49 23:567 | Virtual MIDI Bus 1 | Note On | Note #63 Velocity 100 | 1
```

The MIDI note is clearly sent, but is somehow ditched by QLC+.
If we now take a look at `working.midi`, you'll see that `Scene 5` correctly gets triggered and all is well! The timeline looks like this:

```
16:52 45:316 | Virtual MIDI Bus 1 | Note On | Note #63 Velocity 100
16:52 45:508 | Virtual MIDI Bus 1 | Control Change | CC#1 Value 1 | 1
```

I've noticed that there's never a problem if the `Note On` message is sent before the CC value gets sent. Whenever the CC value is received right before the `Note On` message, the `Note On` is ignored by QLC+.

**To Reproduce**
1. Open `Test.qxw` from the zip archive I provided (see **Additional context**)
2. Load `not-working.midi` in any type of software that can play back a MIDI file
3. Configure QLC+ to correctly receive MIDI messages 
4. Verify `Scene 5` does not triggered when playing back `not-working.midi` even though the correct note gets sent!
5. Compare with `working.midi` file, which has the almost identical MIDI information, but the first CC message is delayed the tiniest amount
6. Verify `working.midi` playback results in `Scene 5` correctly being triggered in QLC+

**Expected behavior**
MIDI notes should always correctly be read by QLC+, regardless of the presence of any CC values.

**Desktop (please complete the following information):**
 - OS: MacOS Ventura `13.2`
 - QLC+ Version  `4.12.6`

[...]

PR #1394, opened 2023-02-18
Fix concurrent MIDI issue on Mac

Timebutt:
This issues fixes issue #1390.

### DIFF
--- a/plugins/midi/src/macx/coremidiinputdevice.cpp
+++ b/plugins/midi/src/macx/coremidiinputdevice.cpp
@@ -49,7 +49,7 @@ static void MidiInProc(const MIDIPacketList* pktList, void* readProcRefCon,
             uchar value = 0;
 
             // MIDI Command
-            cmd = packet->data[0];
+            cmd = packet->data[i];
             if (!MIDI_IS_CMD(cmd))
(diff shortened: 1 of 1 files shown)


mcallegari:
I see this patch is for macOS.
Have you checked the issue doesn't happen on Windows and Linux?

Timebutt:
I have not been able to reproduce this on Windows or Linux yet, however: when I look at the implementation of `win32midiinputdevice.cpp`, I don't see code that could lead to the same issue. I will try and set up the same situation (a virtual MIDI device is harder to set up on Windows than Mac). 

I have meanwhile renamed the PR so that it more accurately reflects the issue.

kpr0th:
For what it's worth, I just tried testing with LoopBe on Win11.  I used MultiPlay 3.x to send 2 MIDI messages in one click, and both of them came through successfully (I added some temporary qDebug lines to prove they both came into win32midiinputdevice.cpp::MidiInProc(), and they came through as two separate calls to this function. As such, I suspect the loop fix proposed for macx doesn't apply in the windows world.

Timebutt:
Awesome, thanks for checking @kpr0th!

PR #1396, opened 2023-02-18
Cuelist slider fix

kpr0th:
Adjust CueList slider value calculation to fully align ValueChanged->Step# and Step#Changed->Value calculations.

Previously, clicking the PreviousStep button didn't always update the slider position, due to off-by-one issues          near whole-numbered calculated values.  Example: with a 6-step chaser, advance to the end, then click the Previous Cue button and note that the slider only jumps up every other click. Also, if you move the slider position with keystrokes up and down (or presumably also via a midi encoder wheel up/down), note that it jumps to the next step at slightly different values than the ones that are calculated when using next/previous step buttons -- and then the displayed value at the top of the slider "jumps" to the top end of the subset of values calculated to belong to the newly selected chaser step #, which causes the visual feedback to "jump".

### DIFF
--- a/qmlui/virtualconsole/vccuelist.cpp
+++ b/qmlui/virtualconsole/vccuelist.cpp
@@ -260,13 +260,13 @@ void VCCueList::setSideFaderLevel(int level)
         if (ch->stepsCount() < 256)
         {
             float stepSize = 256 / float(ch->stepsCount()); //divide up the full 0..255 range
-            stepSize = qFloor((stepSize * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
+            stepSize = (float)qFloor((stepSize * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
             if(level >= 256.0 - stepSize)
                 newStep = ch->stepsCount() - 1;
             else
                 newStep = qFloor(qreal(level) / qreal(stepSize));
+            //qDebug() << "value:" << value << " new step:" << newStep << " stepSize:" << stepSize;
         }
-        //qDebug() << "value:" << value << " steps:" << ch->stepsCount() << "new step:" << newStep;
 
         ChaserAction action;
         action.m_action = ChaserSetStepIndex;
--- a/qmlui/virtualconsole/vccuelist.cpp
+++ b/qmlui/virtualconsole/vccuelist.cpp
@@ -259,13 +259,14 @@ void VCCueList::setSideFaderLevel(int level)
         int newStep = level; // by default we assume the Chaser has more than 256 steps
         if (ch->stepsCount() < 256)
         {
-            float stepSize = 255 / float(ch->stepsCount());
-            if (level >= 255 - stepSize)
+            float stepSize = 256 / float(ch->stepsCount()); //divide up the full 0..255 range
+            stepSize = qFloor((stepSize * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
+            if(level >= 256.0 - stepSize)
                 newStep = ch->stepsCount() - 1;
             else
                 newStep = qFloor(qreal(level) / qreal(stepSize));
         }
-        //qDebug() << "value:" << value << "steps:" << ch->stepsCount() << "new step:" << newStep;
+        //qDebug() << "value:" << value << " steps:" << ch->stepsCount() << "new step:" << newStep;
 
         ChaserAction action;
         action.m_action = ChaserSetStepIndex;
--- a/qmlui/virtualconsole/vccuelist.cpp
+++ b/qmlui/virtualconsole/vccuelist.cpp
@@ -260,8 +260,8 @@ void VCCueList::setSideFaderLevel(int level)
         if (ch->stepsCount() < 256)
         {
             float stepSize = 256 / float(ch->stepsCount()); //divide up the full 0..255 range
-            stepSize = (float)qFloor((stepSize * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
-            if(level >= 256.0 - stepSize)
+            stepSize = qFloor((stepSize * 100000.0) + 0.5) / 100000.0; //round to 5 decimals to fix corner cases
+            if (level >= 256.0 - stepSize)
                 newStep = ch->stepsCount() - 1;
             else
                 newStep = qFloor(qreal(level) / qreal(stepSize));
--- a/ui/src/virtualconsole/vccuelist.cpp
+++ b/ui/src/virtualconsole/vccuelist.cpp
@@ -818,16 +818,17 @@ void VCCueList::slotCurrentStepChanged(int stepNumber)
         float stepVal;
         int stepsCount = m_tree->topLevelItemCount();
         if (stepsCount < 256)
-            stepVal = 255.0 / (float)stepsCount;
+            stepVal = 256.0 / (float)stepsCount; //divide up the full 0..255 range
         else
             stepVal = 1.0;
-        int slValue = (stepVal * (float)stepNumber);
+        // value->step# truncates down in slotSideFaderValueChanged; so use ceiling for step#->value
+        float slValue = stepVal * (float)stepNumber;
         if (slValue > 255)
-            slValue = 255;
+            slValue = 255.0;
 
-        int upperBound = 255 - slValue;
-        int lowerBound = qFloor(upperBound - stepVal);
-        //qDebug() << "Slider value:" << m_slider1->value() << "Step range:" << (255 - slValue) << (255 - slValue - stepVal);
+        int upperBound = 255 - qCeil(slValue);
+        int lowerBound = qFloor((float)256.0 - flValue - stepVal);
+        //qDebug() << "Slider value:" << m_sideFader->value() << "->" << 255-slValue << "( disp:" << slValue << ") Step range:" << upperBound << lowerBound;
         // if the Step slider is already in range, then do not set its value
         // this means a user interaction is going on, either with the mouse or external controller
         if (m_sideFader->value() < lowerBound || m_sideFader->value() >= upperBound)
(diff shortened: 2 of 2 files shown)


mcallegari:
I'm not really convinced by those +/- 0.00001
Can you please share the workspace you were using to test this so I can have a look too?
Thanks

kpr0th:
Yes, I'll try to pack it up tonight after work.  And maybe throw a quick video together to walk through it.  For now I'll try explaining with words, in case that's enough.

I was testing with a super-simple workspace - 1 scene dropped into a chaser 6 times, and attached to the cuelist.  Also tested with it in the chaser 18 times.  I watched the debug closely (I actually changed it to qWarning temporarily so I could ignore all the other debug output).  And then I used the NextCue/PrevCue buttons and keyboard arrows (up/down) to move the steps and slider.

With 6 steps, each step gets 42 2/3 of the 256 DMX values (42.6666666667). First step starts at 255 and ends at 212.333333, so it "owns" the range from 255 down to 213; in other words, any value between 213 and 255 inclusive should equate to step # 1. 
The second step's starting location gets truncated down to 212, and ends at 169.666666, so it "owns" the range from 212 down to 170.  
Third step starts at 169, and the fourth step starts at *exactly* 127, so the third step "owns" the range from 169 down to 128.  And this is where the problem comes in - without those +/-0.00001's, the calculation was deciding that the third step owns 169 down to **127**, and the fourth step owned 127 down to 85, and if 2 steps share one DMX value then it was unclear which one really owned that value - and when I arrowed "up" toward the top of the slider, it would jump when it got to 127 (if i recall correctly).

I picked 0.00001 because floats keep 6-8 digits of precision, and the smallest # of DMX values assigned to one step is 256/255 = 1.00392.  So the 0.00001 is large enough to affect the floating point numbers, but small enough to never push any one step's range over an integer boundary that it shouldn't (at least, that's my belief).

kpr0th:
p.s. I was thinking I was dealing with floating point rounding errors when I was testing the code change. I thought the value that was supposed to be an exact integer was somehow looking like it was slightly less than the integer value. But after responding to your question above, I realize that the real issue I'm solving is that an integer value comes out as the same number if you apply qCeil or qFloor, but to make sure each step owns a specific range of DMX values I needed all the possible dividing lines to return one value after qCeil() and a different value after qFloor()...  So, the comment in the code I checked-in is misleading; I'll try to fix that.  I'll still get you a workspace and a few pictures or a video tonight.

kpr0th:
> I'm not really convinced by those +/- 0.00001 Can you please share the workspace you were using to test this so I can have a look too? (you'll have to zip it to attach here) Thanks

Attaching a zip, containing the test qxw plus a series of PNG's to try and illustrate what I was seeing.

[cuelist-test.zip](https://github.com/mcallegari/qlcplus/files/10809583/cuelist-test.zip)

kpr0th:
> I'm not really convinced by those +/- 0.00001 Can you please share the workspace you were using to test this so I can have a look too? (you'll have to zip it to attach here) Thanks

@mcallegari:  I finally took some time to get back to this.  I setup a bit of quick test code to run all possible values and look for corner cases.  In the process I realized there were a couple of other 255's that should have been 256's.  And I found a few more corner cases that I wasn't looking at before.  By rounding off the "stepSize" variable to 5 decimals I was able to eliminate all of them.  So I removed the +/- 0.00001's, rebased on my feature branch, and now I'm confident this is a solid fix without the hack.  Will you please re-consider this pull request?

(and if I somehow messed up the re-basing, let me know and I can re-create my change on a clean branch instead -- I'm not all that experienced with git yet)

kpr0th:
@mcallegari  :  you had asked for info on the issue I was trying to fix.  I posted a zip file back on Feb 22 with pics showing the issue in the upstream/master, along with how my fix adjusted it.  I just checked, and the original issue still exists in the latest upstream codebase.  My updated commit still fixes it.  In case the pictures aren't making the issue obvious enough, you can reproduce the problem using the Feb 22 workspace in v4:

Go online with the virtual console
Press play on the cuelist widget
Press "Next Cue" twice - observe that the side slider moves each time, and the number at the top also changes
Press "Previous Cue" twice - observe that the side slider only moves on the second button press, and the number at the top also doesn't change until the second button press, even though the selected cue changes both times.

Expected behavior: side slider moves every time previous or next cue is pressed, and the number at the top of the slider also changes.

Another symptom of the issue:

Go online with the virtual console
Press play on the cuelist widget
Press "Next Cue" once - observe that the 2nd cue is selected, the side slider moves down, and the top number shows 42.
Click on the side slider with your mouse, and then press the down arrow on your keyboard (or, use an external input device to send a "43" to the cuelist widget) - observe that the number changes to 43.   
Then press the up arrow on your keyboard (or send in a 42 from an external input device) - observe that the number changes back to 42, and the selected cue changes to Cue 1.

Expected behavior: since 42 is the top of the 2nd cue's range, sending a 42 should stay on cue 2.

ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -817,25 +817,32 @@ void VCCueList::slotCurrentStepChanged(int stepNumber)
 
         float stepVal;
         int stepsCount = m_tree->topLevelItemCount();
-        if (stepsCount < 256)
-            stepVal = 255.0 / (float)stepsCount;
-        else
+        if (stepsCount < 256) {
```
mcallegari, line 820: Please stick to the style of the rest of the code

kpr0th: Oops - and this after you just checked in a huge lot of code style fixes barely a week ago. Sorry about that. Fixed, and pushed to this branch just now.  Awaiting checks but hopefully they'll complete just fine in a bit.

qmlui/virtualconsole/vccuelist.cpp
```diff
@@ -259,13 +259,14 @@ void VCCueList::setSideFaderLevel(int level)
         int newStep = level; // by default we assume the Chaser has more than 256 steps
         if (ch->stepsCount() < 256)
         {
-            float stepSize = 255 / float(ch->stepsCount());
-            if (level >= 255 - stepSize)
+            float stepSize = 256 / float(ch->stepsCount()); //divide up the full 0..255 range
+            stepSize = (float)qFloor((stepSize * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
```
mcallegari, line 263: stepSize is already float, so no need to cast it to float again

qmlui/virtualconsole/vccuelist.cpp
```diff
@@ -259,13 +259,14 @@ void VCCueList::setSideFaderLevel(int level)
...
+            if(level >= 256.0 - stepSize)
```
mcallegari, line 264: Code style: space after 'if'

ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -817,25 +817,36 @@ void VCCueList::slotCurrentStepChanged(int stepNumber)
 
         float stepVal;
         int stepsCount = m_tree->topLevelItemCount();
-        if (stepsCount < 256)
-            stepVal = 255.0 / (float)stepsCount;
-        else
+        if (stepsCount < 256) 
+        {
+            stepVal = 256.0 / (float)stepsCount; //divide up the full 0..255 range
+            stepVal = (float)qFloor((stepVal * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
```
mcallegari, line 823: same as above. Cast not needed

ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -817,25 +817,36 @@ void VCCueList::slotCurrentStepChanged(int stepNumber)
...
+        float slValue = stepVal * (float)stepNumber;
         if (slValue > 255)
-            slValue = 255;
+            slValue = 255.0;
 
-        int upperBound = 255 - slValue;
-        int lowerBound = qFloor(upperBound - stepVal);
-        //qDebug() << "Slider value:" << m_slider1->value() << "Step range:" << (255 - slValue) << (255 - slValue - stepVal);
+        int upperBound = 255 - qCeil(slValue);
+        int lowerBound = qFloor((float)256.0 - slValue - stepVal);
```
mcallegari, line 836: I think casting to float is not needed here too since you specified .0 which the compiler interprets as float

ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -1216,13 +1227,14 @@ void VCCueList::slotSideFaderValueChanged(int value)
         int newStep = value; // by default we assume the Chaser has more than 256 steps
         if (ch->stepsCount() < 256)
         {
-            float stepSize = 255.0 / (float)ch->stepsCount();
-            if (value >= 255.0 - stepSize)
+            float stepSize = 256.0 / (float)ch->stepsCount();  //divide up the full 0..255 range
+            stepSize = (float)qFloor((stepSize * 100000) + 0.5) / 100000; //round to 5 decimals to fix corner cases
```
mcallegari, line 1231: Cast not needed. Explicitly use 100000.0

ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -1216,13 +1227,14 @@ void VCCueList::slotSideFaderValueChanged(int value)
...
+            if(value >= 256.0 - stepSize)
```
mcallegari, line 1232: Code style: space after 'if'

qmlui/virtualconsole/vccuelist.cpp
kpr0th: Corrected for next commit.

ui/src/virtualconsole/vccuelist.cpp
kpr0th: Corrected for next commit.

qmlui/virtualconsole/vccuelist.cpp
kpr0th: Yep -- but the result of calling qFloor is an int, and the resulting stepSize was getting truncated.  But per your request below, I changed "100000" to "100000.0" and removed the now-unnecessary casts to float.

ui/src/virtualconsole/vccuelist.cpp
kpr0th: Same fix as above.

ui/src/virtualconsole/vccuelist.cpp
kpr0th: Tested successfully without the (float). Removed in new commit.

ui/src/virtualconsole/vccuelist.cpp
kpr0th: Tested this as per your recommendation and it works fine that way.  Casts removed.

mcallegari:
Looks good now. Thanks :+1:

PR #1398, opened 2023-02-21
Save side-fader-input to qxw files

kpr0th:
Added missing saveXMLInputControl call for Side Fader external input configuration


mcallegari:
How weird! It was loaded but not saved :sweat_smile:
Thanks for this 👍

PR #1399, opened 2023-02-23
Gh action fix

hjtappe:
This patch fixes the window rendering tests.
- Move tests before install targets
- Enforce QT_QPA_PLATFORM=minimal
- Let the tests evaluate the extra environment variable

Note that this is just one way to make it work. Alternatively, there could be a solution in finding and installing some missing library or library version that causes QT to being unable to render the screens during tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,8 +10,8 @@ jobs:
     strategy:
       #fail-fast: false
       matrix:
-        #task: [compile-qt5, compile-qt5qml,coverage-qt5]
-        task: [compile-qt5qml]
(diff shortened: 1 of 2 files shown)


hjtappe:
As I see in your `winbuild` branch, the fix is similar to the  `--platform offscreen` option. You could try that instead of `minimal`, but I wonder if the test would still do something useful then.
You could try to somehow merge that inside the `GabrielBB/xvfb-action@v1` action or embed a part of that in the `unittest.sh`.

mcallegari:
Great!! Thanks a lot for this.
I'm going to port these changes to the winbuild branch and continue to pursue windows builds too

PR #1400, opened 2023-02-26
Add support for AKAI APC Mini mk2

michaelmertens:
Color feedback values can be found here:
https://www.qlcplus.org/forum/viewtopic.php?f=30&t=16203

Some things to keep in mind:
* The output MIDI channel must be set to 7 to get 100% brightness (when MIDI channel 1 is chosen, you'll only get 10% brightness)
* The control buttons on the side however (Volume, Pan, etc.) have single LEDs which unfortunately only react to messages on MIDI channel 1. Maybe we could add a FixedMidiChannel attribute on the Feedbacks element that could be configured to overwrite the output MIDI channel used for feedback to allow 1 profile can output to multiple MIDI channels?

### DIFF
--- a/resources/inputprofiles/Akai-APCMini-mk2.qxi
+++ b/resources/inputprofiles/Akai-APCMini-mk2.qxi
@@ -1,0 +1,453 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!DOCTYPE InputProfile>
+<InputProfile xmlns="http://www.qlcplus.org/InputProfile">
+ <Creator>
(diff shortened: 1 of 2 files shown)


mcallegari:
Thanks :+1: 
I think a note should be added to the MIDI plugin docs rather than forcing a channel

lumidmx:
AKAI APC Mini mk2 is woking with lumidmx but it does not work leds
i found chataigne to made led work but not the same time with lumidmx
Can u help me?
Best regrads 
Filip Gabriel  (filipgabriel@yahoo.com)

PR #1301, opened 2021-11-30
Initial commit of checkers.js

sandinak:
- can set alternating colors for the squares
- uses a defined width and height that can be manipulated
- can slide them based on the step interval
- can alternate the colors

This has been tested in both the testing tool and QLC+ 4.12.4

PR #1302, opened 2021-12-01
Initial commit of flames.js

sandinak:
- based off simple algorithm for creating flames
- has hilights in differnet colors
- can set amount and depth

PR #1404, opened 2023-03-14
rgbscript: marquee

sandinak:
- adds a marquee function
- has moving "lights" with a count of lights+spaces
- can select the color of the lights
- has fadein depth from outside
- handles corners cleanly so that the lights look even
- tested different sizes on the devtool


sandinak:
So .. codacy is telling me that using 'var ' on a var is bad form.. and I should be using 'let' .. which makse sense .. however; our code interpreter in qlcplus doth not like 'let ' and in fact throws really.. really annoying error:
`
"marquee.js" Error at line: 99 , column: 7 : "Expected `;', `;'"
`
Which tool me going to dinner and complaining to my duck before i figured out what had changed :) Hope this passes all the rest as I think it's a nice script that we're using on this years show.

mcallegari (approved):

mcallegari:
Hey can we please close #1301 and #1302 before throwing more stuff on the table? Thanks

sandinak:
Closed the other two .. one was a dupe .. one needs re-write .. this one is ready; but I can't merge because of Coacy.

mcallegari:
Thanks 👍

PR #1414, opened 2023-03-29
Alternate halfs

hjtappe:
Integrate feature request #1361 into alternate.js: Split the matrix into halfs and by any other divisor.

I would really like to extend the script further, to obey to the colors set in the background (through the virtual console matrix widget) instead of setting them as an array, but we would need to extend the rgbMatrix interface to allow two (or more) colors to be pushed into the matrix script (as an array). Just pushing one color fails in the test script as it does not expect another color to be set.

### DIFF
--- a/resources/rgbscripts/alternate.js
+++ b/resources/rgbscripts/alternate.js
@@ -72,6 +72,9 @@ var testAlgo;
 
   var x = 0;
   var y = 0;
+  // current RGB assigned to the step
+//  algo.currentRgb = 0;
+//  algo.rgbStepColorName = "Matrix RGB";
 
   algo.acceptColors = 0;
   algo.properties = new Array();
@@ -86,7 +89,9 @@ var testAlgo;
 
   algo.color1Index = algo.getColorIndex("Red");
   algo.properties.push("name:color1Index|type:list|display:Color 1|" +
-    "values:" + colorPalette.names.toString() + "|" +
+    "values:" + colorPalette.names.toString() +
+//    ",Matrix RGB" +
+    "|" +
     "write:setColor1Index|read:getColor1Name");
   algo.color2Index = algo.getColorIndex("Green");
   algo.properties.push("name:color2Index|type:list|display:Color 2|" +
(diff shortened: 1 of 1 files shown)


resources/rgbscripts/alternate.js
```diff
@@ -86,7 +89,9 @@ var testAlgo;
 
   algo.color1Index = algo.getColorIndex("Red");
   algo.properties.push("name:color1Index|type:list|display:Color 1|" +
-    "values:" + colorPalette.names.toString() + "|" +
+    "values:" + colorPalette.names.toString() +
+//    ",Matrix RGB" +
```
mcallegari, line 93: this doesn't really seem like a needed change, is it?

resources/rgbscripts/alternate.js
```diff
@@ -59,7 +59,7 @@ var testAlgo;
   colorPalette.makeSubArray = function(_index) {
     var _array = new Array();
     for (var i = 0; i < colorPalette.collection.length; i++) {
-      _array.push(colorPalette.collection[parseInt(i)][parseInt(_index)]);
+      _array.push(colorPalette.collection[parseInt(i, 10)][parseInt(_index, 10)]);
```
mcallegari, line 62: Is radix mandatory?
According to W3C docs it's not
https://www.w3schools.com/jsref/jsref_parseint.asp
However this is ECMA...

resources/rgbscripts/alternate.js
```diff
@@ -72,6 +72,9 @@ var testAlgo;
 
   var x = 0;
   var y = 0;
+  // current RGB assigned to the step
```
mcallegari, line 75: Remove this?

mcallegari (changes requested):
Looks good but needs a bit of cleanup
Lots of commented code left there

Also, a screenshot of the result would be nice.
Thanks

resources/rgbscripts/alternate.js
hjtappe: ... Codacy thinks, it's unsafe not to fix this. The parseint function would otherwise autodetect the basis for the conversion. So it is not necessary, but a bit safer.

resources/rgbscripts/alternate.js
hjtappe: Yes, it is a leftover. Not sure why, I should have seen these things in my final review before opening the pull request. Thanks for double-checking my code. I'll clean that up.

resources/rgbscripts/alternate.js
hjtappe: Yes, this is leftover and should not be there. I'll remove it.

hjtappe:
Here is a screenshot of the "Split" mode:
image
It can be configured to split in 2, 3, 4 etc. segments, horizontal, vertical and interleaved (chess board) and will calculate the sizes of each chunk depending on the total width and height.

mcallegari:
Thanks :+1:

PR #1415, opened 2023-03-30
fixed small bugs in web interface

arneBersch:
I tried to fix two bugs in the web interface code.

### DIFF
--- a/webaccess/res/keypad.html
+++ b/webaccess/res/keypad.html
@@ -97,6 +97,7 @@ function connect()
   };
 
   websocket.onclose = function () {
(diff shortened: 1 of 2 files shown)


arneBersch:
- In keypad.html, the variable isConnected is set to true after connection, but when closing the connection, isConnected was still set to true enabling sending commands when the connection is closed.
- In simpledesk.js, after resetting an universe, the current page was set to 1, but the faders weren't changed and the page number wasn't changed either.

mcallegari:
Makes sense. Thanks for fixing :+1:

Issue #1408, opened 2023-03-21
Update build system to CMake


yestalgia:
[cmake ](https://cmake.org/)is recommended for Qt 6 and QLC+ hasn't switched to it yet. 

There's a pretty bad PR that was supposed to evolve but got closed which included porting to cmake:

- https://github.com/mcallegari/qlcplus/pull/1214

It could take quite a bit of work but the following resources should help:

- https://wiki.qt.io/CMake_Port/Porting_Guide
- https://pypi.org/project/qmake2cmake/

The TODO list is the following:

- [x] Qt5 v4 build on Linux
- [x] Qt5 v4 test units on Linux (make check)
- [x] Qt5 v4 deploy on Linux (.deb package)
- [x] Qt5 v5 build on Linux
- [ ] Qt6 v4 build on Linux
- [x] Qt5 v5 deploy on Linux (appimage package)
- [x] Qt5 v4 build on Windows
- [ ] Qt6 v4 build on Windows
- [x] Qt5 v4 deploy on Windows (.exe bundle)
- [x] Qt5 v5 build on Windows
- [x] Qt5 v5 deploy on Windows (.exe bundle)
- [ ] Qt5 v4 build on macOS
- [ ] Qt6 v4 build on macOS
- [ ] Qt5 v4 deploy on macOS (.dmg bundle)
- [ ] Qt5 v5 build on macOS
- [ ] Qt5 v5 deploy on macOS (.dmg bundle)
- [x] Coverage build

yestalgia:
@Itay-Lifshitz
Please record your progress here.

Itay-Lifshitz:
@mcallegari 
I compiled the GUI program of QLC++4 using Qt5 + MSVC 2017 + CMake.
But when I tried to compile test projects, I got some troubles.
For instance, in the ui/test/aboutbox/aboutbox_test.cpp file, we have the following code block.
`
#define private public
#include "aboutbox.h"
#undef protected
`
And then it accesses the original private function of AboutBox class like the following code.
`
void AboutBox_Test::slotTimeout()
{
    AboutBox ab(NULL);
    ab.m_timer->stop();

    QCOMPARE(ab.m_row, -1);
    QCOMPARE(ab.m_increment, 1);

    for (int i = 0; i <= CONTRIBCOUNT; i++)
    {
        ab.slotTimeout();
        QCOMPARE(ab.m_row, i);
        QCOMPARE(ab.m_increment, 1);
    }

[...]

PR #1418, opened 2023-04-06
Cmake build

Itay-Lifshitz:
#1408 
I reopened this PR for CMake build system with cleaner work history.
@mcallegari 
Please review it and give me feedback.


mcallegari:
Hi, how do I build on Linux using the official Qt versions and not the Ubuntu ones?

hjtappe:
> Hi, how do I build on Linux using the official Qt versions and not the Ubuntu ones?

Get the installer here: https://www.qt.io/download-qt-installer-oss
It requires a free account with them.

Run the installer and install the version used in the official build (currently 5.14.2):
https://www.qt.io/download-qt-installer-oss

mcallegari:
> > Hi, how do I build on Linux using the official Qt versions and not the Ubuntu ones?
> 
> Get the installer here: https://www.qt.io/download-qt-installer-oss It requires a free account with them.
> 
> Run the installer and install the version used in the official build (currently 5.14.2): https://www.qt.io/download-qt-installer-oss

Ehm...thanks but being the maintainer of a Qt based project, I know how to get Qt and how to build QLC+ the way it built in the last 10 years or so :)
My question was related to this PR, so how to build **with cmake** but with an official Qt installation and not with Debian packages. This allows me to switch Qt version easily, as I always did to maintain code compatibility and resolve build issues with different compilers.

Itay-Lifshitz:
@mcallegari 
Did you check the qmake2cmake.md file in my branch?
I was wondering if there is any difference between using official QT and Ubuntu QT?

mcallegari:
> @mcallegari Did you check the qmake2cmake.md file in my branch?
Yes I did, and there is mention only of Ubuntu packages

> I was wondering if there is any difference between using official QT and Ubuntu QT?
There is indeed a difference! Ubuntu packages are built by Ubuntu and delivered as individual packages.
Official Qt versions are delivered via Qt online installers and those binaries are built by the Qt company. That's why I call them "official"

With qmake it's super easy to build with Qt's bundles. You just type qmake with absolute path, like:
`/home/user/Qt/6.4.1/gcc_64/bin/qmake`
and that's it. So you can test different Qt versions very easily.
How do I do this with cmake?

Also, we went from
```
qmake
make
```
to
```
mkdir build
cd build
cmake -Dudev=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH="/usr/lib/x86_64-linux-gnu/cmake/Qt5" ..
make
```
This needs to be simplified for newbies (in this case like me)
To be honest I prefer default settings in CMakeLists.txt (even better with autodetection) and eventually change manually what's needed.
The goal is to achieve
```
cmake ..
make
```

Itay-Lifshitz:
@mcallegari 
I updated the CMake file as well as the documentation.
Please check it and give me feedback.

plugins/ola/CMakeLists.txt
```diff
@@ -0,0 +1,79 @@
...
+    Qt${QT_MAJOR_VERSION}::Widgets
+)
+
+if(MACOS)
+    set_target_properties(${module_name} PROPERTIES
+        MACOSX_BUNDLE FALSE
+    )
+endif()
+
+if((MACOS) AND (NOT (packagesExist(libolalibolaserver))))
```
mcallegari, line 51: This should be `libolaserver`

mcallegari:
thanks @Itay-Lifshitz 
I went much further and found some more issues. See my comment about libolaserver package name.
Also, OLA libraries should be an optional dependency. If not found, the OLA plugin should not be built
I also noticed it checks for 'mad' twice and for portaudio, which is not a QLC+ dependency:

```
-- Checking for module 'mad'
--   Found mad, version 0.15.0b
-- Checking for module 'mad'
--   Found mad, version 0.15.0b
-- Checking for module 'sndfile'
--   Found sndfile, version 1.0.31
-- Checking for module 'portaudio-2.0'
--   No package 'portaudio-2.0' found
-- Checking for module 'fftw3'
--   Found fftw3, version 3.3.8
-- Checking for modules 'libftdi1;libusb-1.0'
--   Found libftdi1, version 1.5
--   Found libusb-1.0, version 1.0.25
-- Checking for modules 'libftdi;libusb'
--   Found libftdi, version 0.20
--   Found libusb, version 0.1.12
Building with libFTDI1 support. Version: 
Using v1.5+ buffer flush API
-- Checking for module 'libusb-1.0'
--   Found libusb-1.0, version 1.0.25
-- Checking for module 'libola'
--   Found libola, version 0.10.8
-- Checking for module 'libolaserver'
--   Found libolaserver, version 0.10.8
CMake Error at plugins/ola/CMakeLists.txt:51 (if):
  if given arguments:

    "(" "MACOS" ")" "AND" "(" "NOT" "(" "packagesExist" "(" "libolalibolaserver" ")" ")" ")"

  Unknown arguments specified
```

Itay-Lifshitz:
Ok, will check it and fix it soon.

Itay-Lifshitz:
@mcallegari 
Hi, which OS did you use for building project with CMake?
And the 'mad' library is actually used in `engine/audio/plugins/plugins.pro` and we need to check if it exists.
`
!android:!ios {
system(pkg-config --exists mad) {
  SUBDIRS += mad
}
`

Itay-Lifshitz:
@mcallegari 
I made some updates, I actually didn't test with Ola library.
Please review it and give me feedback.

mcallegari:
> @mcallegari Hi, which OS did you use for building project with CMake? And the 'mad' library is actually used in `engine/audio/plugins/plugins.pro` and we need to check if it exists. `!android:!ios { system(pkg-config --exists mad) { SUBDIRS += mad }`

I'm still testing on Linux before validating the PR on other OSes. It's easier for me since I work with Linux every day.
As for 'mad', please read my comment again:
- mad is checked twice: why not once?
- portaudio is not a QLC+ dependency; so why checking for it?

Itay-Lifshitz:
@mcallegari 
In my last update, mad is checked only once, and portaudio is not checked.

mcallegari:
@Itay-Lifshitz 
I'm still on Linux. I did `make install` and found many more issues:
- installation is performed to /usr/local/bin: wrong. Please make it as it was with qmake
- tests are installed: wrong
- target `make check` doesn't exist. This would prevent CI to check a good build
- .a files are installed in /usr/local/lib: they shouldn't
- .so files are not created!!! This is very wrong on Linux

To be honest from this PR I expected you checked that the output result was the same of the qmake build, but it seems this is not the case. Please fix

mcallegari:
@Itay-Lifshitz I tried to build again with the latest changes.
The situation has improved but there are still the following:
- mad and sndfile plugins are still built as .a and not .so (SHARED)
- most of the plugins are built as .a instead of .so
- the final qlcplus binary cannot be executed if the build has been done with a Qt version different from the system version. The error is:
`qlcplus: symbol lookup error: qlcplus: undefined symbol: _ZdlPvm, version Qt_5`
I built it like this:
`cmake -DCMAKE_PREFIX_PATH="/home/massimo/projects/Qt5.14.2/5.14.2/gcc_64/lib/cmake/" ..`
So binaries should be linked to libraries in that "custom" folder.
Instead they link to system Qt libraries.

Itay-Lifshitz:
@mcallegari 

- Sorry, I didn't know Qt would export library as a shared object file in Linux by default without specifying any parameters in .pro file. I thought it would export as a static library by default. Let me make a fix on it.

- In terms of the runtime error with undefined symbol, I wrote some comments in the "Running and installing qlcplus" section in the `qmake2cmake.md` file.
As short, the qlcplus binary file actually doesn't link to the system Qt library. It just remembers the name of the Qt library and try to find them in the LD_LIBRARY_PATH, which is system path by default.
One more question here, Is the original binary file linked in static mode or dynamic mode with QT libraries?
If we link qlcplus in static mode, we will not suffer from such link errors.
But if we link it in dynamic mode, which is using Qt library .so files, we will need to specify which so files it should use.

mcallegari:
@Itay-Lifshitz see my comments below. Thanks for your efforts
> @mcallegari
> 
> * Sorry, I didn't know Qt would export library as a shared object file in Linux by default without specifying any parameters in .pro file. I thought it would export as a static library by default. Let me make a fix on it.

Actually it is specified in the pro file. Example:
https://github.com/mcallegari/qlcplus/blob/master/plugins/E1.31/E1.31.pro#L3

> * In terms of the runtime error with undefined symbol, I wrote some comments in the "Running and installing qlcplus" section in the `qmake2cmake.md` file.
>   As short, the qlcplus binary file actually doesn't link to the system Qt library. It just remembers the name of the Qt library and try to find them in the LD_LIBRARY_PATH, which is system path by default.
>   One more question here, Is the original binary file linked in static mode or dynamic mode with QT libraries?
>   If we link qlcplus in static mode, we will not suffer from such link errors.
>   But if we link it in dynamic mode, which is using Qt library .so files, we will need to specify which so files it should use.

Static linking is not considered in this project. Every build is dynamically linked to the Qt libs.
Eventually some build flavours might include the specified Qt version in a bundle like on macOS or in a AppImage package.

> * Next question, when you build qlcplus with qmake, did you specify the path of qmake like "/home/massimo/projects/Qt5.14.2/5.14.2/gcc_64/bin/qmake" or just use qmake in system path? If you used qmake in system path, I guess it will use debian Qt packages instead of the official Qt packages while compiling and running.

If I want to use a specific Qt version, I specify the absolute path of qmake.
When I release the official Debian package (.deb) instead, the system Qt version is used.

mcallegari:
hello @Itay-Lifshitz I have just pushed cmake files in a dedicated branch (cmakebuild)
I took the chance to fix some more plugins still building as static instead of shared objects.
I haven't pushed all the changes you made to the C++ code. Can you please explain why those were needed? Which compiler did you use that found so many errors in the code?

Right now the main issue in the cmakebuild branch is that if I use a specific Qt version to build, the final binaries will link to my system Qt version instead. Therefore the application binary will not run.
Can you please help to resolve that? Thanks

Itay-Lifshitz:
Hello, @mcallegari
- I thought I built all of the plugins as shared object files. Which plugins did I miss?
- I modified the source codes, because of compile errors in MSVC(Microsoft Visual Studio Compiler). I believe that I kept the original logic, and you can review the changes I made it if you want.
- I will resolve the linkage issue soon.
Thank you.

Itay-Lifshitz:
@mcallegari 
I believe linkage issue has been resolved.
Please check it.

mcallegari:
hi @Itay-Lifshitz here's the list of the .a files that I found (and fixed)
```
./engine/audio/src/libqlcplusaudio.a
./plugins/loopback/src/libloopback.a
./plugins/ola/libolaio.a
./plugins/enttecwing/src/libenttecwing.a
./hotplugmonitor/src/libhotplugmonitor.a
```
the 2 plugins were wrong. The other two are correct since they're included in the engine final shared library.
Can you please move and rebase this PR on the [cmakebuild](https://github.com/mcallegari/qlcplus/tree/cmakebuild) branch instead of master?
I think it makes sense to start merging changes step by step

Itay-Lifshitz:
OK, closing this PR and moving to cmakebuild.

mcallegari:
Hi, I manually applied the change to the main cmakelists.txt file and now linking works as expected.
QLC+ starts but doesn't find runtime resources like plugins. It says
`void IOPluginCache::load(const QDir&) "lib/x86_64-linux-gnu/qt5/plugins/qlcplus"`
Which is a relative path, while plugins are installed in
`/usr/lib/x86_64-linux-gnu/qt5/plugins/qlcplus/`

This is because the generated qlcconfig.h has relative paths.
Can you please check?

Itay-Lifshitz:
@mcallegari 
- Please check my last commit 87fba151, the qlcconfig.h should be different for different OS platform.
- Sorry, but I don't know how to split the PR. Should I close this PR and reopen 2 new PRs? I will appreciate it if you can guide me.
- Can you explain the missing linking issue for plugins in more detail? What external dependencies are not linked?

mcallegari:
@Itay-Lifshitz please rebase to the cmakebuild branch
Please follow the commits I do otherwise this PR will always be conflicting.
Please split c++ code changes to another PR.

Basic Linux build is now working. I haven't tried qmlui, appimage, test units. debian, rpm. etc
I have tried to build on Windows and it doesn't work at all.
I think you tested only with Visual Studio, which at the moment is not a supported build environment.

We need cmake to work on MSYS2. Please see the wiki how to build and fix the project configuration to build with cmake+gcc.

Thanks

mcallegari:
Build log attached
[msys2_build_fail.txt](https://github.com/mcallegari/qlcplus/files/11925511/msys2_build_fail.txt)

Itay-Lifshitz:
OK, closing this PR and opening the new PR.

Itay-Lifshitz:
@mcallegari 
Btw, how should I open 2 PRs separating code changes and CMake files?

mcallegari:
> @mcallegari Btw, how should I open 2 PRs separating code changes and CMake files? Should I create 2 branches on my side?

I think so

PR #1421, opened 2023-04-17
Wizard naming

netmindz:
The current naming convention for the functions wizard places the name of the fixture at the end. This makes it hard to find all the scences that relate to a certain fixture
Screenshot from 2023-04-17 16-46-18

I find it much better to see all the scenes for the same fixure grouped together like this
Screenshot from 2023-04-17 16-43-03

https://qlcplus.org/forum/viewtopic.php?t=16337

### DIFF
--- a/ui/src/palettegenerator.cpp
+++ b/ui/src/palettegenerator.cpp
@@ -248,12 +248,12 @@ void PaletteGenerator::createColorScene(QList<SceneValue> chMap, QString name, P
             even = !even;
         }
     }
-    scene->setName(m_model + " - color - " + name);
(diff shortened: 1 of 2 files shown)


netmindz:
Just found an issue, the name widget in the solo frame has changed to match the scene name, not the shorter form. Not sure why

yestalgia:
This would save me so much time!

netmindz:
The only downside I spotted while testing @mcallegari was that when you see the scene name in the cluelist, you can't really see what is what unless you scroll horizontally, but if you do so, then the scroll resets back on the next cue. Not a huge issue as generally you don't need to know the name of the scene at all time

PR #1423, opened 2023-04-27
Fix off-by-one errors and integer interpretation errors.

hjtappe:
With this pull request, I would like to correct some problems in marquee.js:
- Changing / Fading colors (color 1 / rgb) were not taken over into the background feature
- The integer in properties "16" + "3" + 1 were interpreted as string (1631) - added parseInt to ensure it is handled as a number
- The edges of the marquee were set twice, each edge had it's own way to start over. This leads to each edge having a potentially inconsistent marquee pattern (see screenshot).
- The selected marquee color is not applied but added (merged) with the background color. This significantly reduces the reachable color space for the marquee. It could be made an option (@sandinak, what's your opinion on this?), but if the user needs merged colors on the marquee, he could select the marquee merged color right away.

This is how it was looking before this pull request:
Screenshot before

This is how it looks after this pull request:
Screenshot after

### DIFF
--- a/resources/rgbscripts/marquee.js
+++ b/resources/rgbscripts/marquee.js
@@ -77,7 +77,7 @@ var testAlgo;
   algo.properties = new Array();
   algo.edgeDepth = 2;
   algo.properties.push(
-    "name:depth|type:range|display:Depth|values:1,10|write:setDepth|read:getDepth"
+    "name:depth|type:range|display:Depth|values:1,10000|write:setDepth|read:getDepth"
   );
   algo.marquee = 0;
   algo.properties.push(
@@ -85,7 +85,7 @@ var testAlgo;
   );
   algo.marqueeCount = 3;
   algo.properties.push(
-    "name:marqueeCount|type:range|display:Marquee Spaces|values:1,10|write:setMarqueeCount|read:getMarqueeCount"
+    "name:marqueeCount|type:range|display:Marquee Spaces|values:1,100|write:setMarqueeCount|read:getMarqueeCount"
   );
   algo.marqueeColorIndex = 0;
   algo.properties.push(
--- a/resources/rgbscripts/marquee.js
+++ b/resources/rgbscripts/marquee.js
@@ -195,10 +195,11 @@ var testAlgo;
     // only if the dimensions have changes (not on color change)
     if (util.width != width || util.height != height || util.initialized !== true) {
       var length = height * 2 + width * 2 - 4;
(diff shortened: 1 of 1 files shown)


sandinak (approved):
Thanks for the fixes.. much appreciated!

hjtappe:
Yes, if the amount of colors does not devide evenly, then there is a point in which you see the seaming point between start and end of the pattern.

One way around this would be to not provide the distance between the marque bulbs but the number of bulbs to be distributed along the edges. That could be handled much more evenly by the script through Math.round().

What we still could do is:
- Take the wanted distance between marquee bulbs: 
- Calculate the **amount** of points to be spread along the edges: `Math.floor(length / (distance + 1))`
- Calculate the **mediumDistance** between the points: `length / amount`
- Use the **mediumDistance** to detect if a pixel should be highlighted (instead of the raw marqueeCount): `p % mediumDistance < 1`

Something like that, but I was not sure what the intention was and how normally marquee in the wild would handle that (probably by letting the user choose only distances which devide the length to a full number).
I can take care, but I am most probably on radio silence for the next two weeks.

hjtappe:
I  was able to find the brief time for distributing the required spacers into the marquee so the seaming spot is no more as obvious. @sandinak, please briefly check if this is what you were looking for and if I might have missed a matrix size for which it might not work.

yestalgia:
Thanks for all your work on this @hjtappe. This looks like a really nice matrix effect.

hjtappe:
> Thanks for all your work on this @hjtappe. This looks like a really nice matrix effect.

Kodos to @sandinak for introducing it. :-)

mcallegari:
Thanks @sandinak and @hjtappe for the efforts
Please just let me know when this is ready to be merged

hjtappe:
After the background fix today, I think it is ready to go. @sandinak can you please check again?

hjtappe:
While I think let's increase the range limits. We often had feedback that arbitrary small limits cause unnecessary problems on users with larger Matrices.

mcallegari:
> While I think let's increase the range limits. We often had feedback that arbitrary small limits cause unnecessary problems on users with larger Matrices.

Well, keep in mind that for those knowing what JavaScript is, it will be super easy to edit the script and increase the limits if needed. I wouldn't worry too much about magic numbers right now.

If you're both happy, can I merge this?

hjtappe:
Yes, Thank you.

mcallegari:
👍

sandinak:
Thanks for all the hard work on this!

PR #1424, opened 2023-04-27
Introduce error evaluation on the QJSEngine::call calls

hjtappe:
Introduce error evaluation on the QJSEngine::call calls, to detect dynamic script errors during code interpretation (::call --> undefined indices etc.).

The error detection during evaluate() only detects static error classes while loading the scripts (::evaluate --> parse errors, missing brackets etc.).


mcallegari:
Nice addition, thanks 👍

PR #1425, opened 2023-05-03
webaccess: improve vcbutton overflow and alignment

pomowunk:
Some improvements to button widgets in the web view:

- Hide the overflow of text in vcbuttons when the button name does not fit.
- Center the text vertically and horizontally to match the layout in Qt.
- Include the borders in the button width/height to avoid shifting the layout and hiding parts of other widgets.

(Tested in Firefox and Chromium, should work anywhere.)

Attached below are screenshots of some test widgets in the Virtual Console and in Webaccess.

qlcplus-webaccess-vcbutton-pullrequest

### DIFF
--- a/webaccess/res/virtualconsole.css
+++ b/webaccess/res/virtualconsole.css
@@ -11,13 +11,16 @@ form {
 }
 
 .vcbutton {
- display: table-cell;
+ display: flex;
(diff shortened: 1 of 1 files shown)


mcallegari:
Thanks 👍

PR #1434, opened 2023-07-08
Add RGB Script: Propellor by Christoph Klenge

yestalgia:
This PR provides an optimized version of the "Propellor" script which creates a visual effect of a line rotating in 2D space, similar to a propellor. The optimization makes the script more efficient and performant, without altering its original visual output.

The initial concept for this script was discussed on the @qlcplus Instagram account and the effect it generates draws inspiration from similar effects created using GrandMA2. The visual reference is provided below.

The initial version of the script was shared on the QLC+ Forum in [this thread](https://www.qlcplus.org/forum/viewtopic.php?p=68728&hilit=new+instagram#p68728).

This optimized version of the script has undergone thorough testing using the devtool to ensure its accuracy and efficiency. 

Original script created by Christoph Klenge. (Forum user Xeon, instagram: @bodgedbutworks)

Any feedback or further testing would be greatly appreciated!

Original script visual reference:
image


hjtappe:
Hi,
is this something that is reflected already in circular.js or could be integrated therein?

grafik

yestalgia:
> * Add a threshold parameter to switch the borders rather than fade them
> * Add a width parameter to set the width of the bar.

Thanks for the feedback.

I like the idea of combining this effect into circular.js. Frankly I didn't know about the range of effects available in that script.

I'm not sure what you mean about adding a width parameter. Wouldn't it be better to add it as a different mode?

hjtappe:
I was referring to the width parameter I saw in the new script. That parameter would need to be added for all modes in circular.js.

Yes, the new behavior could be a new mode then.

yestalgia:
I think that's the way to go. It looks like you've already made a start, can you please implement it? If so I can close this PR.

Basically we need: 

1. Mode: Propeller
2. Fill Matrix option
3. Number of blades???

hjtappe:
I can give it a try to merge the function into circular.je. Give me some time, I'll send you a ping for review then.

mcallegari:
Superseeded by #1438

PR #1435, opened 2023-07-12
CMake fix on Windows

Itay-Lifshitz:
Based on your current cmakebuild branch, added the following modification.
- Import .rc file on Windows so that exe files can display correct icons
- Translation installation
- Not install .a files
- Fixed bugs in resources installation like `meshes`.

Some tips for easier review.
- In CMake, once `pkg_check_modules` is called for the specific module e.g. `pkg_check_modules(PORTAUDIO_2 IMPORTED_TARGET portaudio-2.0)`, the variables like `PORTAUDIO_2_FOUND` and `PORTAUDIO_2_LIBRARIES` are set globally, so we don't have to and are not recommended to call `pkg_check_modules` in several places.
- I moved checking of 'portaudio-2.0' module into the root CMakeList.txt because they were used in 3 places.
- For similar reason, I also moved checking of `fftw3` module into engine/CMakeList.txt
- I also moved checking of `libusb-1.0` module into plguins/CMakeList.txt
- Maybe we can keep checking if QT version is less than 5 e.g. `if (QT_VERSION_MAJOR LESS 5)` in CMake even though it is no longer supported, because we keep similar blocks in QT .pro files. e.g. `lessThan(QT_MAJOR_VERSION, 5) {}`. I was trying to keep and not to destroy the original logic in QT .pro files.
- If you see the changes like the following, it is for not installing .a files on Windows.
image
- qlcconfig.h.win32.in is actually different from unix one. I don't know why they should be different, but they were different in QT .pro file as follows.
image

If you think this PR is good, please merge it directly, not manually.


Itay-Lifshitz:
You may also want to test qlcplus v5 (QML) on Windows. I tested it on my side and it seemed to work well.

mcallegari (changes requested):
I reviewed all the files.
Again, there's a mix of different topics, not just Windows deploy.
Please revert dmxusb changes. I already tested the build on Windows and I don't understand why you had to make other modifications to it.
I don't agree also on different versions of qlcconfig files.
There should be only one, properly created by variables.cmake. 
On Linux and Windows the upstream version is OK. When I'll deal with macOS and other OSes I will adjust the variables where needed.

Itay-Lifshitz:
> Merged, thanks

OK, great. What is the next?

mcallegari:
@Itay-Lifshitz I'm testing right now and I'll update #1408 shortly.
Please restore the "make check" target.
It's only for Linux and right now it says:
```
19:21:49:massimo:build > make check
/bin/sh: 1: ./unittest_cmake.sh: not found
make[3]: *** [CMakeFiles/unittests.dir/build.make:70: CMakeFiles/unittests] Error 127
make[2]: *** [CMakeFiles/Makefile2:2337: CMakeFiles/unittests.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2370: CMakeFiles/check.dir/rule] Error 2
make: *** [Makefile:182: check] Error 2
```
this is required to create Debian and AppImage packages

PR #1437, opened 2023-07-21
Added missing unittest_cmake.sh

Itay-Lifshitz:
#1408 
@mcallegari 
- I am not so clear how `engine/test/qlci18n` works.
Because `qlci18n_fi_FI.ts_` is required for testing `qlci18n` and `qlci18n_fi_FI.ts` is required for compiling the target and now both are omitted from Git.
Tested on my side and it looks like working when I manually added these 2 files.
- It looks like `make check` for Windows is missing in the todo list. Don't we need it anymore? It looks like that it was there in the original "qlc.pro" file.
- I manually modified the 1st line of `resources/fixtures/scripts/fixtures-tool.py` from `#!/usr/bin/env python` to `#!/usr/bin/env python3`, otherwise it did not work. Not sure if python3 is default in your develop environment.

### DIFF
--- a/unittest_cmake.sh
+++ b/unittest_cmake.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# Define the source and destination directories
+SOURCE_DIR="."
+DEST_DIR="./build"
+if [ -d "$2" ]; then
+  DEST_DIR="$2"
+fi
+
+echo "Using the destination directory $DEST_DIR"
+
+if [ ! -d $DEST_DIR/resources/ ]; then
+  mkdir -p $DEST_DIR/resources/
+fi
+
+# Copy resources directories necessary for unittest
+cp -r $SOURCE_DIR/resources/colorfilters $DEST_DIR/resources
+cp -r $SOURCE_DIR/resources/fixtures $DEST_DIR/resources
+cp -r $SOURCE_DIR/resources/gobos $DEST_DIR/resources
+cp -r $SOURCE_DIR/resources/icons $DEST_DIR/resources
+cp -r $SOURCE_DIR/resources/inputprofiles $DEST_DIR/resources
+cp -r $SOURCE_DIR/resources/rgbscripts $DEST_DIR/resources
+cp -r $SOURCE_DIR/resources/schemas $DEST_DIR/resources
+
+# Find all files necessary for tests recursively in the source directory and copy to destination directory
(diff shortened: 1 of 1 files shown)


mcallegari:
@Itay-Lifshitz there's never been test units on Windows. They're executed only on Linux which is the reference OS (that's why I insisted to have a build working there first)
As for i18n I'll check later what was the intention of that test.
Thanks

Itay-Lifshitz:
@mcallegari You can review this PR anyway, I added the missing shell script file.

mcallegari:
Looks good, thanks
Next is creating Debian package
See: https://karthikkalyanaraman.medium.com/creating-debian-packages-cmake-e519a0186e87
Seems like adding the following and using `cpack` will do the trick
```
SET(CPACK_GENERATOR "DEB")
SET(CPACK_DEBIAN_PACKAGE_MAINTAINER "KK") #required
INCLUDE(CPack)
```

Itay-Lifshitz:
@mcallegari 
- I was wondering how the original `create-deb.sh` worked.
It simply runs `dpkg-buildpackage -rfakeroot -I.git --jobs=auto` and then it compiles, tests, builds packages and signs automatically.
Where are these operations defined?
- And I also got the following error when I tried to build debian package using the old `create-deb.sh`. How can I solve it? 
image

mcallegari:
I think `dpkg-buildpackage` recognized it was a qmake project and built it that way.
As I mentioned in my previous comment, with cmake we need to use `cpack`.
I'm gonna give it a try when I have some time

Itay-Lifshitz:
> I think `dpkg-buildpackage` recognized it was a qmake project and built it that way. As I mentioned in my previous comment, with cmake we need to use `cpack`. I'm gonna give it a try when I have some time

I think I found it. Please refer to this [link](https://stackoverflow.com/questions/59669276/creating-a-debian-package-from-cmake-project)
It says Ubuntu is referring to `debian` directory when it is building a package.
Working on it now.

mcallegari:
@Itay-Lifshitz please wait, I'm working on it right now.
I'll push some changes soon

Itay-Lifshitz:
> @Itay-Lifshitz please wait, I'm working on it right now. I'll push some changes soon

OK, please let me know the updates when you push them.

mcallegari:
I just pushed. A debian package is created, but I'm not sure if it is a proper one.
I think dependencies need to be fixed.
You can work on AppImage creation if you want

Itay-Lifshitz:
OK, will work on appimage.

PR #1438, opened 2023-07-21
Rgbscript circular propellor

hjtappe:
This pull request merges the propellor functionality as proposed in #1434 into the circular.js script, thus:
- Merging center gap into circular.js modes
- Adding a parameter for line width / weight in circular.js
- Adding possibility for multiple blades for the propellor
- Adding possibility to limit the propellor to a circular area

The direction can be set by the direction (forward / reversed) matrix setting

Please review and check the functionality from propellor.js is integrated correctly in circular.js, @yestalgia

### DIFF
--- a/resources/rgbscripts/circular.js
+++ b/resources/rgbscripts/circular.js
@@ -31,7 +31,7 @@ var testAlgo;
     algo.properties = new Array();
 
     algo.circularMode = 0;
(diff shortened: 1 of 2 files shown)


yestalgia:
Very cool, I'll be taking a look today!

mcallegari:
Just tested, this is really cool!
Thanks for submitting :+1:

PR #1439, opened 2023-07-22
AppImage build using CMake

Itay-Lifshitz:
@mcallegari I already tested on my side. Please review this PR.


mcallegari:
@Itay-Lifshitz thanks but I found this error if I use system Qt libraries:
```
CMake Error at platforms/linux/cmake_install.cmake:102 (file):
  file INSTALL cannot find "/usr/lib/lib/libQt5Core.so.5": No such file or
  directory.
Call Stack (most recent call first):
  platforms/cmake_install.cmake:47 (include)
  cmake_install.cmake:97 (include)
```
Can you please have a look?

mcallegari:
@Itay-Lifshitz  any news on this?
I will test cmake on macOS but I think the current status is already good.
Thanks

Itay-Lifshitz:
> @Itay-Lifshitz any news on this? I will test cmake on macOS but I think the current status is already good, so merging on master could be near. Thanks

Hi, sorry for delay, I have been busy this week.
I will update by this weekend.
And I think current CMake is not ready for MacOS, because I have never tested due to lack of environment.

PR #1442, opened 2023-07-29
Bugfix on AppImage.

Itay-Lifshitz:
#1408 
@mcallegari I was able to make AppImage using both of debian and official QT packages.
It could run with no issue when I built AppImage with the official QT package.
But it showed a white blank screen when I built with the debian QT package.
I checked the original qmake files and it looked like we never built AppImage with the debian QT packages.
So I am opening this PR and hope you can review it.
Thanks.
PS: I made a small modification on Windows deployment, and I believe it has no issue to merge.

### DIFF
--- a/create-appimage-cmake.sh
+++ b/create-appimage-cmake.sh
@@ -1,9 +1,10 @@
 #!/bin/bash
 #
-# Script to create a self contained AppImage
+# Script to create a self contained AppImage using CMake
 # Requires wget and chrpath
-# Export QTDIR before running this, like:
-#   export QTDIR=/home/user/Qt5.9.4/5.9.4/gcc_64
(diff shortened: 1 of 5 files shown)


mcallegari:
Thanks :+1: 
Appimage with official Qt works as expected

Itay-Lifshitz:
> Thanks 👍 Appimage with official Qt works as expected

Great.
What is the next step?
If it is to work on MacOS, it would be great if you could kindly provide MacOS environment via either of SSH or AnyDesk/TeamViewer.
MacOS part in CMake is currently missing because I have no development environment.
Thanks.

mcallegari:
@Itay-Lifshitz next is coverage.
I have a Mac Mini but I cannot give you access.
I'm gonna try to build with cmake in the next days and let you know how it goes.

PR #1446, opened 2023-08-05
NanoDMX/StageProfi/Enttec fix 

brianwiddas:
For any value over 127, the nanodmx code sends the update to the device repeatedly rather than just once, as happens for any value <= 127

This is caused by a signed/unsigned value comparison in nanodmx.cpp

Fixed and tested. Also fixed the same code in stageprofi.cpp and enttecdmxusbpro.cpp, but not tested.

### DIFF
--- a/plugins/dmxusb/src/enttecdmxusbpro.cpp
+++ b/plugins/dmxusb/src/enttecdmxusbpro.cpp
@@ -530,7 +530,7 @@ void EnttecDMXUSBPro::run()
                 {
                     uchar val = uchar(m_outputLines[i].m_universeData[j]);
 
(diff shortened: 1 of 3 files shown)


mcallegari:
Thanks for the patch, makes sense!
However, looking at the code, maybe the right way is to declare 'val' as char, since it comes and goes into QByteArrays.
What do you think?

brianwiddas:
Hi,

Yes, that makes sense. I looked over the code for all three and can't see any reason why it would be a problem.

Updated the code and tested nanodmx

mcallegari:
Thanks 👍

PR #1447, opened 2023-08-14
Fix random order chaser always starts with first step

Wazzledi:
Instead of using the random order of a chaser, the chaserrunner sets the step index to zero when started by a virtual console button. Fixed by using the randomStepIndex method instead of just setting the index.


mcallegari:
Thanks! :+1:

PR #1450, opened 2023-09-03
Updated catalan and spanish translations in fixtureeditor

sbenejam:
Updated catalan and spanish translations in fixtureeditor


mcallegari:
Thanks 👍

PR #1452, opened 2023-09-06
Update German translations

kripton:


ui/src/qlcplus_de_DE.ts
```diff
@@ -4261,157 +4261,157 @@ Der Assistent kennt keinen Unterschied zwischen einem Schalter und einem Regler,
...
         <location filename="rdmmanager.ui" line="168"/>
         <source>Write</source>
-        <translation type="unfinished"></translation>
+        <translation>Schreiben</translation>
     </message>
     <message>
         <location filename="rdmmanager.ui" line="183"/>
         <source>Byte</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
```
FloEdelmann, line 4325: Does an empty `translation` mean that the source string will be used? Otherwise, this string and the following ones are missing.

ui/src/qlcplus_de_DE.ts
```diff
@@ -4569,37 +4569,37 @@ Der Assistent kennt keinen Unterschied zwischen einem Schalter und einem Regler,
...
         <location filename="rgbmatrixeditor.ui" line="165"/>
         <source>White</source>
-        <translation type="unfinished">Weiß</translation>
+        <translation>Weiß</translation>
     </message>
     <message>
         <location filename="rgbmatrixeditor.ui" line="170"/>
         <source>Amber</source>
-        <translation type="unfinished"></translation>
+        <translation></translation>
```
FloEdelmann, line 4587: Again

ui/src/qlcplus_de_DE.ts
mcallegari: `<source>Byte</source>`
That's the string to translate
AFAIK there are no empty translation strings in QLC+

FloEdelmann: Yeah, but should it be `<translation>Byte</translation>` then, or is `<translation></translation>` equivalent?

mcallegari: If translation is empty I think it will fallback to the original string, therefore English
If the word is fine in German, then I think this is OK even though I'd prefer `<translation>Byte</translation>`

kripton: I'll change those to explicitly have the English wording in the translation elements

kripton: Fixed

ui/src/qlcplus_de_DE.ts
kripton: Fixed

FloEdelmann (approved):
Looks good now :)

mcallegari:
Thanks guys :+1:

PR #1459, opened 2023-09-30
Update Shehds-LED-Flat-Par-18x18W-RGBWA+UV.qxf

sbenejam:
The mode names were renamed and the aliases don't match the mode names anymore.


mcallegari:
Thanks! This might be a bug in the fixture definition editor. I'll double check

PR #1462, opened 2023-10-19
Overriding flash functionality

Wazzledi:
Hello,

I have used QLC the last year for my personal usage, but I always wanted a flash button which really overrides the flash scene to the output.

Since other people also requested this feature in the forums, I decided to take some time and implement this functionality.
Instead of just implementing this feature, I read in the forums that other users might not want this functionality.

Therefore, I implemented the possibility to activate this effect when the user wants to.
Two options were added:
1. Flash Override Priority - This adds a higher priority to the generic fader (Universe::Flashing). This priority is between Override and SimpleDesk. So, flash buttons will override sliders and also running chasers. We could also consider adding the possibility to choose the priority instead of toggling it to not override sliders but override running chasers. 
2. Flash Force LTP - This simply forces the flash scene channels to behave like LTP channels. This will override intensity channels also. 

I am quite new to the code base of qlc+. Therefore, this is more or less a first proposal and some things have to be discussed.
First, only the implementation was done. If that is an acceptable approach to give the users overriding flash buttons. I am willing to update the documentation and also look into the tests. 

Three things have to be discussed:
1. QLC+ does not have a central place where to store overall settings and behaviors. That's why I was unsure where to exactly add the flash property. Finally, I added the properties inside the Doc object. We should consider placing every overall setting into a class. 
DocStorage

1. Since QLC+ also does not have a specific settings page for overall settings, I added the options for QLC+ 4 in the Virtual Console General Setting Page. Is that acceptable? Otherwise, I will change the location to somewhere else.
widgetUiPropertyWindow

1. For QLC+ 5 I added a settings button as used in the functions and devices tab. Is that acceptable too, or should that be placed somewhere else?
qmlPropertyWindow

Please tell me your opinion on the implementation and whether you want to include this in the code base or not.
Also, tell me if i should add tests and documentation if this is accepted.


mcallegari:
Hi, thanks for this.
Here's my comments:
- can you please point me to the relevant discussions that led you to submit this PR?
- QLC+ does have a "central" place where to store settings (see all QSettings used) but they're not save into projects
- I need to know why this has to be a global setting rather than per-button one. As far as I know, flashing is needed only with buttons. I don't see a reason to use it elsewhere
- I'm not sure a new type of state is needed to stay before Simple Desk. A few years ago I added priorities to faders. Maybe that's enough to achieve priority flashing
- in general, such big changes should be discussed (e.g. in the development section of the forum) before submitting a PR to save time to you to write an already approved code and to me so that I don't have to review ideas not discussed and without the proper knowledge of the code base

Thanks

Wazzledi:
Hi,

- primarily, this was implemented due to personal needs and i thought it would be helpful for others. That's why I submitted a PR. Other forum discussions requiring this: https://www.qlcplus.org/forum/viewtopic.php?t=6871 and https://qlcplus.org/forum/viewtopic.php?t=14796 
- Alright, i did not Knew the QSettings.
- This does not have to be a global setting. I also thought about adding the possibility to the button itself. But that would add a decision to the user, which does not exactly know what this is for. And if a user really needs this, he can easily activate it. For all other users it would stay the same, and they do not have to make any further decision. I can change this with pleasure. I thought, normally, this decision would be a decision for the whole show file and not for each button. 
- I also don't think a new state is really needed. But what if later on in the development of QLC+ some functions also have the override priority set but should not override flash scenes. I just added it so that further implementation does not break this functionality. But yeah, the existing states would be enough to achieve this. 
- Of course. I literally forgot about the possibility to create a discussion in the forum and thought the discussion could be held here. I will add a forum discussion with pleasure if you want to.

Thanks

Wazzledi:
Forum discussion: https://www.qlcplus.org/forum/viewtopic.php?t=16739

yestalgia:
Massive thanks for sharing this.

This feature would be invaluable to my club shows and would open up a brand new way to program "bumps."

My thoughts:

- I can see why it would be useful to have buttons with both Override priority on and off for this reason I believe it would be best to have it available to select per button.
- This said, the "global" setting would be nice to have as a default for new "flash" buttons added to the console
- For the forum discussion: is there a reason you can think of to have a priority level selector? Would the user like a flash button be an override over other flashbuttons?

> Forum discussion: https://www.qlcplus.org/forum/viewtopic.php?t=16739
This link doesn't work for me

Wazzledi:
This link does not work right now because the thread was not approved by a moderator.
We have to wait.
I would like to continue this discussion in the forum as soon as the thread got accepted. 
Please also submit this response in the forum later.

mcallegari:
@Wazzledi  What is the difference between override and forceLTP now? I think one option could be enough for users since it's clear they want to override everything else when flashing?

Wazzledi:
@mcallegari Let me explain the difference.

ForceLTP:
This option just adds the `FadeChannel::ForceLTP` Flag to the created FadeChannel for the GenericFader.
This results in the new handling inside the `write` function for Flashing FadeChannels. When a FadeChannel is Flashing then it sets the `forceLTP` parameter of the `Universe::write(int channel, uchar value, bool forceLTP)` method according to whether the `FadeChannel::ForceLTP` flag is set or not.
This results in the universe handling the specific value as a LTP value.

OverridePriority:
As you might know, the universe has a priority list of GenericFaders which values should be written to the output.
For example: It exists GenericFader `a` and GenericFader `b` both have the same priority and `a` is added before `b`. Then, the value of `b` would override the value of channel `a` (assuming the value is LTP), because `b` is handled after `a`. This is normal behavior, but for flashing we want to prioritize this value. For that, we add a new priority so that the Flashing channels would be handled after the normal ones.
This happens for example when a chaser is running, and we want to flash a button. When the next step of the chaser is added, this overrides the flash button without this option.
Right now, the `Universe::Flashing` priority is greater than the `Universe::Override` priority, which results in Flash buttons overriding virtual console faders. 

I could imagine that the override priority configuration might be unnecessary or too much. But I think some users might not want to override some faders with the flash button. Maybe we can change the name and the handling to override faders.
Therefore, we would add a priority before `Universe::Override` and after. This will result that we would override everything per default except faders, and if we tick the option we also override faders.


In higher level:
ForceLTP forces the channels of the Flash button to be handled as LTP values. With this we can also flash dimmer values which are lower than the current dimmer values.

OverridePriority forces the handling of the channels to be later than other running functions. This will override virtual console faders and functions triggered after triggering the flash button.

engine/src/function.cpp
```diff
@@ -958,12 +962,16 @@ QList<quint32> Function::components()
...
 
-void Function::flash(MasterTimer *timer)
+void Function::flash(MasterTimer *timer, bool override, bool forceLTP)
 {
     Q_UNUSED(timer);
 
     if (m_flashing == false)
+    {
         emit flashing(m_id, true);
+        m_flashOverrides = override;
```
ClayPakyOfficial, line 972: I suspect some compilers will complain about this sooner or later, since it's a reserved keyword.
https://en.cppreference.com/w/cpp/language/override
Please rename

mcallegari, line 972: I suspect some compilers will complain about this sooner or later, since it's a reserved keyword.
https://en.cppreference.com/w/cpp/language/override
Please rename

engine/src/function.cpp
```diff
@@ -88,6 +88,8 @@ Function::Function(QObject *parent)
     , m_overrideFadeOutSpeed(defaultSpeed())
     , m_overrideDuration(defaultSpeed())
     , m_flashing(false)
+    , m_flashOverrides(false)
+    , m_flashForceLTP(false)
```
mcallegari, line 92: I wonder if we really want to flash anything other than Scenes.
In other words, are these really needed into the Function class or can they be moved to the Scene class?

engine/src/function.cpp
Wazzledi: Done!
Just a quick thing I noticed while going through the renaming: In simpledesk.cpp:130 there also exists a bool variable named override.

mcallegari: > Done! Just a quick thing I noticed while going through the renaming: In simpledesk.cpp:130 there also exists a bool variable named override.

Yep, you're right! Thanks for spotting. Will change it later today.

engine/src/function.cpp
mcallegari: I'd say let's do this step by step.
Cuestack is a sort of abandoned feature (I haven't even ported it to v5)
For now let's stick to Scene-only flashing.
Thanks

Wazzledi: Alright. I moved it to the Scene class.

mcallegari:
Alright, I'm testing this change. While forcing LTP is clear and works pretty well, priority override is still not clear to me.
When you flash a Scene, isn't it by design the last Function running?
Do you have a usage case in mind where while you're flashing a Scene other Function(s) are started over the one you're controlling and using the same channels? (e.g. a Chaser continuously changing)
It's the only case where I would see the necessity for keeping the flashing Scene prioritized. Or am I missing something?

Wazzledi:
Yes, the flash function is mostly the last function.
Indeed, the running Chaser is also the main thing I had in mind. I am not quite sure how it behaves with other running functions like EFX and Sequences. But this should also be overridden.

But, the most important thing for this feature is the fact that this can also override VCSliders. When setting the value of a slider, the output will be the flashing scene value.

mcallegari:
There was an error in the loadXML method.
See attached project. CueList won't load with your code
[flashOverride.zip](https://github.com/mcallegari/qlcplus/files/13301316/flashOverride.zip)
.

Wazzledi:
Okay. That's strange, because I only adapted the loadXML and saveXML function for the VCButton class.

I assume you are fixing this? Or should I?

mcallegari:
I just pushed a change.
Meantime I will test qmlui too

mcallegari:
Many thanks for this! :+1:

Wazzledi:
You're welcome. Should I also edit the documentation for this?

mcallegari:
> You're welcome. Should I also edit the documentation for this?

Sure, you can send a PR to this repo
https://github.com/mcallegari/qlcplus-docs

Otherwise I can do it when I have some time

PR #1474, opened 2023-10-31
The enable state of frame and solo frame is now supported in web access.

Itay-Lifshitz:


ui/src/virtualconsole/vcframe.cpp
```diff
@@ -127,6 +128,11 @@ void VCFrame::setDisableState(bool disable)
     updateFeedback();
 }
 
+bool VCFrame::enableState()
```
mcallegari, line 131: Please rename this to "enableButtonChecked"

mcallegari:
I tested this and it doesn't really work.
Please try the attached workspace.
I can see the enable buttons now but all the items inside a disabled frame keep working from the web interface. They shouldn't.

[frame_enable_test.zip](https://github.com/mcallegari/qlcplus/files/13236744/frame_enable_test.zip)

Also I don't understand why you changed indentation from 2 to 4 spaces. 4 spaces just increase the files size so they will be slower to load.

webaccess/res/virtualconsole.js
```diff
@@ -76,25 +76,25 @@ function setCueIndex(id, idx) {
 
 function sendCueCmd(id, cmd) {
  if (cmd === "PLAY") {
-   var obj = document.getElementById("play" + id);
```
mcallegari, line 79: Other indentation to restore

PR #1480, opened 2023-11-10
Cue List "Play/Stop + Pause" layout is now supported.

Itay-Lifshitz:
@mcallegari 
Please review and give feedback.

### DIFF
--- a/ui/src/virtualconsole/vccuelist.cpp
+++ b/ui/src/virtualconsole/vccuelist.cpp
@@ -1062,8 +1062,6 @@ void VCCueList::setPlaybackLayout(VCCueList::PlaybackLayout layout)
     }
 
     m_playbackLayout = layout;
-
-    emit playbackLayoutChanged();
 }
 
 VCCueList::PlaybackLayout VCCueList::playbackLayout() const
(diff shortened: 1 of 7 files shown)


mcallegari:
@Itay-Lifshitz there's no need to add an event for everything
Web interface is not meant to be live while editing so please remove the notification of layout change.
The user will just hit F5 on the browser.

Itay-Lifshitz:
@mcallegari 
Can you please check again?

ui/src/virtualconsole/vccuelist.cpp
```diff
@@ -877,6 +881,7 @@ void VCCueList::slotFunctionRunning(quint32 fid)
     else if (playbackLayout() == PlayStopPause)
         m_playbackButton->setIcon(QIcon(":/player_stop.png"));
     m_timer->start(PROGRESS_INTERVAL);
+    emit buttonAppearanceChanged();
```
mcallegari, line 884: I meant this is not needed and related slot

Itay-Lifshitz: @mcallegari If we remove this signal, the playback button icon will not sync.

Itay-Lifshitz: @mcallegari 
Of course, there are other ways to sync the button icons but the current version is very simple. 
Your thought?

mcallegari: Maybe we're just not using the proper terminology.
One thing is the layout change (Play + Stop vs Play + Pause), which is not needed during operate mode.
One other thing is the playback status change, which highlights the buttons with different colors.
Since you named this "buttonAppearanceChanged" I assumed it was the first case.
Then please rename it to "playbackStatusChanged"

mcallegari:
Thanks :+1:

PR #1490, opened 2023-12-14
straightline.js Pattern

skidoodl:
Greetings,

My apologies if my programming lingo isn't too great.

I am a novice programmer making a pattern for QLC+ not only as a school project but also for potential use in our show choir performance. I am looking for recommendations on what I should change in this effect and things I should add. The ultimate goal in my eyes is to have my pull request committed one day. Over the past year that I have been helping with the stage crew for our show I have become very passionate about this program and it would be a very high honor to have my code as a part of it. Although I understand if it needs some major work.

The intent for this pattern was for when we want a color/effect to only show up on a single row or section of the entire matrix. In the past we have just created groups for the individual rows/sections we would like to use, which some may know can be a time consuming task after having to do it for multiple sections. This effect will hopefully save us time when going through the process of creating the QLC+ scenes for our show.

**TO-DO LIST**
- [ ] **Implementing more effects from other QLC+ patterns** (noise especially, which I have had trouble trying to configure)
- [ ] **Changing the wording and function of some of the menu options** such as the fact that I have it set up so you can remove pixels from a line but then only move it left (for horizontal) or up (for vertical)
- [ ] **Allowing the user to stretch the line out**
- [ ] **Putting orientations to have 2 lines visible on the matrix at once** ex. two horizontal lines, one on top, other on bottom, both using the noise effect
- [ ] **Cut down the amount of code, there is _much_ that can be combined** ex. 1. the separate sections for each effect when using a vertical or horizontal line 2. the code for the noise effect being repeated twice in the code
- [ ] **Cutting down on the amount of commits in my branch** I plan on doing this later, I'd prefer to keep a timeline of my work
- [ ] **Making it so it does not get the width and the height of the matrix twice**

Will update this if/when I come up with more to add to the to-do list. Although I don't want the effect to be cluttered ;). 

Thanks
-aidan


mcallegari:
Hi, before reviewing this, please clean it up.
You pushed 35 commits where you added/removed files (probably by mistake or while experimenting?)
I suggest you close this PR and open a new one with a single commit of the script you want to add.
Thanks

PR #1492, opened 2023-12-17
Wizard for VC Widgets

8-Lambda-8:
I created a copy of the Function Wizard which creates widgets for all channels of the selected Fixtures.

Depending on the channel group, the widgets are configured differently (XY Pad, RGB, Click & Go, Intensity, ...).
image

i wanted this feature for myself, because i often work with different setups and wanted a quick way to setup widgets for groups of fixtures to directly control the channels.

Features i want to add in the Future:
- options for widget sizes
- option to split the selected features to frame pages
- input mapping


mcallegari:
Function wizard creates VC widgets too!
Why not adapting the existing tool?

8-Lambda-8:
Yes i know, but those widgets are to control the created functions and not to controll the fixtures directly. Because of that i thought it would be better to split them.
But of course i can implement my stuff into the existing one if you like it better like that.
if i do that, should i move the frames of the function widgets into the frame of the "channel widgets" or should i keep them seperate?

mcallegari:
Yes, I prefer to have one tool that does everything.
You can add a "Skip" button in the function creation tab.
> if i do that, should i move the frames of the function widgets into the frame of the "channel widgets" or should i keep them seperate?

What are "function widgets"?

8-Lambda-8:
Now everything is in one Wizard

yestalgia:
> Now everything is in one Wizard

Thanks for the contribution and Merry Christmas!

mcallegari:
OK, looks good. Can you please explain what are the numbers displayed beside the widget descriptions?

image

8-Lambda-8:
The numbers are the channel numbers of the fixture on which the widgets will act

yestalgia:
Can we add the wizard shortcut to the VC page as well? I liked the look of that.

This would make sense since it can be used for vc widgets.

mcallegari:
Here's another usage case: add a RGB panel (all defaults will do)
Start the wizard and select a few RGB panel rows.
This is the result

image

I would expect one slider for all the heads, but still someone would argue they want to select specific heads.
Also, listing involved channels in this case would be messy, so I would remove it or change it to "(N channels)" rather than the whole list which (at least for me) is irrelevant

8-Lambda-8:
Yes for big RGB Pannels it could be to much widgets, but maybe some users also want this
and yes one Slider to controll all Pixels would be good
image
maybe like this?

For multihead fixtures (and also groups of fixtures) maybe it would make sense to have multiple Pages of the Frame with the first page to controll ALL and the next pages for each head seperate. Maybe with a limit of pages?
image

I think the channel numers are a sometimes a helpful info. but yes for longer lists it may be better to write something like "(2-6)" instead of "(2, 3, 4, 5, 6)" or the start channel and the count "(5 channels starting with 2)"

should i put the wizard button also on the VC Page like @yestalgia sugested?

ETE-Design:
@8-Lambda-8 Is it ready for merge yet?

8-Lambda-8:
@ETE-Design no its not ready yet.
like @mcallegari said i have to change the behaviour a bit for RGB Pannels and Multihead fixtures.

8-Lambda-8:
Hi, it took a while, but 

The Wizard button is back on the VC Page, like @yestalgia wanted:
> Can we add the wizard shortcut to the VC page as well? I liked the look of that.
image


for the Multihead problem i solved it with Frame Pages. 
with the 3 checkboxes you can select what pages should be created. 

if the "All" page is created, the faders of all other pages have the "Monitor selected channels" option enabled 

example 8x60 RGB Matrix:
image
image
image

example 2x Chauvet Intimidator Spot duo
image
image

yestalgia:
This is awesome 😎

yestalgia:
@8-Lambda-8 Can you please merge master into your branch and push? Then I'll be able to test it.

8-Lambda-8:
i merged master again into my branch for testing.
is there anything i should change or do you think it is ready to be merged?

mcallegari:
Merged, thanks a lot for this contribution :+1:

mcallegari:
@8-Lambda-8 
This doesn't build with latest Qt version.
https://www.qlcplus.org/forum/viewtopic.php?t=18027
I just checked and you opted for a obsolete signal in UI.
Now I need to fix it

8-Lambda-8:
I'm sorry, I didn't know that

PR #1495, opened 2023-12-22
Color Picker & Padding

Taste62:
Added color pickers next to the color input fields and fixed a problem, where all preceding zeros in the hex color value would be removed on reload.
image


mcallegari:
Thanks :+1:

PR #1501, opened 2024-01-05
Flash override: Change button border to red when button uses override functions.

Wazzledi:
As suggested by mcallegari in the forum discussion about the overriding flash functionality, I changed the visualization of the flash button to display the border red instead of green when FlashForceLTP or OverridePriority is used.


yestalgia:
🔥

mcallegari:
Thanks :+1:

PR #1503, opened 2024-01-11
QLC+ 5: Add linear regression to BPM tapping

Wazzledi:
As described in forum thread https://www.qlcplus.org/forum/viewtopic.php?p=71374, I added the linear regression method used in QLC+ 4 to the BPM tapping functionality of QLC+ 5.

### DIFF
--- a/qmlui/qml/KeyPad.qml
+++ b/qmlui/qml/KeyPad.qml
@@ -20,6 +20,8 @@
 import QtQuick 2.0
 import QtQuick.Layouts 1.1
 
+import "TimeUtils.js" as TimeUtils
+
 import "."
 
 Rectangle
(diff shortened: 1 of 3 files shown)


qmlui/qml/KeyPad.qml
```diff
@@ -95,18 +101,60 @@ Rectangle
...
+                        var newTime = currTime - lastTap
+                        
+                        tapHistory.push(newTime)
+                        var tapHistorySorted = []
+
+                        //reduce size to only 16 taps
+                        while (tapHistory.length > 16) tapHistory.splice(0,1)
+
+                        //copy tap history to sort it
+                        for(var i = 0; i < tapHistory.length; i++)
```
mcallegari, line 122: space after `for`
But how about this?
`tapHistorySorted = tapHistory.slice();`

qmlui/qml/KeyPad.qml
```diff
@@ -95,18 +101,60 @@ Rectangle
...
+                        }
+                        tapHistorySorted.sort()
+
+                        // Find the median time between taps, assume that the tempo is +-40% of this
+                        var tapHistoryMedian = tapHistorySorted[Math.floor(tapHistorySorted.length/2)]
+                        
+                        //init needed variables
+                        var n = 1, tapx = 0, tapy = 0, sum_x = 0, sum_y = 0, sum_xx = 0, sum_xy = 0
+                        
+                        for(var i = 0; i < tapHistory.length; i++)
```
mcallegari, line 134: space after `for`

qmlui/qml/KeyPad.qml
Wazzledi: Adjusted!
Yes, this is far better.

qmlui/qml/KeyPad.qml
Wazzledi: Added the space.

mcallegari:
Hey, how about the Time edit tool too?
https://github.com/mcallegari/qlcplus/blob/master/qmlui/qml/TimeEditTool.qml

There's another time tapping

Wazzledi:
I totally forgot about this thing.
For that, I moved the algorithm in the TimeUtil.js and added it to the TimeEditTool.
While testing the TimeEditTool, I noticed that the Tap button is also shown when the chaser is set to beat.
Is that intended? Also, I dont see x/2 and 2x anymore to get half beat chasers.

mcallegari:
> For that, I moved the algorithm in the TimeUtil.js and added it to the TimeEditTool. 
Excellent. I would have done the same!

> While testing the TimeEditTool, I noticed that the Tap button is also shown when the chaser is set to beat. Is that intended? 
Nope. It's actually not needed

> Also, I dont see x/2 and 2x anymore to get half beat chasers.
I guess you mean like Speed Dial? TimeEditTool is not meant to be a replacement for Speed Dial

Wazzledi:
> I guess you mean like Speed Dial? TimeEditTool is not meant to be a replacement for Speed Dial
When editing the chaser step length with the TimeEditTool and the type is set to beat, I assumed the buttons x/2 and 2x, which are present in the code, should be present.
If they are not intended to be visible, then just ignore this comment.

mcallegari:
They are visible on a Beat-type RGB Matrix

image

mcallegari:
Many thanks for this :+1:

PR #1505, opened 2024-01-13
Save dialog geometry for all QLC+ dialogs

Timebutt:
This PR stores geometry settings for all dialogs in QLC+ that did not yet have this feature implemented. Some of the dialogs already had this (`AddFixture`, ...) which is very convenient, but some of them did not which can be incredibly frustrating when working on a 4K monitor and having to repeat an action many times during programming, but always having to resize the tiny dialog before you can make any change.

This PR does not include the same functionality for `QWidgets`, I think I saw a few of those that also used the `saveGeometry()` logic but not entirely sure if they all need this as well. If you think so @mcallegari I'll gladly do a pass over those as well!

### DIFF
--- a/fixtureeditor/addchannelsdialog.cpp
+++ b/fixtureeditor/addchannelsdialog.cpp
@@ -17,9 +17,13 @@
   limitations under the License.
 */
 
+#include <QSettings>
+
 #include "qlcchannel.h"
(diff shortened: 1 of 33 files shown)


mcallegari:
Very useful, thanks :+1:

PR #1507, opened 2024-01-21
resources: Add fixture definition for Bright XBAR

cybic:
Added fixture definition for the Bright XBAR 12 cell RGBAWUV LED bar

### DIFF
--- a/resources/fixtures/Bright/Bright-XBAR.qxf
+++ b/resources/fixtures/Bright/Bright-XBAR.qxf
@@ -1,0 +1,612 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE FixtureDefinition>
+<FixtureDefinition xmlns="http://www.qlcplus.org/FixtureDefinition">
+ <Creator>
+  <Name>Q Light Controller Plus</Name>
+  <Version>4.12.7</Version>
+  <Author>Øystein Steimler</Author>
+ </Creator>
+ <Manufacturer>Bright</Manufacturer>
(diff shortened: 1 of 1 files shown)


mcallegari:
Please name modes like "96 Channel"
Please add color macros

cybic:
Thanks for the feedback! I've tried to fix it ;)

> Please name modes like "96 Channel"
Fixed

> Add color macros
Fixed

> Add shutter presets
Fixed as best I could. Didn't find anything that matched "Strobe Audio Slow -> Fast".

> Also doesn't pass fixture validator https://www.qlcplus.org/fixture_validator.php
I actually hadn't seen the validator. I probably should have read some documentation.  Thank you for the patience :)

resources/fixtures/FixturesMap.xml
```diff
@@ -280,6 +280,9 @@
     <F n="Botex-DSP-405" m="DSP-405"/>
     <F n="Botex-SP-1500" m="SP-1500"/>
   </M>
+  <M n="Bright">
+    <F n="Bright-XBAR" m="Bright XBAR"/>
```
mcallegari, line 284: This is wrong. Manufacturer should be "Bright".
Please remove this change and leave this to me. (must be created with the dedicated Python script)

cybic: Ah, I saw the addition of the FixtureMap entry in some other commit adding a fixture, but that was of course your commit.
Done!

mcallegari:
Thanks :+1:

PR #1508, opened 2024-01-21
resources: Add fixture definition for Showtec LED Par 64 Short V2

cybic:
Add fixture definition for the Showtec LED Par 64 Short V2


mcallegari:
Thanks :+1:

PR #1513, opened 2024-02-02
resources: add Talent-SSL2

ktdreyer:
Started with Talent-SSL1 and updated according to Talent's user manual.

### DIFF
--- a/resources/fixtures/Talent/Talent-SSL2.qxf
+++ b/resources/fixtures/Talent/Talent-SSL2.qxf
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE FixtureDefinition>
+<FixtureDefinition xmlns="http://www.qlcplus.org/FixtureDefinition">
+ <Creator>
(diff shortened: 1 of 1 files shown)


yestalgia:
G'day, thanks for contributing to QLC+.

Unfortunately, this is not ready to merge yet. Even though it passes https://www.qlcplus.org/fixture_validator.php

The manual I found is here: https://www.parts-express.com/pedocs/manuals/244-520--talent-ssl2-user-manual-revised.pdf

Based on this manual I have a couple of suggestions:

1. Change x/y channels to pan/tilt per the doc
2. It should be 3 pin instead of 5 pin
3. Instead of "Red Light diming" it's okay to just be "Red"
4. The "Not Used" channel is actually a speed channel - I believe it's Macro speed?


Everyone has to start somewhere but Massimo is very fussy for good reason about the fixtures we add to QLC+.

I'd recommend posting fixtures on the forum instead of making pull requests here. Thanks :)

mcallegari:
@ktdreyer can you please review?

PR #1515, opened 2024-02-04
Added web interface for cue list notes live editing.

Binary-Vanguard-12138:
Hi, @mcallegari 
I added this PR for web interface of cut list notes live editing.
Please review it and give me feedback.

### DIFF
--- a/webaccess/src/webaccess.cpp
+++ b/webaccess/src/webaccess.cpp
@@ -773,6 +773,8 @@ void WebAccess::slotHandleWebSocketRequest(QHttpConnection *conn, QString data)
                     cue->slotNextCue();
                 else if (cmdList[1] == "STEP")
                     cue->playCueAtIndex(cmdList[2].toInt());
+                else if (cmdList[1] == "CUE_STEP_NOTE")
+                    cue->slotStepNoteChanged(cmdList[2].toInt(), cmdList[3]);
                 else if (cmdList[1] == "CUE_SHOWPANEL")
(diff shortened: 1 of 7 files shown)


webaccess/src/webaccess.cpp
```diff
@@ -1549,7 +1562,7 @@ QString WebAccess::getCueListHTML(VCCueList *cue)
         {
             QString stepID = QString::number(cue->id()) + "_" + QString::number(i);
             str += "<tr id=\"" + stepID + "\" "
-                   "onclick=\"enableCue(" + QString::number(cue->id()) + ", " + QString::number(i) + ");\">\n";
```
mcallegari, line 1552: Why? setCueIndex doesn't send the feedback back to the C++ code

Binary-Vanguard-12138: In the C++ version, the cuelist does not play when you click the step.
But in the web version, the cuelist played when you clicked the step.
So I changed the the enableCue function to setCueIndex.

What do you think?
I am open to discuss this.

mcallegari:
I just checked this one. Notes editing is OK but selecting a step from web and hitting play starts from the first element instead of from the selected one.
Not sure if this is because of the change discussed above or if it ever was like this but it needs to be fixed.
Can you please check?

Binary-Vanguard-12138:
@mcallegari Hi, I reverted setCueIndex to enableCue. Please review it again.

PR #1516, opened 2024-02-05
fixtures: Add Cameo P2 FC

cmuellner:
The Cameo P2 FC has 12 different modes.
This patch adds support for 7 of them, which don't depend on the device setting channel.


cmuellner:
Updated PR:
* adding dimensions (zero values caused a fail of the CI)

cmuellner:
Thanks for adding the missing modes.
I would have done this as well if you had asked for it.

Thanks for merging!

PR #1517, opened 2024-02-08
Webaccess virtual console grand master

Binary-Vanguard-12138:
webaccess virtual console grand master is supported.

### DIFF
--- a/webaccess/src/webaccess.cpp
+++ b/webaccess/src/webaccess.cpp
@@ -20,7 +20,7 @@
 #include <QDebug>
 #include <QProcess>
 #include <QSettings>
-#include <cmath>
+#include <qmath.h>
 
 #include "webaccess.h"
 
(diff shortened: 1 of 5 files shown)


webaccess/src/webaccess.cpp
```diff
@@ -2081,6 +2089,66 @@ QString WebAccess::getChildrenHTML(VCWidget *frame, int pagesNum, int currentPag
     return unifiedHTML;
 }
 
+void WebAccess::slotGrandMasterValueChanged(uchar value)
+{
+    GrandMaster::ValueMode value_mode = m_vc->properties().grandMasterValueMode();
+    QString g_value_show;
```
mcallegari, line 2095: Code style: we don't use word_one_two_three in this code. We use valueOneTwoThree.
Also, variables should have a self-explaining meaning. In this case: gmDisplayValue

webaccess/src/webaccess.cpp
```diff
@@ -2081,6 +2089,66 @@ QString WebAccess::getChildrenHTML(VCWidget *frame, int pagesNum, int currentPag
...
+    if (value_mode == GrandMaster::Limit)
+    {
+        g_value_show = QString("%1").arg(value, 3, 10, QChar('0'));
+    }
+    else
+    {
+        int p = floor(((double(value) / double(UCHAR_MAX)) * double(100)) + 0.5);
```
mcallegari, line 2102: Perhaps `<cmath>` include is not needed if you use `qFloor`

Binary-Vanguard-12138: Pushed fix.

PR #1520, opened 2024-02-12
inputprofiles: Adding Worlde Easypad.12 MIDI pad

cmuellner:
The Worlde Easypad.12 is a relatively cheap MIDI pad, that can be used as MIDI input. All controls that standard MIDI messages are supported here.

However, there are a few missing features:
* The slider sends SYSEX commands
* The bank button sends SYSEX commands
* Unclear how to set LED color. (using the feedback mechanism did not change anything)

I created this profile six years ago and forgot about it.
When I recently stumbled over the pad, I wanted to add support but found the old profile in my `.qlcplus` folder.
Enough reason to create a PR and see if it will be accepted.

### DIFF
--- a/resources/docs/html_en_EN/supported-input-devices.html
+++ b/resources/docs/html_en_EN/supported-input-devices.html
@@ -337,6 +337,16 @@ in <a href="https://www.qlcplus.org/forum/">QLC+ forum</a>.
 </TR>
 
 <TR>
+<TD><a href="https://www.worlde.com.cn/products_37/38.html">Worlde Easypad.12</a></TD>
+<TD><a href="midiplugin.html">MIDI</a></TD>
+<TD>1</TD>
+<TD>Worlde Easypad.12</TD>
+<TD>none</TD>
+<TD>no</TD>
+<TD>No bank button and slider (require SYSEX support)</TD>
+</TR>
+
+<TR>
 <TD><a href="https://www.zoom-na.com/products/production-recording/multi-track-recorders/zoom-r16-recorder-interface-controller">Zoom R16</a></TD>
 <TD><a href="midiplugin.html">MIDI</a></TD>
 <TD>1-16</TD>
--- a/resources/inputprofiles/Worlde-Easypad.12.qxi
+++ b/resources/inputprofiles/Worlde-Easypad.12.qxi
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE InputProfile>
+<InputProfile xmlns="http://www.qlcplus.org/InputProfile">
+ <Creator>
+  <Name>Q Light Controller Plus</Name>
+  <Version>4.13.0 GIT</Version>
+  <Author>Christoph Müllner</Author>
+ </Creator>
+ <Manufacturer>Worlde</Manufacturer>
+ <Model>Easypad.12</Model>
+ <Type>MIDI</Type>
+ <Channel Number="36908">
+  <Name>Record</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36909">
+  <Name>Play</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36910">
+  <Name>Stop</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36911">
+  <Name>Back</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36912">
+  <Name>Next</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36913">
+  <Name>Refresh</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36993">
+  <Name>Left</Name>
+  <Type>Button</Type>
+ </Channel>
+ <Channel Number="36994">
+  <Name>Right</Name>
+  <Type>Button</Type>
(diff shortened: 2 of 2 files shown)


yestalgia:
This controller is wild lol. This is from their docs at https://www.swamp.net.au/images/otherimages/manuals/EASYPAD_USER_MANUAL.pdf
image

Very odd.

cmuellner:
Yeah, the manual is a bit...
But I don't care what's written in the manual, as long as the device serves my use-case.

FWIW, I ran `aseqdump` to dump the SYSEX commands.

Slider:
* Bottom: `F0 7F 7F 04 01 00 00 F7`
* Top: `F0 7F 7F 04 01 00 7F F7`
* The second byte from the end goes linear with the position 00..7F

Bank button:
* Switched to bank 1: `F0 42 40 00 01 04 00 5F 4F 00 F7`
* Switched to bank 2: `F0 42 40 00 01 04 00 5F 4F 01 F7`
* Switched to bank 3: `F0 42 40 00 01 04 00 5F 4F 02 F7`
* Switched to bank 4: `F0 42 40 00 01 04 00 5F 4F 03 F7`
* Switching goes in circle (1->2->3->4->1...)

I looked into the QLC+ and as far as I understood, there is not much support for SYSEX commands.
However, when looking at the sent messages, it is obvious how mapping the slider and bank button to elements in the virtual console could be done.
The issue here is, that having device-specific code in QLC+ does not make sense. But having generic SYSEX support that can be utilized by the .qxi files sounds like quite some work (e.g. this could be done by defining a message mask and a conversion method to extract the input value).

My use case is to control the lights in a theater and play audio effects.
At the moment I have an external 6-channel DMX mixer for cross-fading (stepping through the cue list). The buttons of the pad are used for the audio effects. If the fader of this device could be used to control a slider on the virtual console, I could leave the 6-channel DMX mixer in the box.

mcallegari:
@cmuellner thanks for the PR and your analisys. I am currently working on the topic raised in #1514 and trying to find a generic solution for devices with different needs. I might commit something more later today.
SysEx is indeed supported by the MIDI plugin when it sends initialization messages. See
https://github.com/mcallegari/qlcplus/blob/master/plugins/midi/src/common/midiplugin.cpp#L309

In the meantime I purchased a couple of these Worlde controllers to be able to make tests myself. The more QLC+ is flexible, the better. I'll keep you posted on the progress

cmuellner:
> @cmuellner thanks for the PR and your analisys. I am currently working on the topic raised in #1514 and trying to find a generic solution for devices with different needs. I might commit something more later today. SysEx is indeed supported by the MIDI plugin when it sends initialization messages. See https://github.com/mcallegari/qlcplus/blob/master/plugins/midi/src/common/midiplugin.cpp#L309

I've seen this, but this is the other direction (sending a custom message to the device and not receiving a SYSEX message from the device that will be further processed).

I could imagine to have SYSEX messages supported like this in the qxi:
```
 <Channel Number="SYSEX">
  <Name>Fader 1</Name>
  <Type>Slider</Type>
  <SysexMask>FF FF FF FF FF FF 00 FF</SysexMask>
  <SysexVal>F0 7F 7F 04 01 00 00 F7</SysexVal>
  <ExtractionType>SingleByte</ExtractionType>
  <ExtractionTypeArgs>6</ExtractionTypeArg>
 </Channel>
```

So whenever a SYSEX message is received the channel description in the qxi will be probed for:
* Number="SYSEX"
* (message AND $SysexMask) == $SysexVal
* The message is then further processed by a value extractor
* For the channel above the value extractor is "SingleByte" (extracting a single byte and using its value as input value)
* The "ExtractionTypeArgs" allows to specify additional parameters for the value extractor (in this case, the index of the relevant byte in the message)

This allows keeping the device-specific part in the qxi, while being extendable for other value extractors.

> In the meantime I purchased a couple of these Worlde controllers to be able to make tests myself. The more QLC+ is flexible, the better. I'll keep you posted on the progress

Sounds good! Let me know if I can help.

At this point, I am curious if the channel numbers are equal on all devices on the market.

Independent of the SYSEX topic, I would be happy if the device support would be accepted with a slightly reduced feature set (i.e., what is covered in the PR).

mcallegari:
@cmuellner merged, thanks. Please keep in mind that [docs are not bundled anymore](65e1906d5c3badcfbbeb3ea1eeb2f2ca4bce7a82) with QLC+. To add a supported device you need to submit a PR to the [documentation repository](https://github.com/mcallegari/qlcplus-docs), which is in sync with https://docs.qlcplus.org/v4

cmuellner:
I looked at the docs repo, and there is no list of supported Midi devices. There is just a page describing Midi profiles (https://docs.qlcplus.org/v4/plugins/midi generated from pages/09.plugins/06.midi/default.v4.md). Since there are no changes to the Midi profiles (we just added support for a new device), no changes in the docs are needed.

mcallegari:
Hey @cmuellner today I just received my easypad 12. In omni mode I can receive all the pads+button+fader as control change.
Perhaps they change the firmware and now it's much more standardized? I saw there might be firmware updates and configuration tools in their website.
https://www.worlde.com.cn/list_17/

cmuellner:
Good to know! I currently don't have it at home, but I will get my hands on it again later this week and try to update the FW.

Are the channel numbers the same odd values (e.g. 37031 for Pad 1)?

mcallegari:
Yes, everything seems to be same as your mapping. I just added the XY fader

PR #1526, opened 2024-03-16
Updated catalan ans spanish translations

sbenejam:
Updated catalan ans spanish translations


mcallegari:
Thanks @sbenejam :+1:

PR #1530, opened 2024-03-20
Add beamZ BAC500 and BAC506 fixtures

omichel:
This PR adds a couple of beamZ fixtures:

- [x] Passed the [validator test](http://www.qlcplus.org/fixture_validator.php) for both fixtures.


mcallegari:
Thanks :+1:

PR #1537, opened 2024-03-25
VC Click & Go Gobo control on limited-range knobs and sliders

hjtappe:
Originally, I came from the v4 issue that having a gobo knot which is limited to values for static gobos only did not select the correct gobo values when clicking on the CnG Gobo entry.

Please find attached the simple workspace showing an example setup:
"Midi" input (simulated on an Artnet universe) --> Range- limited knob --> Fixture

[gobo-vc.qxw.txt](https://github.com/mcallegari/qlcplus/files/14746163/gobo-vc.qxw.txt)

Along the way when also trying the example in v5, I cam across a few otherv5  knob / slider  limitations which are fixed in this patch:

- Gobo images were not shown
- VCslider limits were not respected when attaching to an Input Event (e.g. Midi)
- Mouse Wheel inputs were not yet supported for the vcKnob
- only the youngest knob or fader was able to modify a LTP channel value
- The highest value was not accessible in the CnG selection list

I think, the functionality only applies to the Gobo / CnGPresets CnG type, not the CngColors type. Please review if that is correct.

I hope this helps. Please review carefully as this is my first QML patch.

### DIFF
--- a/qmlui/qml/virtualconsole/VCSliderItem.qml
+++ b/qmlui/qml/virtualconsole/VCSliderItem.qml
@@ -230,7 +230,7 @@ VCWidgetItem
                 if (Qt.platform.os === "android")
                     presetImageBox.source = cngResource
                 else
-                    presetImageBox.source = "file:/" + cngResource
+                    presetImageBox.source = "file:" + cngResource
             }
 
             onClicked: colorToolLoader.toggleVisibility()
--- a/qmlui/qml/virtualconsole/VCSliderItem.qml
+++ b/qmlui/qml/virtualconsole/VCSliderItem.qml
@@ -17,7 +17,7 @@
   limitations under the License.
 */
 
-import QtQuick 2.0
+import QtQuick 2.14
 import QtQuick.Layouts 1.1
 
 import org.qlcplus.classes 1.0
@@ -149,7 +149,7 @@ VCWidgetItem
             to: sliderObj ? sliderObj.rangeHighLimit : 255
             value: sliderValue
 
-            onMoved: if (sliderObj) sliderObj.value = value // position * 255
+            onValueChanged: if (sliderObj) sliderObj.value = value
         }
 
         // widget name text box
--- a/qmlui/qml/QLCPlusKnob.qml
+++ b/qmlui/qml/QLCPlusKnob.qml
@@ -82,6 +82,34 @@ Dial
             context.stroke()
             context.closePath()
         }
+
+        MouseArea
+        {
+            anchors.fill: parent
+            z: 2
+            onWheel: {
+                console.log("Wheel delta: " + wheel.angleDelta.y)
+                from: sliderObj ? sliderObj.rangeLowLimit : 0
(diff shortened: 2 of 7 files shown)


mcallegari:
@hjtappe thanks for this
First comment: I saw you added a MosueArea to the Knob item just to handle mouse wheel. It seems Dial has a wheelEnabled property inherited from Control. Can you please try that? It would simplify the code

hjtappe:
Hello @mcallegari, yes, I also tried that in a first attempt. It saves a lot of MouseArea code.

> `+            wheelEnabled: true`
> `+            stepSize: 1.0`

It turns the knob, but it does not trigger the onMoved in VCSliderItem.qml or if I add this into QLCPlusKnob.qml.

> `-            onMoved: if (sliderObj) sliderObj.value = value // position * 255`
> `+            onMoved: {`
> `+                console.log("onMoved: " + value)`
> `+                if (sliderObj) sliderObj.value = 23 //value // position * 255`
> `+            }`

Consequently, it does not update the sliderObj.value and thus does not update the other connected sliders.

Do you have an idea what blocks the onMoved event?

I have no idea why this event is not triggered when the knob turns through the mouse wheel (but only when dragged).

mcallegari:
We might have stumbled on a Qt bug, since onMoved works correctly with the Slider item.
I tried onValueChanged and it works with mouse wheel. However I'm not sure if it has specific drawbacks.
For sure it's not "interactive user move" but it involves also values changed programmatically.

hjtappe:
I have now tried a few things to try to make it work and increased the imported version. But the onMoved might work only with a more recent QT version.
So with onValueChanged it now works with less code, even though the event is triggered twice for each change. Nevertheless, in total, it should be faster to execute and works like this.
Should we add some hint that this workaround might need attention in a later version?

qmlui/qml/virtualconsole/VCSliderItem.qml
```diff
@@ -230,7 +230,7 @@ VCWidgetItem
                 if (Qt.platform.os === "android")
                     presetImageBox.source = cngResource
                 else
-                    presetImageBox.source = "file:/" + cngResource
```
mcallegari, line 233: I noticed you removed the extra slash in a few places. Which operating system are you on? I suspect this won't work on every supported platform.

hjtappe: It is Linux AppImage. Gobos don't show otherwise. The path is otherwise searched somewhere in the root directory rather than in the relative filesystem.

mcallegari:
@hjtappe unfortunately the qml part was not entirely ok
The preset capability item referenced the slider object even when no slider was involved, so it broke all the usages outside the VC slider click & go one.
Proper fix is to define the range low/high limit in the main preset tool and then determine the visibility of each capability at construction time.
Can you please verify that my last change still works with C&G limited range? Thanks

hjtappe:
@mcallegari thanks for your review and fix. I confirm that the current master works on v4 and v5 both on Windows and Linux AppImage.

PR #1539, opened 2024-03-27
qmlui VC "not implemented" widgets

hjtappe:
Would it make sense to implement a quick stub for the VC widgets which are currently missing to save confusion with users trying the pre-version of v5, such as https://www.qlcplus.org/forum/viewtopic.php?t=17142

The VC will then just display a "not implemented" inside a widget if it is loaded from an existing file or if it is added from the sidebar.


mcallegari:
Excellent code contribution, thanks :+1: 
Next preliminary thing to do is to make sure load/save doesn't lose information on an existing project.
I'll start adding some code too

Issue #1542, opened 2024-03-30
[QLC+ 5] Cannot use the full 512 channels of a universe


mxdumas:
**Describe the bug**
QLC+ does not allow to use the full 512 channels on a universe. For example, if you try to add 16x 32 channels fixtures, it will patch the first 15 to universe 1 and the last one to universe 2, even though they should fit on a single universe.

**To Reproduce**
Add 16 Martin Mac Viper Performance in Basic mode to the patch.

**Expected behavior**
16x32=512, so it should fit in a single universe. Fixture 16 should be patched at 1.481.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - QLC+ 5.0.0 Beta 3

mxdumas:
See proposed fix in PR #1543. Thank you!

mcallegari:
Fixed by #1543

PR #1543, opened 2024-04-01
Fix fixture patch address verification

mxdumas:
Fix for issue #1542 . 

fxAddress is 0 based (0-511). If the requested address is 511 and the unit has 1 channel, it should fit. In this case, (fxAddress = 511) + (channels = 1) == (UNIVERSE_SIZE = 512) => Should be ok. 

Simply need to remove the "=" to fix the issue.

Easy fix for a first timer! Was fun to figure my way around. ;)


mcallegari:
Nice catch, thanks :+1:

Issue #1532, opened 2024-03-20
[QLC+ 5][win11] RGB Panel patching cause crashing


testbild-media:
On patching a RGB Panel to a channel which is already used by an existing RGB Panel, qlc5 cause instant crashing/closing.

**Expected behavior**
A Message that Channels are in use and can not be patched.

**Desktop:**
 - OS: Windows 11
 - QLC+ Version 5.0.0 Beta 3

mcallegari:
Not reproduced.
Please indicate the exact steps to reproduce the issue

lilyCalla:
I was able to reproduce.

**Desktop:**
- OS: Ubuntu 22.04, MacOS 13.4.1
- QLC+ Version 5.0.0 Beta 3

**Reproduction Steps**
1. Patch a fixture (can be on any universe or channel, does not need to overlap with later fixtures)
2. Remove fixture
3. Patch new fixture
4. Add Generic RGB Panel such that its channels overlap with the channels of the fixture added in step 3
5. Observe Crash

Truncated Log (all information before interaction with application removed):
https://gist.github.com/lilyCalla/9a1d5f263700d796021f188f1f50df6b

mcallegari:
The crash itself is fixed.
For a proper address validation much more work needs to be done

PR #1551, opened 2024-04-19
qmlui: verify fixture validity in addRGBPanel

lilyCalla:
Fix for issue #1532

slotFixtureAdded should only be called if a fixture is added. This check does not currently exist in addRGBPanel, and causes instability. Adding this check appears to fix the application crash.


lilyCalla:
Closing due to clean up commit

PR #1553, opened 2024-04-21
UI scaling improvements

lilyCalla:
Fix general UI scaling bugs including

- Top bar elements drift downwards when scale goes under 1.0
- Background for "Actions" bar didn't automatically adjust to rescaled sizing (required second click on "Actions" menu item)
- Prevent BPM "ratchet" issues where scaling up increases keypad button height, but scaling down leaves it at the maximum height scaled to
- Beat Indicator / Stop all button not scaling with UI
- interface jitter in UI Setting editor
- Width of text elements in UI settings editor, but font size changes
- Undo/Redo buttons to layout to scale properly
- Prevent text overflow in about dialog

Current (Beat generator panel has been scaled up, then returned to smaller size):
pre-pr

Pull Request:
post-pr

### DIFF
--- a/qmlui/qml/UISettingsEditor.qml
+++ b/qmlui/qml/UISettingsEditor.qml
@@ -37,7 +37,6 @@ Rectangle
 
     onVisibleChanged:
     {
-        console.log("visibility change");
         origItemHeight = UISettings.listItemHeight
         origIconMedium = UISettings.iconSizeMedium
         origTextSizeDefault = UISettings.textSizeDefault
--- a/qmlui/qml/UISettingsEditor.qml
+++ b/qmlui/qml/UISettingsEditor.qml
@@ -30,13 +30,17 @@ Rectangle
     anchors.fill: parent
     color: "transparent"
 
-    property real origItemHeight: UISettings.listItemHeight
-    property real origIconMedium: UISettings.iconSizeMedium
+    property real origItemHeight: { origItemHeight = UISettings.listItemHeight }
+    property real origIconMedium: { origIconMedium = UISettings.iconSizeMedium }
+    property real origTextSizeDefault: { origTextSizeDefault = UISettings.textSizeDefault}
+    property real origIconDefault: {origIconDefault = UISettings.iconSizeDefault}
 
     onVisibleChanged:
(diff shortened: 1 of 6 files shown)


qmlui/qml/UISettingsEditor.qml
```diff
@@ -30,13 +30,17 @@ Rectangle
     anchors.fill: parent
     color: "transparent"
 
-    property real origItemHeight: UISettings.listItemHeight
-    property real origIconMedium: UISettings.iconSizeMedium
+    property real origItemHeight: { origItemHeight = UISettings.listItemHeight }
```
mcallegari, line 33: Are you sure this syntax makes sense?

qmlui/qml/UISettingsEditor.qml
```diff
@@ -30,13 +30,17 @@ Rectangle
...
+    property real origIconMedium: { origIconMedium = UISettings.iconSizeMedium }
+    property real origTextSizeDefault: { origTextSizeDefault = UISettings.textSizeDefault}
+    property real origIconDefault: {origIconDefault = UISettings.iconSizeDefault}
 
     onVisibleChanged:
     {
+        console.log("visibility change");
```
mcallegari, line 40: Debug print?

mcallegari:
Hey, thanks for this.
Just 2 comments but the rest looks fine to me.

qmlui/qml/UISettingsEditor.qml
lilyCalla: This syntax is meant to prevent a binding from being created on `origItemHeight.` To my knowledge this is the cleanest way to do this inline, although I'm open to being wrong

qmlui/qml/UISettingsEditor.qml
lilyCalla: Sorry, didn't notice that, fixed

qmlui/qml/UISettingsEditor.qml
mcallegari: TBH I've never seen that syntax before.
I'll merge this as it is. Thanks

PR #1556, opened 2024-04-28
Use threading in VinceUSBDMX512 plugin and fix it

jeromelebleu:
The VinceUSBDMX512 plugin is not working anymore in v4.13.0, the interface opening and closing are working but nothing is send to the output. I didn't spent too much time to try to understand why but instead I rewrite some part of it to use threading as the other dmxusb plugins. The output is now working again, there is still the input to code but I don't have required equipment to test.

Any review is welcome since I am not really familiar anymore with C++/Qt/FTDI… Thanks!

### DIFF
--- a/plugins/dmxusb/src/vinceusbdmx512.cpp
+++ b/plugins/dmxusb/src/vinceusbdmx512.cpp
@@ -18,41 +18,25 @@
 */
 
 #include <QDebug>
+
 #include "vinceusbdmx512.h"
 
-VinceUSBDMX512::VinceUSBDMX512(DMXInterface *interface, quint32 outputLine)
-    : DMXUSBWidget(interface, outputLine, DEFAULT_OUTPUT_FREQUENCY)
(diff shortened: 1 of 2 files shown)


mcallegari:
@jeromelebleu thanks for this :+1: 
Curious thing is that I can't really find anything about this interface by googling it!
I wanted to purchase one to provide proper support...but it seems it doesn't exists!
Looking at your changes I presume you have one and tested DMX output is working correctly now?
Thanks

jeromelebleu:
Unfortunately this interface is no longer available for sale, and the website is down for a while… Going back in time thanks to the Wayback Machine we can still have some information about it [here](https://web.archive.org/web/20111128114636/http://www.dmx512-online.net/fr-article65.html) but links to schematics and other files are down unfortunately - it was an open hardware.

I still have and using mine - and I'm maybe one of the last dinosaur who uses it especially with QLC+ - and the DMX output is working again with these changes.

mcallegari:
Many thanks for the info! :+1: 
I'll merge it right away then, thanks

PR #1561, opened 2024-05-01
Fix color picker full mode

lilyCalla:
This PR fixes a bug where using the "Full" color picker mode doesn't update the color of fixtures or UI color settings.


mcallegari:
Thanks :+1:

PR #1518, opened 2024-02-09
Android build v5

Binary-Vanguard-12138:
Used Android SDK31/33, JDK17, NDK21/25, and gradle 7.3 on Windows/Ubuntu Qt.
It crashes in Nox emulator, will need to test it on real device and check if it works.

PR #1562, opened 2024-05-01
Added missing resources to assets in Android build

Binary-Vanguard-12138:
I am opening PR to add missing resources in assets directory in Android build.
We discussed it [here](https://github.com/mcallegari/qlcplus/pull/1518#issuecomment-2084782404).


resources/CMakeLists.txt
```diff
@@ -1,5 +1,9 @@
 project(resources)
 
+if (NOT qmlui)
```
mcallegari, line 3: Please remove. Docs are no longer bundled with QLC+

PR #1563, opened 2024-05-02
QML version bumps

mlowczynski:


mcallegari:
Hey @mlowczynski thanks for this
I'm wondering if it makes sense to actually omit the version number so that the latest version is always imported:
https://doc.qt.io/qt-6/qtqml-syntax-imports.html

mlowczynski:
@mcallegari I have bumped QML module imports to the latest versions supported by Qt5 (5.15 LTS). This makes QML compatible with both Qt5 and Qt6. If the lack of compatibility with Qt5 is no longer a problem for us, it is better to use unqualified module import, as you suggested

mlowczynski:
I think this is a good approach. Do I understand correctly that we can close this PR for now?

mcallegari:
Yeah, it's probably better to submit a new one. Sorry for the change of mind but this is a big change that needs careful attention.

mlowczynski:
No problem

PR #1564, opened 2024-05-10
Refresh Cue List Widget on Addition or Removal of a Step from Chaser

lilyCalla:
- Currently adding or removing a widget is not reflected in the cue list widget unless the project is saved and reopened 
- Also apparent cause of crash at https://www.qlcplus.org/forum/viewtopic.php?t=17147, as the cue list would attempt to access a non-existent index
- This PR reloads the chaser cue list if a function is added or removed from the linked chaser, allowing updates to be seen in real time and preventing the crash

### DIFF
--- a/engine/src/chaser.cpp
+++ b/engine/src/chaser.cpp
@@ -132,6 +132,7 @@ bool Chaser::addStep(const ChaserStep& step, int index)
         }
 
         emit changed(this->id());
+        emit stepListChange(this->id());
         return true;
     }
     else
@@ -150,6 +151,7 @@ bool Chaser::removeStep(int index)
         }
 
         emit changed(this->id());
+        emit stepListChange(this->id());
         return true;
     }
     else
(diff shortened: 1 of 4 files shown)


yestalgia:
Love your work @lilyCalla

mcallegari:
Looks good, thanks 👍

mcallegari:
Actually, I overlooked these changes.
- first signals should be using past form, so in chaser.h/cpp it's `stepsListChanged`, not `stepListChange`
- also, a signal to update the UI was already present: https://github.com/mcallegari/qlcplus/blob/master/qmlui/virtualconsole/vccuelist.h#L198

I think the only change needed in vccuelist.cpp is this:
connect(function, SIGNAL(stepsListChanged(quint32)),
                this, SIGNAL(stepsListChanged()));

Please try that and re-submit a new PR. Thanks

mcallegari:
Nevermind, I sorted this myself.

PR #1568, opened 2024-05-16
Added beamZ PS10W fixture

YesterKo:
Hi! I Added beamZ PS10W fixture mapping. There is no info about DMX modes in the manual, so manually tested them. Should be all good and tested the mappings, works perfectly.

Manual here: https://www.beamzlighting.com/product/ps10w-led-pin-spot-10w-4-in-1-dmx/

### DIFF
--- a/resources/fixtures/beamZ/beamZ-PS10W.qxf
+++ b/resources/fixtures/beamZ/beamZ-PS10W.qxf
@@ -16,21 +16,21 @@
  <Channel Name="White" Preset="IntensityWhite"/>
  <Channel Name="Strobe">
   <Group Byte="0">Shutter</Group>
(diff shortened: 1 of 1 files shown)


yestalgia:
Fixture passes validator ✅
Fixture Manual: https://www.beamzlighting.com/wp-content/uploads/2024/05/151.259-PS10W-LED-Pin-Spot-10W-RGBW-DMX_manual_V1.3.pdf

My observations:
Nice work! I think you need to change the names of the modes from "4ch" to "4 Channel"

The manual also refers to your Color Wheel channel as "Auto." Perhaps you need to change the description to that.

yestalgia:
Also within the "Auto" channel, it should state the descriptions found in the manual. E.g. "No Function" and "Colour random, from slow to fast"

mcallegari:
Fixed and added, thanks

PR #1571, opened 2024-05-22
Proposal of a rework of HID device names

qfulmina:
Hi,

this is a proposal for a more detailed and more robust naming of HID devices.

It covers:
1. HIDDMXDevices always have a serial number, thus it is appended in parenthesis to be able to distinguish several connected devices.
2. Those HIDJSDevices that expose a serial number append it in parenthesis as well.
3. HIDJSDevices not exposing _any_ product or manufacturer name use VID:PID as a naming scheme instead. (Previously those had a completely empty name.)

I'm open to changes of course.

### DIFF
--- a/plugins/hid/hidjsdevice.cpp
+++ b/plugins/hid/hidjsdevice.cpp
@@ -39,8 +39,9 @@ static QString assembleDevicesName(struct hid_device_info *info)
     if (name_part.trimmed().isEmpty())
     {
         //use the vendor and product_id combination if name is empty
-        name_part = QString::number(info->vendor_id, 16) + ":" +
-                    QString::number(info->product_id, 16);
+        name_part = "HID Input Device (" +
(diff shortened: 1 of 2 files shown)


qfulmina:
Sure:

https://forum.qlcplus.org/download/file.php?id=13665

Sorry for the bad crop (missing the first letter and the closing brace on the second line). Right now I don't have those devices around. I hope you get an idea. (I can prepare another screenshot once I get back to the devices.)

First one didn't tell any name thus showing vendor ID and product ID.
The second did expose a name. And a serial number: AZ-RB- ... .

mcallegari:
For devices with no name/vendor I would give them a generic name like "HID Input device (VID:PID)" with VID/PID toUpper()

qfulmina:
Alright, will rework that.

qfulmina:
Changed it accordingly and it looks great:

hid-device-names

PR #1573, opened 2024-05-22
Add feature stopOnExit to script

ldebs:
This feature introduces a new script command: `stoponexit`.

The purpose of this command is to control whether functions started during the script should be stopped at the end. When `stoponexit` is set to `true`, all functions started afterward will be stopped when the script finishes. By default, `stoponexit` is `true`.

Example usage:
```plaintext
startfunction:1

stoponexit:false
startfunction:2

stoponexit:true
startfunction:3

stoponexit:false
startfunction:4
```
In this script, functions 1 and 3 will be stopped at the end, but functions 2 and 4 will continue running.


mcallegari:
is this backward compatible or does it break current projects?

ldebs:
It should be backward compatible since the default value of this flag is true, which is the standard behavior.

ldebs:
Sorry for the initialization order; I hadn't compiled the QML version. I just did, and I changed the implementation of the QML version to be compliant with the description I gave of this feature. Currently, my local code also includes a new feature to wait for a function in the script. I need to separate this new feature in order to extract the right modification to amend the stopOnExit feature.

ldebs:
Commit correction of v5 behavior done.
Test on v5 ok.

ldebs:
Merged with #1581

PR #1581, opened 2024-06-02
Add features stopOnExit and waitFunctionStart and waitFunctionStop to scripts

ldebs:
The purpose of this feature is to add **waitFunctionStart** and **waitFunctionStop** commands that are used to synchronize the execution of functions.

The start and stop signals are used by the commands to wait for the function to start or to complete before continuing the script execution.

The waitFunctionStart command causes the script to wait until a function with the given ID starts running. The script will continuously poll the function's state until it is determined that the function has started. Once the function is found to be running, the script will resume execution from where it left off.

Similarly, the waitFunctionStop command causes the script to wait until a function with the given ID stops running. The script will continuously poll the function's state until it is determined that the function has stopped. Once the function is found to be stopped, the script will resume execution from where it left off.

Example usage:

* QLC v4:
```
waitfunctionstart:2 // Wait for function 2 to start
waitfunctionstop:0 // Wait for function 0 to stop
startfunction:3 // Start function 3
```

* QLC v5:
```
Engine.waitFunctionStart(2); // Wait for function 2 to start
Engine.waitFunctionStop(0); // Wait for function 0 to stop
Engine.startFunction(3); // Start function 3
```

In this script, the function 3 is started only if function has started and function 0 has stopped after function 0 as started.

-----------------

**The pull request #1573 has been merged to this one**

The stopOnExit feature introduces a new script command: stoponexit.

The purpose of this command is to control whether functions started during the script should be stopped at the end. When stoponexit is set to true, all functions started afterward will be stopped when the script finishes. By default, stoponexit is true.

Example usage:

* QLC v4:
```
startfunction:1

stoponexit:false
startfunction:2

stoponexit:true
startfunction:3

stoponexit:false
startfunction:4
```

* QLC v5:
```
Engine.startFunction(1);

Engine.stopOnExit(false);
Engine.startFunction(2);

Engine.stopOnExit(true);
Engine.startFunction(3);

Engine.stopOnExit(false);
Engine.startFunction(4);
```

In this script, functions 1 and 3 will be stopped at the end, but functions 2 and 4 will continue running.


ldebs:
Tested on v4 and v5 and working fine

ldebs:
>     * please don't mix pull requests. It's a bad practice and lead to confusion

As this pull request need and include the #1573 one, I close the #1573.

>     * no need for polling the function status (which is very inefficient). The Doc class has signals for function running/stopped

I can't find the signals running/stopped handling in the Doc class.

For v4 (Script class), as I implemented it, there is no such things as polling the function status, except for the first execution of the waitFunction commands and I just call the running() or stopped() methods of the function that just return the internal field value.
The write loop exits because the waiting() function return false because the m_waitFunction is not NULL.
the value of m_waitFuntion is set to NULL when the function signals is received.

>     * this feature is very risky and can lead to Script hanging while waiting for external triggers. You need to be 100% sure the Script can be killed at any time and doesn't consume CPU for no reason.

The implementation has the same behavior as the wait command. The write loop is interruptible in the same way.

I adapted the v5 (ScriptRunner class) to handle signals like the v4 version and thus consume less CPU during the write loop.


Please, let me know how to handle running/stopped signals through Doc class

engine/src/scriptrunner.cpp
```diff
@@ -206,6 +232,23 @@ bool ScriptRunner::write(MasterTimer *timer, QList<Universe *> universes)
     return true;
 }
 
+void ScriptRunner::slotWaitFunctionStarted(quint32 fid)
+{
+    if (m_waitFunction != NULL && m_waitFunction->id() == fid) {
```
mcallegari, line 237: Indentation: parenthesis on new line

engine/src/scriptrunner.cpp
```diff
@@ -206,6 +232,23 @@ bool ScriptRunner::write(MasterTimer *timer, QList<Universe *> universes)
...
+        disconnect(m_waitFunction, SIGNAL(running(quint32)), this, SLOT(slotWaitFunctionStarted(quint32)));
+        m_waitFunction = NULL;
+    }
+}
+
+void ScriptRunner::slotWaitFunctionStopped(quint32 fid)
+{
+    if (m_waitFunction != NULL && m_waitFunction->id() == fid) {
```
mcallegari, line 245: Indentation: parenthesis on new line

engine/src/scriptrunner.h
```diff
@@ -199,24 +224,45 @@ public slots:
...
     // Queue holding Fixture values to send to Universes
     QQueue<FixtureValue> m_fixtureValueQueue;
+    // Indicate to add (true) or to not add (false) to the Functions started by this script
+    bool m_stopOnExit;
     // IDs of the Functions started by this script
     QList <quint32> m_startedFunctions;
     // Timer ticks to wait before executing the next line
     quint32 m_waitCount;
+    // Function that the script is waiting for
+    Function* m_waitFunction;
```
mcallegari, line 265: I don't like this. Instead of saving pointer to Function, just save the quint32 functionID. (and remove the include to function.h)
Then instead of using Function signals, use these instead:
https://github.com/mcallegari/qlcplus/blob/master/engine/src/mastertimer.h#L115
https://github.com/mcallegari/qlcplus/blob/master/engine/src/mastertimer.h#L118

And compare IDs rather than working on pointers

engine/src/scriptrunner.h
```diff
@@ -199,24 +224,45 @@ public slots:
...
+    /** Triggered when the script's execution pauses to await the completion of a function */
+    void slotWaitFunctionStopped(quint32 fid);
+
 protected:
     /** QThread reimplemented method */
     void run();
 
 private:
+    /** ScriptRunner function operations enum to handle start/stop/wait commands */
+    enum class SRFuncOpe : char
```
mcallegari, line 240: This is not how enums are defined in this project. See
https://github.com/mcallegari/qlcplus/blob/master/engine/src/chaser.h#L217
and please name it FunctionOperation

resources/docs/html_en_EN/scripteditor.html
```diff
@@ -231,6 +231,30 @@ <H3>Language syntax</H3>
 wait:03s.20 // Waits for 3 seconds and 200 milliseconds</PRE>
   </TD>
  </TR>
+ <TR>
```
mcallegari, line 234: Docs in HTML form are no longer bundled and maintained.
Please remove this all and submit a PR for this
https://github.com/mcallegari/qlcplus-docs/blob/main/pages/04.function-manager/08.script-editor/default.v4.md

engine/src/scriptrunner.cpp
ldebs: Fixed on commit f3d3ca0d0fae741279f2ad6434bf0eed902cd16b

engine/src/scriptrunner.cpp
ldebs: Fixed on commit f3d3ca0d0fae741279f2ad6434bf0eed902cd16b

engine/src/scriptrunner.h
ldebs: Fixed on commit f3d3ca0d0fae741279f2ad6434bf0eed902cd16b

engine/src/scriptrunner.h
ldebs: Fixed on commit f3d3ca0d0fae741279f2ad6434bf0eed902cd16b

resources/docs/html_en_EN/scripteditor.html
ldebs: Documentation moved to PR https://github.com/mcallegari/qlcplus-docs/pull/42

mcallegari:
Merged, thanks :+1:

PR #1584, opened 2024-06-13
Let cursor keys and PgUp/PgDown work as key combination on VCButtons

### LINKED ISSUE #1443: PgUp, PgDown and arrows don't work as keyboard shortcuts
**Describe the bug**
When defining keyboard shortcuts in virtual console widgets, pgup and pgdown (and arrows) don't work.
It can be usefull to use a presenter remote as remote control in a theatre play.

**To Reproduce**
1. Go to 'Virtual Console'
2. Add a button, choose any function
3. In the button properties, choose a keyboard shortcup, press "Page Down" : PgDown key is selected
4. Switch to production mode
5. Press "Page Down"
7. See error

**Expected behavior**
PgUp and PgDown sould activate a function when selected as keyboard shortcut for this function in a virtual console widget.

**Screenshots**

 - OS: Windows 10
 - QLC+ Version 4.12.7

**Additional context**
Nothing

Pac72:
As stated on #1443, PgUp, PgDown and cursor keys do not work when configured as key combination on VCButtons.
This happens because the QScrollArea hosting the controls intercepts these keys to move around the scroll area.

This change lets the mentioned keys pass through to the buttons when the user chooses them, but, of course, the keys chosen as key combinations will not be used to move around on the scroll area, since it would be way confusing to keep both functions.

Closes #1443

### DIFF
--- a/ui/src/CMakeLists.txt
+++ b/ui/src/CMakeLists.txt
@@ -141,6 +141,7 @@ add_library(${module_name}
     virtualconsole/vcmatrixproperties.cpp virtualconsole/vcmatrixproperties.h virtualconsole/vcmatrixproperties.ui
     virtualconsole/vcproperties.cpp virtualconsole/vcproperties.h virtualconsole/vcproperties.ui
     virtualconsole/vcpropertieseditor.cpp virtualconsole/vcpropertieseditor.h
+    virtualconsole/vcscrollarea.cpp virtualconsole/vcscrollarea.h
     virtualconsole/vcslider.cpp virtualconsole/vcslider.h
     virtualconsole/vcsliderproperties.cpp virtualconsole/vcsliderproperties.h virtualconsole/vcsliderproperties.ui
     virtualconsole/vcsoloframe.cpp virtualconsole/vcsoloframe.h
--- a/ui/src/src.pro
+++ b/ui/src/src.pro
@@ -155,6 +155,7 @@ HEADERS += virtualconsole/addvcbuttonmatrix.h \
            virtualconsole/vcmatrixproperties.h \
            virtualconsole/vcproperties.h \
            virtualconsole/vcpropertieseditor.h \
+           virtualconsole/vcscrollarea.h \
            virtualconsole/vcslider.h \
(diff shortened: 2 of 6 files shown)


yestalgia:
Nice work - When a build becomes available I'll give it a test.

mcallegari:
@Pac72 thanks for this.
For now I have just one comment. Please stick to the rest of the project code indentation: parenthesis go to new line.

Pac72:
@mcallegari sure, sorry for that.
Now the indentation should be ok.

qfulmina:
Hi,
thanks for your contribution (indeed one I could make use of).

I tested a bit using the generated AppImage. As you state, it (unfortunately) only works for VCButtons (yet) and not for e.g. Cue Lists.

Speaking of which, PgUp and PgDown are consumed by Cue Lists if in focus up until the upper or lower end of the list is reached. Only then PgUp and PgDown bubble up and trigger any key combination event.

I think it would be best if Cue Lists were supported as well. (For e.g. being able to trigger the next Cue with PgDown etc.)

qfulmina:
Interestingly special keys like Space, Return, PgUp/PgDown, or Numbers are treated differently.

In the scope of inputting to a Duration Widget:

 * PgUp and PgDown never bubble up (not even when reaching the end).
 * Numbers are _always_ consumed completely by the text inputs without ever bubbling up.
 * Space is _always_ consumed completly by each input.
 * Return _always_ bubbles through.

Pac72:
Hi @qfulmina

> I tested a bit using the generated AppImage. As you state, it (unfortunately) only works for VCButtons (yet) and not for e.g. Cue Lists.
@me guilty again. I did not understand the issue was general and specifically selected the VCButtons :-/

> I think it would be best if Cue Lists were supported as well. (For e.g. being able to trigger the next Cue with PgDown etc.)
I'll try to cover also the other controls.

Pac72:
Hi everyone.
I tried a different approach with #1586
If you like it better, this pull request can be closed.

Pac72:
Development is going on with #1586, we can close this.

Issue #1588, opened 2024-06-22
Linux: Art-Net IPv4 packets not received


Pac72:
**Describe the bug**
While starting to investigate on #1419 I realized I was not receiving Art-Net packets sent to my IPv4 address.
Investigating on the new issue, I saw that when binding the UDP socket with
`
udpSocket->bind(ARTNET_PORT, QUdpSocket::ShareAddress | QUdpSocket::ReuseAddressHint)
`

netstat showed only an IPv6 bind
> netstat -pnul | grep 6454
> (Not all processes could be identified, non-owned process info
>  will not be shown, you would have to be root to see it all.)
> udp6       0      0 :::6454                 :::*                                50071/qlcplus       

and strace on the process showed
> socket(AF_INET6, SOCK_DGRAM|SOCK_CLOEXEC|SOCK_NONBLOCK, IPPROTO_IP) = 4
> setsockopt(4, SOL_SOCKET, SO_BROADCAST, [1], 4) = 0
> setsockopt(4, SOL_IPV6, IPV6_RECVPKTINFO, [1], 4) = 0
> setsockopt(4, SOL_IPV6, IPV6_RECVHOPLIMIT, [1], 4) = 0
> setsockopt(4, SOL_SOCKET, SO_REUSEADDR, [1], 4) = 0
> setsockopt(4, SOL_IPV6, IPV6_V6ONLY, [0], 4) = 0
> bind(4, {sa_family=AF_INET6, sin6_port=htons(6454), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::", &sin6_addr), sin6_scope_id=0}, 28) = 0
> getsockname(4, {sa_family=AF_INET6, sin6_port=htons(6454), sin6_flowinfo=htonl(0), inet_pton(AF_INET6, "::", &sin6_addr), sin6_scope_id=0}, [28]) = 0
> getsockopt(4, SOL_IPV6, IPV6_V6ONLY, [0], [4]) = 0
> getpeername(4, 0x7ffc645dde20, [28]) = -1 ENOTCONN (Transport endpoint is not connected)
> getsockopt(4, SOL_SOCKET, SO_TYPE, [2], [4]) = 0

Replacing the bind with
`
udpSocket->bind(QHostAddress::SpecialAddress::AnyIPv4, ARTNET_PORT, QUdpSocket::ShareAddress | QUdpSocket::ReuseAddressHint)
`

solves the problem and qlcplus receives the UDP packets regularly.

[...]

PR #1589, opened 2024-06-22
plugins/artnet: force IPv4 when binding UDP socket

Pac72:
qlcplus on Ubuntu 22.04 right now when configuring ArtNet input on an IPv4 address seems not to receive UDP packets (see  #1588).
Forcing IPv4 when binding UDP socket solves the issue.

Another more complicated approach is to create a socket for each input address and bind it explicitly to the address

### DIFF
--- a/plugins/artnet/src/artnetplugin.cpp
+++ b/plugins/artnet/src/artnetplugin.cpp
@@ -412,7 +412,7 @@ QSharedPointer<QUdpSocket> ArtNetPlugin::getUdpSocket()
     udpSocket = QSharedPointer<QUdpSocket>(new QUdpSocket());
     m_udpSocket = udpSocket.toWeakRef();
 
-    if (udpSocket->bind(ARTNET_PORT, QUdpSocket::ShareAddress | QUdpSocket::ReuseAddressHint))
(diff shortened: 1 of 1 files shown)


Pac72:
Still to be checked, since qlcplus 4.13.1 on Windows 11 and qlcplus 4.12.7 on Mac seem to work out of the box.

mcallegari:
@Pac72 thanks for this but I'm afraid it won't work.

image

The first argument as SpecialAddress has been introduced in Qt 6.2. QLC+ still needs to be compatible with Qt 5.15 for a while.

Pac72:
@mcallegari I was puzzled, since I built against 5.15.2 and, as you say, the bind(QHostAddress::SpecialAddress addr,...) is not defined.
Then I realized that there is a constructor
> QHostAddress(SpecialAddress address)

so the call
> udpSocket->bind(QHostAddress::SpecialAddress::AnyIPv4, ARTNET_PORT, QUdpSocket::ShareAddress | QUdpSocket::ReuseAddressHint)

is using
> bool bind(const QHostAddress &address, quint16 port = 0, BindMode mode = DefaultForPlatform);

Anyway, this patch is not relevant anymore, since, until now, I haven't been able to reproduce the issue #1588 I opened.
Either it was some network sillyness (likely) or some strange linking issue (unlikely).

I'm giving it a last try, then I'll close the issue as PEBKAC

PR #1606, opened 2024-08-31
Add channel address in add fixture dialogue

yestalgia:
Forum thread: https://www.qlcplus.org/forum/viewtopic.php?t=17674

This pull request is made in response to the above forum thread. When adding a fixture, it can be helpful to have the address available for reference. 

Testing required:
- EFX Editor
- Scene Editor


yestalgia:
Completed UI tests on windows. Here are the results:

EFX editor:
image


VC XY Pad:
image

Scene Editor:
image

PR #1609, opened 2024-09-02
Build option+ui fix

Ledjlale:
Add a compile option to build for QLC5.
Remove qml slot depreciation.
Adapt disclaimer text size to content.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.16)
 project(qlcplus VERSION 4.13.1 LANGUAGES C CXX)
 
+option(qmlui "Build for QLC+ 5 QML UI" OFF)
+
 # Set Release build type by default
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose the type of build, options are: Debug Release RelWithDebInfo MinSizeRel." FORCE)
--- a/qmlui/qml/popup/CustomPopupDialog.qml
+++ b/qmlui/qml/popup/CustomPopupDialog.qml
@@ -95,7 +95,7 @@ Dialog
 
             contentItem.implicitHeight: UISettings.iconSizeDefault
 
-            onClicked:
+            onClicked: function(button)
             {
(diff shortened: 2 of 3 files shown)


mcallegari:
@Ledjlale thanks for splitting the PRs.
What about the cmake option? I see there is already a message printed when building with qmlui=ON
https://github.com/mcallegari/qlcplus/blob/master/CMakeLists.txt#L77

Furthermore, the documentation says the default value for boolean is OFF anyway
https://cmake.org/cmake/help/latest/command/option.html

Can you help me understand that change? thanks

Ledjlale:
The option keyword allows the IDE to know what variable to propose to the user.
For instance, I'm using QtCreator and without it, the variable is not shown in the configuration : we have to enter it manually. That means that we have to be aware of this variable outside of the configuration.

A more simple example : run cmake-gui and configure the project. The variable qmlui doesn't appear.
But if you add this option() line, you will get your variable in the configuration. Then, you don't need an external documentation/check the code for this ;)

Ledjlale:
By the way, maybe we could rename the variable to a QLC_5_BUILD or something more clear?

mcallegari:
> (and I almost post the question about how to build QLC5 on the forum, before checking the code.. so having this kind of options could save some bile :p)

Well, instructions were in the wiki actually... :wink: 
https://github.com/mcallegari/qlcplus/wiki/QLC-Plus-5-build-HOWTO

I agree to have the variable at hand, but please let's keep it "qmlui" for now. In the end v5 is the v4 engine+plugins built with a QML UI.

Please fix this build by updating the ICU version on windows, like this:
https://github.com/mcallegari/qlcplus/commit/4eaa31033687432359585d3294d7d789dc92b125

Thanks

Ledjlale:
The branch was just not rebased from the rebased [callegari:qmltoqt6](https://github.com/mcallegari/qlcplus/tree/qmltoqt6) ^^  
Or I didn't understand your request. The ICU is 75

mcallegari:
> The branch was just not rebased from the rebased [callegari:qmltoqt6](https://github.com/mcallegari/qlcplus/tree/qmltoqt6) ^^

Ok, it was probably this.
Merged then, thanks :+1:

PR #1615, opened 2024-09-14
Prevent crash with rotation stop capability on some fixtures

lilyCalla:
- Currently, activating 3d mode with fixtures that have the same value for minimum and maximum for RotationStop (for example the Robe ColorSpot 250 AT) will cause the application to crash due to division by 0
- This PR skips the calculation (which is never actually used in RotationStop capabilities) to prevent the application crash

Initial Forum Report: https://www.qlcplus.org/forum/viewtopic.php?p=74082

### DIFF
--- a/qmlui/fixtureutils.cpp
+++ b/qmlui/fixtureutils.cpp
@@ -423,7 +423,8 @@ bool FixtureUtils::goboTiming(const QLCCapability *cap, uchar value, int &speed)
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


yestalgia:
I successfully reproduced the bug you reported using the beta 3 image from qlcplus.org. 

Using your build, I confirmed that the application doesn't crash anymore. 

Thank you for the contribution.

mcallegari:
Thanks for this :+1:

PR #1617, opened 2024-09-24
ui: SequenceItem draws step note-text in ShowEditor timeline

Anarchid:
image


mcallegari:
Improved a bit and merged, thanks

PR #1636, opened 2024-11-06
Adjust CMake build behavior to match Qt project files

cmuellner:
I'm building qlcplus for Fedora using my own SPEC file, where I invoke qt-make (similar to the SPEC file in the qlcplus repo).
My assumption is, that the CMakeFile scripts should be the preferred build scripts, so there is a motivation to move there.
I tried to switch to cmake several times but always gave up as I ran into weird build issues.
This time, I decided to fix them.

I also switched my build from Qt5 to Qt6.

In total, I had to solve four issues, resulting in the four commits of this PR:
1. Hard-coded "qt5" in the PLUGINDIR path definition
2. Hard-coded Debian-style path in the LIBSDIR path definition
3. Missing build of the HTML docs
4. Installation of velleman library

One thing, that I did not fix is, that QLC+ can't be built out-of-tree because of generated header files (the include directory is relative to the source root and not the build root). I did not fix that, as it was not really a problem to do an in-tree build.

There is also a difference in the version-postfix of shared libraries (foo.so vs foo.so.1.0.0) between the CMakeFile scripts (creating foo.so) and the Qt project files (creating foo.so.1.0.0). I did not change that.


mcallegari:
Hi, thanks for this but:
- docs are no longer bundled to QLC+. Documentation is online here https://docs.qlcplus.org
- velleman plugin is being phased out, so sooner or later it will be removed from build/installation

mcallegari:
> Should I add a commit to remove doc generation from the Qt project (`resources/resources.pro` includes `SUBDIRS += docs`)?

Yes please.

> > * velleman plugin is being phased out, so sooner or later it will be removed from build/installation
> 
> I need to handle it somehow. If it is built and installed, then `rpmbuild` complains if it is not packaged. So, if I don't want it in the package, then I must prevent the build (that's what the commit does).

Up to you, but as I said, at some point I will remove it everywhere.

As for hardcoded 'qt5' paths I think there will be many more to fix. I checked also the [qmltoqt6 branch](https://github.com/mcallegari/qlcplus/tree/qmltoqt6) and I think many changes are missing there too

cmuellner:
> As for hardcoded 'qt5' paths I think there will be many more to fix. I checked also the [qmltoqt6 branch](https://github.com/mcallegari/qlcplus/tree/qmltoqt6) and I think many changes are missing there too

Most of which I have seen are gated by `appimage`.
For a UNIX/Linux build, I did not see any other `qt5` issues.
Building and running a quick test with audio and video worked fine.

cmuellner:
I've udpated the PR:
* replace the commit that adds the docs to the CMake build with a commit that removes the docs from the Qt project

variables.cmake
```diff
@@ -89,7 +89,7 @@ if (WIN32)
 elseif (APPLE)
     set(LIBSDIR "Frameworks")
 elseif (UNIX)
-    set(LIBSDIR "lib/${CMAKE_C_LIBRARY_ARCHITECTURE}")
+    set(LIBSDIR "${CMAKE_INSTALL_LIBDIR}")
```
mcallegari, line 92: I need to check if this works on the Raspberry Pi and in general on Debian package creation

resources/resources.pro
```diff
@@ -1,6 +1,6 @@
 TEMPLATE = subdirs
 
-!qmlui: SUBDIRS += docs
+!qmlui:
```
mcallegari, line 3: Please remove the whole line

variables.cmake
```diff
@@ -289,7 +289,7 @@ elseif (UNIX)
     if (appimage)
         set(PLUGINDIR "../lib/qt5/plugins/qlcplus")
```
mcallegari, line 290: I'd port the change also to this line

variables.cmake
cmuellner: On Fedora this gives `lib64`.

The CMake implementation explicitly mentions that it works different for Debian and does so here: https://gitlab.kitware.com/cmake/cmake/-/blob/v3.26.4/Modules/GNUInstallDirs.cmake?ref_type=tags#L276

But yes, testing it is the best thing to do (I don't have a RasPi, so I cannot do that).

cmuellner:
I've updated the PR as requested.

variables.cmake
mcallegari: Just checked on Ubuntu

```
---> CMAKE_C_LIBRARY_ARCHITECTURE x86_64-linux-gnu
---> CMAKE_INSTALL_LIBDIR lib/x86_64-linux-gnu
```

So this is not gonna work on Debian-like systems.
How can we solve this?

mcallegari: EDIT: just checked on the Raspberry Pi and this is the result

```
---> CMAKE_C_LIBRARY_ARCHITECTURE aarch64-linux-gnu
---> CMAKE_INSTALL_LIBDIR lib
```

So, no joy :cry:

mcallegari:
Merged, thanks :+1:

cmuellner:
> Hi @cmuellner, I'm testing building qlcplus for Fedora also. My latest Git snapshot is at https://copr.fedorainfracloud.org/coprs/ktdreyer/qlcplus/ , using the in-tree `platforms/linux/qlcplus.spec` file.

I'm using my own spec file. You can find all details here: https://copr.fedorainfracloud.org/coprs/cmuellner/qlcplus/

ktdreyer:
> We've got Fedora builds since years here
Would you please consider updating to the latest Fedora versions? I see Fedora 37 and 38 there, and those are [EOL](https://discussion.fedoraproject.org/t/fedora-linux-38-end-of-life-in-one-week/117069).

PR #1645, opened 2024-12-10
universe: clamp 16-bit values in writeRelative

mgschossmann:
A friend of mine noticed that whenever he executed a relative EFX on 16-bit PAN/TILT which goes out of range (e.g. a circle with radius 10 at PAN offset 5), the PAN/TILT faders jump to the respective other end of the range.

This seems to be a regression introduced with the 16bitfade branch, since in version 4.12.7, the values are still being clamped.

This commit fixes this issue and adds related tests.


mcallegari:
Looks good to me. Thanks a lot for spotting and fixing it :+1:

PR #1649, opened 2024-12-28
Function Wizard: create Effects

8-Lambda-8:
Hi,
I added the possibility to create Effects with the Function Wizard
What types of effects arc creates is determined by the fixture capabilities (Dimmer, RGB, Position).
for each type a normal and a staggered version (start offset spread equally) will be created.

image
image
image

### DIFF
--- a/ui/src/functionwizard.cpp
+++ b/ui/src/functionwizard.cpp
@@ -31,6 +31,7 @@
 #include "vcsoloframe.h"
 #include "vccuelist.h"
 #include "rgbmatrix.h"
+#include "efx.h"
 #include "vcwidget.h"
 #include "vcbutton.h"
 #include "vcslider.h"
@@ -328,6 +329,15 @@ void FunctionWizard::updateAvailableFunctionsTree()
                 addFunctionsGroup(fxGrpItem, grpItem,
                                   PaletteGenerator::typetoString(PaletteGenerator::Animation),
                                   PaletteGenerator::Animation);
+                addFunctionsGroup(fxGrpItem, grpItem,
+                                  PaletteGenerator::typetoString(PaletteGenerator::Effect),
+                                  PaletteGenerator::Effect);
+            }
+            else if (cap == QLCChannel::groupToString(QLCChannel::Pan))
+            {
+                addFunctionsGroup(fxGrpItem, grpItem,
+                                  PaletteGenerator::typetoString(PaletteGenerator::Effect),
+                                  PaletteGenerator::Effect);
             }
             else if (cap == QLCChannel::groupToString(QLCChannel::Gobo))
                 addFunctionsGroup(fxGrpItem, grpItem,
@@ -427,6 +437,12 @@ void FunctionWizard::updateResultFunctionsTree()
                     item->setText(KFunctionName, matrix->name());
                     item->setIcon(KFunctionName, matrix->getIcon());
                 }
+                foreach (EFX *effect, palette->effects())
+                {
+                    QTreeWidgetItem *item = new QTreeWidgetItem(getFunctionGroupItem(effect));
+                    item->setText(KFunctionName, effect->name());
+                    item->setIcon(KFunctionName, effect->getIcon());
+                }
             }
         }
     }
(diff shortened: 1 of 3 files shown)


yestalgia:
Ooh I like it

mcallegari:
Hi, this doesn't build.
Can you please fix it?

ui/src/palettegenerator.cpp
```diff
@@ -453,6 +465,65 @@ void PaletteGenerator::createRGBMatrices(QList<SceneValue> rgbMap)
     }
 }
 
+EFX *PaletteGenerator::createEffect(QList<Fixture *> fixtures, bool staggered, EFXFixture::Mode mode)
+{
+    EFX *efx = new EFX(m_doc);
+
+    QString modeString = (new QString[]{KXMLQLCEFXFixtureModePanTilt, KXMLQLCEFXFixtureModeDimmer, KXMLQLCEFXFixtureModeRGB})[mode];
```
mcallegari, line 472: This is the line that fails to build. Can you just use a QStringList?

8-Lambda-8:
i fixed it

i did not see that EFXFixture has a modeToString function

i removed the line and used this function instead

ui/src/palettegenerator.cpp
```diff
@@ -561,6 +630,15 @@ void PaletteGenerator::createFunctions(PaletteGenerator::PaletteType type,
             if (m_redList.size() > 1 && m_greenList.size() == m_redList.size() && m_blueList.size() ==  m_redList.size())
                 createRGBMatrices(m_redList);
         }
+    break;
```
mcallegari, line 633: Indentation please

ui/src/palettegenerator.cpp
```diff
@@ -102,6 +104,7 @@ QString PaletteGenerator::typetoString(PaletteGenerator::PaletteType type)
         case Gobos: return tr("Gobo macros");
         case ColourMacro: return tr("Colour macros");
         case Animation: return tr("Animations");
+        case Effect: return tr("Effects");
```
mcallegari, line 107: Why not "EFX"? Users are used to that term

8-Lambda-8: i will rename it

yestalgia:
Hey guys I've had a think about this.

Eventually I'd like to expand the EFX movement effect list. I think these should be separated in the wizard.

My suggestion would be:

- Primary Colors
- 16 Colors
- Animations
- EFX
  - Position
    - Line
    - Circle
    - etc.
  - RGB

Is there a better way for us to feed settings into the position effects like "Staggered", Split Reverse", "Asymmetrical" etc.?

8-Lambda-8:
Hi,
Yes, i will add the Modes (Pos, RGB, int) as sub items in the tree, but im not sure about the patterns. I think the list will get too long. And in my opinion you rarely want to create all of them, so you have to select/deselect to get your liked outcome. For me it would be enough to just create one and then manually duplicate and change the pattern. I see the real timesaver of this feature in generating the mode/offset/reverse settings.

For the offset/reverse settings, i could add columns with checkboxes (like odd/even)(i also have to implement odd/even)
What exactly do you mean with "Split Reverse" and "Asymetrical"? so i can add them as options

8-Lambda-8:
Hi, i added the mode sub items and the checkbox columns (not implemented the generation yet)

now it looks like this, what do you think?
image

yestalgia:
Mate we should be calling you the wizard! This is fantastic stuff!

>What exactly do you mean with "Split Reverse" and "Asymetrical"? so i can add them as options

I'm referring to the fixture order setting in the EFX:
image

No stress if this is too complex.

> im not sure about the patterns. I think the list will get too long.

Personally, I would find the patterns extremely helpful, but I understand the concern. Perhaps this would need to form part of a separate step.

8-Lambda-8:
> I'm referring to the fixture order setting in the EFX:

oh, i totally forgot that those exist, shouldn't be a problem to implement this.
and i realized that my "staggered" does the same as "Serial", so i will remove it.

also i noticed the "Serial" and "Asymmetric" are not displayed in the movement preview (all points are on the same spot), i will try to fix this and create a separate PR for that.

> Personally, I would find the patterns extremely helpful, but I understand the concern. Perhaps this would need to form part of a separate step.
okay, i'll try to do that

finnie2006:
What is the status on this pr?

8-Lambda-8:
Hi,
i still have to think about how i can add a selection of the pattern
and fix some stuff on the VC Widget creation page
and fix the tests :)

8-Lambda-8:
Hi,
i now have added the patterns like @yestalgia [suggested](https://github.com/mcallegari/qlcplus/pull/1649#issuecomment-2731365064). They are collapsed by default, so that the list is not too long. It is not that bad as i first thought it would be :)

image

yestalgia:
Well.... That's incredible work! I'll be doing some testing especially with multi-head fixtures next week.

finnie2006:
image
I also builded your fork, going to test it with my movingheads

8-Lambda-8:
i think the parts of the code that i touched conform with the code style. Did i miss anything?

ui/src/functionwizard.cpp
```diff
@@ -283,12 +287,29 @@ QList <quint32> FunctionWizard::fixtureIds() const
...
                                        QString name, PaletteGenerator::PaletteType type)
 {
     if (grpItem == NULL)
         return;
 
+    if ((type == PaletteGenerator::EfxDimmer ||
+        type == PaletteGenerator::EfxRGB ||
+        type == PaletteGenerator::EfxPosition))
+    {
+        if (efxItem == NULL){
```
mcallegari, line 302: This one

finnie2006: Is it ready to merge now?

PR #1662, opened 2025-01-10
webaccess: show page names of multi-page (solo-) frames

jpue:
Instead of the untranslated "Page: %1", the web virtual console now displays the page names that were set in the local QLC+ virtual console.
"Page: %1" is now only used as a backup and can be translated.

This PR also fixes a bug where it was not possible to change the page of a frame via the web interface.
To reproduce this issue, the frame name needs to contain a `"` or end on a `\`, as these strings were not escaped before being inserted into the JavaScript code.


mcallegari:
Thanks :+1:

PR #1663, opened 2025-01-12
Fix "undefined symbol" by removing definition

autoantwort:
There is no implementation for that definition and I get an undefined symbol error.


mcallegari:
Building on which OS?
Which Qt version?
Which compiler?

autoantwort:
macOS 15.2
Qt 6.7.2
Apple clang 16

mcallegari:
Then again this is fixed in qmltoqt6 branch
https://github.com/mcallegari/qlcplus/blob/qmltoqt6/qmlui/inputprofileeditor.h#L51

PR #1665, opened 2025-01-12
Github actions: Use cmake -S . -B build to simplify file

autoantwort:
`-S` for the source dir and `-B` for the build dir. Also uses `-DCMAKE_BUILD_TYPE` instead of `sed` to set the build type

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,20 +150,12 @@ jobs:
       - name: Configure for QT5 build
         if: ${{ matrix.task == 'compile-qt5' }}
         run: |
-          # force a release build
-          sed -i -e 's/Debug/Release/g' CMakeLists.txt
-          mkdir build
-          cd build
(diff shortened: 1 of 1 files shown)


mcallegari:
Thanks for this
I guess this should be applied also here
https://github.com/mcallegari/qlcplus/blob/master/.github/workflows/build.yml#L467
and here
https://github.com/mcallegari/qlcplus/blob/master/.github/workflows/build.yml#L475
right?

I see you use macOS. Do you have any experience in deploying on macOS and signing/stapling bundles?

autoantwort:
> right?

I don't know how I missed them... I checked all invocations of cmake 

> I see you use macOS. Do you have any experience in deploying on macOS and signing/stapling bundles?

Sadly not 

But I wanted to try to integrate sentry so that one gets crash reports with stack traces so it gets simpler to fix them.

mcallegari:
Thanks 👍

Issue #1656, opened 2025-01-03
USB DMX output timing odd behaviour on Windows


basilhussain:
I have discovered some odd behaviour with regard to timing of DMX output frames when using a USB DMX interface on Windows.

Basically, it seems the simple act of opening and closing the output plug-in configuration window - without adjusting any of the settings - ruins the timing of DMX frames being output using that interface. The break and MAB periods become excessively long (15+ ms each) and QLC+ is no longer able to output full frames of all 512 slots at the required frame frequency because there isn't enough time - the slots get cut off (i.e. <512) by the break of the subsequent frame.

Bizarrely, when this happens, system timer accuracy is reported by the plug-in to be "Good", despite that *not* actually being the case. In fact, the only time I have managed to get sensible frame timing is when accuracy is reported as "Bad". (This I believe is because under that situation, QLC+ doesn't attempt to do any timing for break or MAB, and just lets them occur as quickly as possible.)

From digging through the QLC+ source code, I can't figure out exactly the root of the problem, but I think it is probably something to do with the following:

- QThread's [`usleep`](https://doc.qt.io/qt-5/qthread.html#usleep) implementation on Windows rounds up the given `usecs` argument to the nearest millisecond. This renders QLC+'s attempt to do microsecond-resolution sleeps (using `DMX_MAB` of 16 and `DMX_BREAK` of 110 µs) pointless.
- The `usleep` functionality is delegated to Windows' [`Sleep`](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-sleep) API function, which also only deals in milliseconds. And, importantly, the minimum resolution of `Sleep` is reliant on the current resolution of the system timer.
- Unless a Windows application specifically requests a lower system timer resolution, it can often be as high as 15.625 ms. This probably explains the 15 ms break and MAB timings I have been seeing.
- If the system timer resolution being 15 ms is the cause of the problem, I am puzzled by how QLC+ determines that timing accuracy is "Good" in that situation. It seems it measures if a 1000 µs `usleep` actually takes more than 3 ms, and is "Bad" if so. But if it is taking 15 ms, how can it be less than 3 ms? Perhaps the accuracy of `QElapsedTimer` is *also* affected by system timer resolution; if so, it would not be surprising if such a test gives flawed results.
[...]

Issue #1666, opened 2025-01-12
 [Windows] Connection problems with Enttec OpenDMX USB Interface


alessandrocaseti:
**Describe the bug**
QLC has some issues in connecting my Enttec OpenDMX USB interface. If the device is plugged in, QLC won't start at all. If disconnected, it will start but as soon as I plug the cable in it immediatly crashes. This only happens in version >= 4.13.0. I haven't tried on V5 but since it isn't stable I can't use it for my current projects. I tried 4.12.7 and had no issues at all.

**To Reproduce**
1. Insert Enttec OpenDMX USB interface
2. Open QLC Version >= 4.13.0
3. Disconnect Enttec OpenDMX USB interface
4. Re-open QLC and plug in the cable again

**Temporary solution**
From the forum, I found a temporary solution:
1. Open Device Managment
2. Disable and re-enable the device under USB Controllers
3. Quickly open QLC

**Desktop (please complete the following information):**
 - OS: Windows 11 Version 23H2
 - QLC+ Version 4.13.1

**Additional context**
This bug has been reported from other users and additional context can be found here:
https://www.qlcplus.org/forum/viewtopic.php?t=17163
https://www.qlcplus.org/forum/viewtopic.php?t=17960
https://www.qlcplus.org/forum/viewtopic.php?t=17370
https://www.qlcplus.org/forum/viewtopic.php?t=18044

mcallegari:
What we know about this so far:

- Happens only on Windows
- Doesn't happen on every Windows machine
- Doesn't happen at every startup
- QLC+ 4.13.0 and higher has switched to a 64bit build. Compared to 32bit, one file needed to be added to the deploy: ftd2xx64.dll
- Some users fixed the issue by installing the FTDI from ENTTEC and removing duplicate DLL ([this post](https://www.qlcplus.org/forum/viewtopic.php?p=73737#p73737))

alessandrocaseti:
@mcallegari any news on this? Is there any chance that this issue is related to #1656 ?

mcallegari:
> [@mcallegari](https://github.com/mcallegari) any news on this? Is there any chance that this issue is related to [#1656](https://github.com/mcallegari/qlcplus/issues/1656) ?

No

leonreucher:
Same issue here - tried on two Win 11 machines, driver was reinstalled several times - but with the interface connected, QLC won't start. When the interface is connected after the program start, the software crashes.

mcallegari:
[...]

PR #1667, opened 2025-01-14
Fix for bad DMX USB plug-in timing on Windows

basilhussain:
This pull request fixes bad DMX timing from the DMX USB plug-in when running on Windows. See issue #1656 for discussion.

The following changes are implemented:

- Requests a system timer resolution of (by default) 1 millisecond during instantiation of any class that inherits from `DMXUSBWidget`. Closes that request during destruction.
- Allows a different timer resolution value to be requested via use of `dmxusb/windowstimerresolution` registry key setting. (I don't imagine much use for this, but added it just in case.)
- On Windows 11, during application initialisation, requests that the system timer resolution not be throttled whenever the app is minimised, occluded, etc.

This will resolve the following problems:

- Inconsistent DMX frame intervals for all DMX USB plug-in variants.
- For the Enttec Open DMX USB in particular:
    - System timer accuracy should now *always* be legitimately "Good".
    - Avoids high CPU usage (caused by busy-wait being used to wait for frame interval) when system timer accuracy was detected as "Bad".
    - Avoids terrible break/MAB timing and cut-off frames when timer accuracy was bad, but mistakenly detected as "Good".
- Any other plug-in variant that attempts to perform system timer accuracy detection should also now get proper "Good" results.
- On Windows 11, minimising or occluding the app window will not lead to degraded timing.

I have tested these changes on Windows 11, Windows 10, and Windows 7.

Some other notes:

- I'm not experienced with CMake, and so I don't really understand what `PUBLIC` or `PRIVATE` within `target_link_libraries` in the `CMakeLists.txt` specifies, what the difference is, or if it even matters. Looking at other areas where `winmm` is linked, there is no consistent usage - both are used.
- Casting of the return value from the call to `GetProcAddress` was triggering a compilation error, but I am puzzled as to why *only* for this code and not for any other usage of `GetProcAddress` in other code elsewhere (e.g. `peperoni`, `hid` plug-ins). Are the compilation error/warning options different for the `dmxusb` module? I had to add a `#pragma` to ignore the error, because the cast is absolutely necessary.


yestalgia:
This could be ground-breaking, thank you for taking the time.

Issue for reference: #1666

mcallegari:
Hi, thanks for this but it needs some review for the indentation. Please stick to the style of the rest of the codebase (space after if/for/while/etc statements, parenthesis on newline)
As for PUBLIC/PRIVATE here's the docs:
https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents

I don't like the pragma to ignore an errror. It needs to be fixed.
Thanks

basilhussain:
> Please stick to the style of the rest of the codebase (space after if/for/while/etc statements, parenthesis on newline)

Okay, will do.

> As for PUBLIC/PRIVATE here's the docs: https://cmake.org/cmake/help/latest/command/target_link_libraries.html#libraries-for-a-target-and-or-its-dependents

I've read those docs and I'm still not sure what the difference is. It doesn't illuminate much when I don't know what a "link interface" is. 😅 As I said before, I'm totally clueless about CMake. 🤷

To be frank, the usage of `PUBLIC` vs `PRIVATE` for linking winmm is not consistent within the qlcplus code anyway. Both are used in various places:

* [engine/src/CMakeLists.txt](https://github.com/mcallegari/qlcplus/blob/ea2900c08808b77b225cbb9831a69b1b86de9aee/engine/src/CMakeLists.txt#L126) - `PUBLIC`
* [plugins/hid/CMakeLists.txt](https://github.com/mcallegari/qlcplus/blob/ea2900c08808b77b225cbb9831a69b1b86de9aee/plugins/hid/CMakeLists.txt#L77) - `PRIVATE`
* [plugins/midi/src/win32/CMakeLists.txt](https://github.com/mcallegari/qlcplus/blob/ea2900c08808b77b225cbb9831a69b1b86de9aee/plugins/midi/src/win32/CMakeLists.txt#L49) - `PRIVATE`

So, if you could kindly recommend which I should be using, that would be best I think.

> I don't like the pragma to ignore an errror. It needs to be fixed.

Agreed. But at the time I couldn't figure out a way around it. The function pointer cast is strictly necessary, and all the information I could find basically said either "just don't use `-Wcast-function-type`" or "ignore the warning".

However, I've done some more research and it seems [one solution](https://trac.nginx.org/nginx/ticket/1865) is to first cast the pointer to `void *` before re-casting to the function type. I've tried it and it works. Will that be an acceptable solution?

mcallegari:
I'm not a cmake expert either but I think the linking should be `PRIVATE `as we don't need to expose winmm methods to other components

mcallegari:
hey @basilhussain sorry it took so long but I had to test this myself a bit as  this is quite a delicate change.
While studying the code once again I wondered if this timer resolution code can be used also for the QLC+ mastertimer. Not sure if it can be applied.
In any case, thank you very much for your contribution! 👍

PR #1668, opened 2025-01-15
Add beamZ MHL75 Hybrid Moving Head fixture definition

wiibaker:
Add beamZ MHL75 Hybrid Moving Head fixture

Created and verified to be working with the given fixture


mcallegari:
Thanks :+1:

PR #1671, opened 2025-01-19
ui/audioeditor, ui/videoeditor: fix name field not updating after changing input file

jpue:
When changing the input file/url of an existing audio or video function, the name field is not updated.
It only shows the change after closing and reopening the properties "window".

This PR fixes this by immediately setting the new value to the text field using code from the corresponding constructor.


mcallegari:
This is a wanted behavior.
If you create an Audio Function, select a source file and then name it "Entrance audio effect" you don't want it to be overwritten if you select the audio source again.

PR #1674, opened 2025-01-24
Backward-compatibility for loading QXW files with color knobs

hjtappe:
Implement backward-compatibility for loading VC animation widgets with knobs.

I noticed that qxw files created in previous versions would lose the VC animation knobs and their settings and show color buttons instead. This fixes it for v4. It seems not yet implemented in V5/qml (which prints already "Control" attribute not found / parsed).


mcallegari:
Ah yes, thanks for this :+1:

PR #1680, opened 2025-01-31
webaccess: make more strings translatable

jpue:
While working on the German translation a few weeks ago, I noticed some untranslatable strings on the webaccess pages.

This PR fixes this by enabling full translation of the configuration and authentication error page.


mcallegari:
Thanks 👍

PR #1686, opened 2025-02-26
Rename Linux desktop file to be consistent

noahgitsham:
When searching for QLC+ in my app launcher, I think to search for "QLC+", not "Q Light Controller Plus".

This change also makes it consistent with the name of the fixture editor desktop entry, which is "QLC+ Fixture Editor".

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -243,7 +243,7 @@ jobs:
           wget -c https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage -O ./appimagetool-x86_64.AppImage
           chmod a+x ./appimagetool-x86_64.AppImage
           ./appimagetool-x86_64.AppImage -v ${INSTALL_ROOT}
-          mv -v Q_Light_Controller_Plus-x86_64.AppImage qlcplus-${{env.TASK}}-${{env.APPVERSION}}-${{env.BUILD_DATE}}-${{env.GIT_REV}}.AppImage
+          mv -v Q_Light_Controller_Plus_-_QLC+-x86_64.AppImage qlcplus-${{env.TASK}}-${{env.APPVERSION}}-${{env.BUILD_DATE}}-${{env.GIT_REV}}.AppImage
 
       - name: Test Load AppImage
         if: ${{ matrix.task == 'compile-qt5qml' }}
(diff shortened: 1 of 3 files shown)


mcallegari:
I'm not sure this will work for everyone.
According to this
https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html
I think we can add also `GenericName` but I'm not sure it will be used in searches.
Can you please try?

noahgitsham:
My app launcher does search for Name and GenericName, but GenericName is more for the general use of the software, I.e. "Text Editor" or "DMX Lighting Control Software".

This entry actually already has a GenericName "Lighting control", which makes sense.

I suppose if you wanted to be able to search for both, we could change the Name to something like "QLC+ (Q Light Controller)" or vice versa.

yestalgia:
My preference would be similar to how the fixture editor is currently named:
`Q Light Controller Plus - QLC+`
And
'Fixture Definition Editor - QLC+`

Happy to test on my various machines when you guys agree on the name.

noahgitsham:
I'm curious as to why you wouldn't put "QLC+" first. That is what I thought people unanimously called this software, with even the website and repo name using it. Is it out of habit from having to type the current entry's name? I didn't even know what QLC stood for to be completely honest, and initially thought the desktop entry was missing when I couldn't search for it.

mcallegari:
> I'm curious as to why you wouldn't put "QLC+" first. That is what I thought people unanimously called this software, with even the website and repo name using it. Is it out of habit from having to type the current entry's name? I didn't even know what QLC stood for to be completely honest, and initially thought the desktop entry was missing when I couldn't search for it.

The website? You mean this website where you can clearly read "Q Light Controller+"?
image

We all write QLC+ (or qlcplus since you mentioned the repo) because "Q Light Controller+" is quite long...but it's the project original name!
For the same reason everyone says "VLC" but it would be "VideoLAN Client" (curiosity; the project started as VideoLAN Server...VLS)
image

In the end I agree with @yestalgia
TBH I wouldn't continue this discussion any longer.
In any case this PR doesn't build.

noahgitsham:
I was referring to the domain name, but apologies for not being more observant! My point still stands though that everyone calls it "QLC+" regardless, and using your example, the VLC desktop entry has the Name "VLC Media Player". I've changed them anyway.

Semantics aside, I believe the CI is failing as it is hardcoded to use the old appimage names, which are now generated with new names from the new desktop entries.

yestalgia:
Thanks for pushing the change - I'll download and test. Can you confirm this solves your search problem on Linux? I'll have a play on Windows.

mcallegari:
Thanks 👍

PR #1687, opened 2025-02-27
scripts: fix QT 6 builds

jpue:
Reported: https://www.qlcplus.org/forum/viewtopic.php?t=18149

I could reproduce the described behaviour when building QLC+ with QT 6 on an up-to-date Ubuntu 24.10 machine.

- When build with QT 6, QLC+ uses the “new” script style, based on JavaScript, while builds with QT 5 still utilize the “old” custom syntax scripts (see [here](https://github.com/mcallegari/qlcplus/blob/master/engine/src/src.pro#L191)).
- However, the `scripteditor` [always includes](https://github.com/mcallegari/qlcplus/blob/master/ui/src/scripteditor.cpp#L38) the `script.h` (old scripts), from which it uses the `QList<int> Script::syntaxErrorsLines()` function to check the script syntax (see [here](https://github.com/mcallegari/qlcplus/blob/master/ui/src/scripteditor.cpp#L471) and [here](https://github.com/mcallegari/qlcplus/blob/master/engine/src/script.h#L117)).
That is why, when compiling the `ScriptEditor` class, the compiler assumes getting a `QList<int>` from that function call, but later, it is linked against `QStringList Script::syntaxErrorsLines()` from `scriptv4.cpp` (see [here](https://github.com/mcallegari/qlcplus/blob/master/engine/src/scriptv4.cpp#L225)), resulting in strange output or crashes when [its values are later used as an index into another list](https://github.com/mcallegari/qlcplus/blob/master/ui/src/scripteditor.cpp#L481).
- When opening a workspace created by a QLC+ version built with QT 5, scripts are [automatically converted into the new syntax](https://github.com/mcallegari/qlcplus/blob/master/engine/src/scriptv4.cpp#L392). However, the conversion of the `setBlackout` command seems to be wrong:
is: `blackout:[on/off]` -> `Engine.setBlackout([on/off]);`
should: `blackout:[on/off]` -> `Engine.setBlackout([true/false]);`.
- When adding commands via the UI buttons to a script in a QT 6 based QLC+ build, the old syntax is used. Also, the `waitkey`, `sethtp` and `setltp` commands have been ignored as they had been marked `not supported yet` ever since the initial import of QLC+ and because they have no equivalent implementation in the “new” script syntax.
- It seems as if the “new” scripts need to have a newline at the end in order to run (correctly). Since I did not want to introduce code modifying existing workspaces, the corresponding commit always simply appends a newline on the fly to given script code before execution. Therefore, it is guaranteed that a script can run, even if it does not end with a newline.
- [TODO] In QLC+ versions built with QT 5, scripts terminate when all of their commands have been executed, releasing all of the channel “overrides”. On QT 6 builds, a script just stays on the last command until it is manually stopped.

### DIFF
--- a/engine/src/scriptrunner.cpp
+++ b/engine/src/scriptrunner.cpp
@@ -35,7 +35,7 @@
 ScriptRunner::ScriptRunner(Doc *doc, QString &content, QObject *parent)
     : QThread(parent)
     , m_doc(doc)
-    , m_content(content)
+    , m_content(content + "\n")
     , m_running(false)
     , m_engine(NULL)
     , m_stopOnExit(true)
--- a/engine/src/scriptv4.cpp
+++ b/engine/src/scriptv4.cpp
@@ -522,6 +522,22 @@ QString Script::convertLine(const QString& str, bool *ok)
                     value.append("\"");
                 }
             }
+            else if (command == blackoutLegacy)
+            {
+                if (value == "on")
+                {
+                    value = "true";
+                }
+                else if (value == "off")
+                {
+                    value = "false";
+                }
+                else
+                {
+                    value.prepend("\"");
+                    value.append("\"");
+                }
+            }
 
             values << value.trimmed();
         }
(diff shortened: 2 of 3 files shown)


mcallegari:
QLC+ version 4 must use the old syntax
JavaScript Scripts are for version 5 only

mcallegari:
This is why it would be a good practice to discuss changes in the forums before starting to modify code.
As I said, I'll have at this later today. Might be a trivial build change.

PR #1697, opened 2025-03-22
bugfix: Have Plasma RGB Script remember its colors

Jurrie:
By removing the rgbMapGetColors function, the colors are not overwritten by the default values. This fix is based on how balls.js works.

The bug is described in this post: https://forum.qlcplus.org/viewtopic.php?p=77014.

### DIFF
--- a/resources/rgbscripts/plasma.js
+++ b/resources/rgbscripts/plasma.js
@@ -281,11 +281,6 @@ var testAlgo;
       util.initialized = false;
     }
 
-    algo.rgbMapGetColors = function()
-    {
-      return util.colorArray;
-    }
-
     algo.rgbMap = function(width, height, rgb, step)
     {
       if (util.initialized === false)


hjtappe:
I think, the whole idea was that the "default" colors are set externally when defining or modifying (programmatically) the RGBmatrix.

Can you please elaborate more on the scenario where the colors are overwritten and trhey shouldn't?
Which version of QLCplus does your scenario refer to?

The idea in these API version 3 scripts is that when initiating the RGBscript function, the UI gets the default colors from the script into the matrix effect, but from then onwards, the colors are maintained in the matrix effect and no more in the script.
If that resets the colors anywhere, the fix needs to be applied in the script CPP code, not in the matrix script.

Jurrie:
I agree! I know my "solution" works, and I see it works the same as balls.js, but something about removing a valid function doesn't feel right 😀 The functionality you describe is more logical - that should be the desired functionality.

I actually tried plowing through the CPP code for a short while, but failed to find the bug.

Regarding your questions on elaborating; I tried explaining the bug in https://forum.qlcplus.org/viewtopic.php?p=77014:

> Preview seems to display these colors. Deselect your RGB matrix (for example, by clicking on another RGB matrix in the left panel), and select it again. The colors are back to the default.
> I believe this is a bug, and I think it applies only to the "Plasma" RGB matrix. I don't see the bug with the "Balls" RGB matrix.
> I am running QLC+ 4.14.1 (from the repository at [http://download.opensuse.org/repositori ... ntu_22.04/](http://download.opensuse.org/repositories/home:/mcallegari79/xUbuntu_22.04/)) on Ubuntu Linux 24.04.2 LTS on x64.

But I think a screen recording explains more than a thousand words: [QLC+ PR1697.webm](https://github.com/user-attachments/assets/7f285c81-3cca-45e0-b95d-f70229de19e4)

Let me know if you need more info from me.

mcallegari:
Let's close this and progress in #1727

PR #1698, opened 2025-03-23
feature: Add favicons to web access

Jurrie:
This includes Apple touch icons (which are used when bookmarking the webpage to a "desktop" launcher on Android / iPhone.

This is generated with the help of https://favicon.io/favicon-converter/.

Example on Android with Chrome:
image


mcallegari:
Hi, thanks for this, but it looks to me very Apple-oriented... How about the rest of the world/browsers?
For example, the QLC+ website has just this line
`<link rel="icon" type="image/png" href="/user/themes/qlcplus/images/favicon.png" />`
Isn't that enough for everyone?

Jurrie:
> Hi, thanks for this, but it looks to me very Apple-oriented... How about the rest of the world/browsers? For example, the QLC+ website has just this line `<link rel="icon" type="image/png" href="/user/themes/qlcplus/images/favicon.png" />` Isn't that enough for everyone?

Hi Massimo,

I assume you mean the line `<link rel="apple-touch-icon" ...`? Because other lines don't seem to hint at Apple? Indeed: that line is for Apple support. The other lines are for other things like Chrome. See the attached screenshot in the PR description for an example on how it looks on Android (emulator) using Chrome.

mcallegari:
Ok, I got tricked by the type of icon you used.
The square-ish one with gray background is only used for the Apple bundle.
The official QLC+ logo has transparent background
https://github.com/mcallegari/qlcplus/blob/master/resources/icons/svg/qlcplus.svg
and that should be used for everything.

Question: why are several sizes needed? Are they mandatory on mobile OSes or can we just say 128x128px is good for all scaling and provide just one? (sorry but I hate to add new stuff especially if it is redundant)

Jurrie:
> The official QLC+ logo has transparent background [`master`/resources/icons/svg/qlcplus.svg](https://github.com/mcallegari/qlcplus/blob/master/resources/icons/svg/qlcplus.svg?rgh-link-date=2025-03-23T14%3A35%3A41Z) and that should be used for everything.

Right! I changed the logos to the official QLC+ logo.

> Question: why are several sizes needed? Are they mandatory on mobile OSes or can we just say 128x128px is good for all scaling and provide just one? (sorry but I hate to add new stuff especially if it is redundant)

I am no expert in this; I just used what https://favicon.io/favicon-converter/ generated for me 😀 It seems to generate only the common sizes. There are a lot more that one can include if I may believe [this favicon cheat sheet](https://github.com/audreyfeldroy/favicon-cheat-sheet). That cheat sheet also states that one should at least include:

- favicon.ico with 16x16 and 32x32 image (I checked: these are in there, as wel as an 48x48 size)
- favicon-180x180.png

The 192x192 and 512x52 sizes are used in site.webmanifest. This is [a specification](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest) that is mainly used for progressive web applications (a.k.a. "websites disguised as an application"). You could very well argue that QLC+ Webaccess is a PWA.

So the files that Favicon.io includes, but might be left out are:
- favicon-16x16.png
- favicon-32x32.png

They are included by Favicon.io because browsers sometimes scale down the 180x180 size very poorly. I have removed these two sizes. So as it looks to me, all files that are left are mandatory.

yestalgia:
You're a legend this has bugged me for a while.

mcallegari:
This is still not OK.
PNG pictures are blurry as if they were upscaled from a lower resolution.
Again, please use the official SVG logo to produce good quality PNGs.
If you're not familiar with tools like GIMP or Photoshop, please leave this to me.
Also, please remove the 180x180 and 512x512 versions.
I'd say 192x192 can cover all the needs. (in case, we could provide a 256x256)
Apple and friends will downscale the icon anyway to the target display size.

mcallegari:
Also, QLC+ is not a PWA application (those are node-js based or React native, maybe Springboot, stuff like that) so I even doubt the web manifest is needed in this case.

Jurrie:
> PNG pictures are blurry as if they were upscaled from a lower resolution.
> Again, please use the official SVG logo to produce good quality PNGs.

They _are_ blurry, aren't they? I did use the official SVG logo with Favicon.io. I dismissed it as the GitHub preview mangling the images, but I guess that Favicon.io does a very bad job of converting SVG to PNG. I manually converted the images and they look crisp now.

> Also, please remove the 180x180 and 512x512 versions.
> I'd say 192x192 can cover all the needs. (in case, we could provide a 256x256)
> Apple and friends will downscale the icon anyway to the target display size.

Just to let the record show: I think you are wrong. But :shrug: - QLC+ is not my project, and my opinion doesn't matter in this. I removed them as you wish.

> Also, QLC+ is not a PWA application (those are node-js based or React native, maybe Springboot, stuff like that) so I even doubt the web manifest is needed in this case.

No, [Spring Boot](https://spring.io/projects/spring-boot) is used for regular web applications, microservices, console apps, stuff like that.

[Mozilla defines a PWA](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Guides/What_is_a_progressive_web_app) as "...an app that's built using web platform technologies, but that provides a user experience like that of a platform-specific app." In that sense one could see QLC+ Webaccess as a PWA. Sure, things like service workers are not used, so it wouldn't be the textbook example of a PWA.

In that sense I think a web application manifest is appropriate here. But again: QLC+ is not my project, and my opinion doesn't matter. I removed it as you wish. (Your last comment was not 100% clear to me on this, but I interpreted it as a request to remove them.)

I force-pushed my changes - binary files and Git history you know 😀

mcallegari:
Thanks :+1:

PR #1700, opened 2025-03-25
EFX preview improvements

8-Lambda-8:
Hi,
as mentionend in a comment of [PR 1649](https://github.com/mcallegari/qlcplus/pull/1649#issuecomment-2748251130) "Serial" and "Asymmetric" Fixture orders are not displayed in the EFX preview.
I fixed this.

additionally i improved the visbility of the fixture points by:
- fixing usage of qt text color (problem on dark themes)
- adding Antialiasing
- improved text centering 
- bigger font.
- increased preview step count from 128.0 to 512.0 for smoother animation

old:
image
old on dark theme: (Breeze Dark KDE system theme)
image

new: (light and dark)
image


8-Lambda-8:
sry, i have to fix the test cases,
i always forget to run `make check`

mcallegari:
Hi, thanks for this.
I might be wrong, but IIRC changing the number of preview steps is wrong.
It means the preview will not be consistent with the actual EFX output on the DMX line.
I need to check with a moving head.
Maybe with non-pan/tilt this is OK.

8-Lambda-8:
Hi,
I tested if the preview and the actual Output are in sync, by comparing the colors of RGB fixtures int the fixture monitor with the Color Background of the preview. Those matched.
image

But i only tested RGB and only with the fixture Monitor. I didn't think about pan/tilt could behave different, but i also will test this, and with real fixtures.

mcallegari:
Looks good on moving heads too. Thank you for this

PR #1703, opened 2025-03-26
Various small things in RGB script devtool

Jurrie:
While working with the RGB script devtool I cam across a couple of tiny things that I thought could be improved. Nothing major, just basic stuff.

- Every form control now has a label (to comply with [WCAG 2.0 H44](https://www.w3.org/TR/WCAG20-TECHS/H44.html))
- The buttons now have nice UTF-8 symbols
- Some tables had invalid columns
- There were some typo's

### DIFF
--- a/resources/rgbscripts/devtool.html
+++ b/resources/rgbscripts/devtool.html
@@ -37,6 +37,11 @@
   input[type=button]:hover {
     background-color: #8db8dd;
   }
+  table:not(#map), table input[type=text], table input[type=number], table input[type=color], table select {
+    width: 100%;
+    min-width: 75px;
+    box-sizing: border-box;
+  }
 </style>
 <script type="text/javascript" src="devtool/devtool.js"></script>
 </head>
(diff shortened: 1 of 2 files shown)


resources/rgbscripts/devtool.html
```diff
@@ -126,7 +131,7 @@ <h2>Grid Size</h2>
 
 <div class="box">
 <form id="pixelColorChooser">
-<h2>Pixel colors</h2>
+<h2>Pixel Colors</h2>
```
yestalgia, line 134: QLC+ doesn't have a style guide for sentence/title case. Back when I was in University we learned APA style, but keeping sentence case seems to be consensus generally:

Let's keep titles in sentence case. I'll review the documentation for this as well. 

https://learn.microsoft.com/en-us/style-guide/capitalization
https://apastyle.apa.org/style-grammar-guidelines/capitalization/title-case
https://libguides.jcu.edu.au/HarvardAGSM/titles

yestalgia (changes requested):
image

Apart from my grammar suggestion I think we're good :)

mcallegari:
Thanks :+1:

PR #1708, opened 2025-04-06
simplify work with containers

shaforostoff:
also do not create temporary keys container,
iterate directly over map

### DIFF
--- a/engine/src/qlcfixturemode.cpp
+++ b/engine/src/qlcfixturemode.cpp
@@ -64,13 +64,7 @@ QLCFixtureMode& QLCFixtureMode::operator=(const QLCFixtureMode& mode)
         m_heads = mode.m_heads;
         m_masterIntensityChannel = QLCChannel::invalid();
 
-        m_actsOnMap.clear();
-        QMapIterator<quint32, quint32> ait(mode.m_actsOnMap);
-        while (ait.hasNext())
-        {
-            ait.next();
-            m_actsOnMap.insert(ait.key(), ait.value());
-        }
+        m_actsOnMap = mode.m_actsOnMap;
 
         /* Clear the existing list of channels */
         m_channels.clear();
--- a/engine/src/universe.cpp
+++ b/engine/src/universe.cpp
@@ -216,24 +216,19 @@ QSharedPointer<GenericFader> Universe::requestFader(Universe::FaderPriority prio
 
     {
         QMutexLocker fadersLocker(&m_fadersMutex);
-        if (m_faders.isEmpty())
-        {
-            m_faders.append(fader);
-        }
-        else
         {
             for (int i = m_faders.count() - 1; i >= 0; i--)
             {
-                QSharedPointer<GenericFader> f = m_faders.at(i);
+                const QSharedPointer<GenericFader>& f = m_faders.at(i);
                 if (!f.isNull() && f->priority() <= fader->priority())
                 {
                     insertPos = i + 1;
                     break;
                 }
             }
-
-            m_faders.insert(insertPos, fader);
         }
+
+        m_faders.insert(insertPos, fader);
 
         qDebug() << "[Universe]" << id() << ": Generic fader with priority" << fader->priority()
                  << "registered at pos" << insertPos << ", count" << m_faders.count();
@@ -255,11 +250,11 @@ void Universe::dismissFader(QSharedPointer<GenericFader> fader)
 void Universe::requestFaderPriority(QSharedPointer<GenericFader> fader, Universe::FaderPriority priority)
 {
     QMutexLocker fadersLocker(&m_fadersMutex);
-    if (m_faders.contains(fader) == false)
-        return;
-
     int pos = m_faders.indexOf(fader);
     int newPos = 0;
+
+    if (pos == -1)
+        return;
 
     for (int i = m_faders.count() - 1; i >= 0; i--)
     {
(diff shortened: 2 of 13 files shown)


engine/src/qlcfixturemode.cpp
```diff
@@ -264,11 +258,10 @@ quint32 QLCFixtureMode::channelActsOn(quint32 chIndex)
 
 void QLCFixtureMode::setChannelActsOn(quint32 chIndex, quint32 actsOnIndex)
 {
-    if (m_actsOnMap.contains(chIndex))
-        m_actsOnMap.remove(chIndex);
-
     if (actsOnIndex != QLCChannel::invalid())
         m_actsOnMap[chIndex] = actsOnIndex;
+    else
+        m_actsOnMap.remove(chIndex);
```
shaforostoff, line 264: this reduces number of map iterations

engine/src/qlcfixturemode.h
```diff
@@ -320,7 +320,6 @@ class QLCFixtureMode
 
     /** Save a mode to an XML document */
     bool saveXML(QXmlStreamWriter *doc);
-    QHash<QLCChannel *, QLCChannel *> actsOnChannelsList() const;
```
shaforostoff, line 323: there is no definition of this method in cpp file

engine/src/fixture.cpp
```diff
@@ -1428,7 +1425,6 @@ bool Fixture::saveXML(QXmlStreamWriter *doc) const
         doc->writeTextElement(KXMLFixtureForcedLTP, list);
     }
 
-    if (m_channelModifiers.isEmpty() == false)
```
shaforostoff, line 1431: no need to check if the map is empty. If it is empty, the loop below will have zero iterations, so de facto isEmpty check is being made by it.hasNext()

engine/src/universe.cpp
```diff
@@ -216,25 +216,18 @@ QSharedPointer<GenericFader> Universe::requestFader(Universe::FaderPriority prio
 
     {
         QMutexLocker fadersLocker(&m_fadersMutex);
-        if (m_faders.isEmpty())
+        for (int i = m_faders.count() - 1; i >= 0; i--)
         {
-            m_faders.append(fader);
```
shaforostoff, line 221: no need to this special case: If  m_faders.count()==0, then the for loop will have zero iterations, and insertPos will remain 0. insert(0, fader) in this case is an equivalent of both append and prepend, according to Qt documentation

mcallegari:
Hi, thanks for this but please don't use the "auto" type. It's C++11 and the QLC+ engine still preserve older GCCs compatibility.
This is not the case for the QMLUI code path.
When v5 will become the default, the engine can be ported to more modern C++ standard.
Thanks

shaforostoff:
I have pushed the change replacing auto with real types. BTW, it would be nice if macOS builds supported older macOS versions, ideally macOS 12 Monterey. Right now I have to start the app via commandline: /Applications/QLC+.app/Contents/MacOS/qlcplus

engine/src/qlcfixturemode.cpp
```diff
@@ -64,13 +64,7 @@ QLCFixtureMode& QLCFixtureMode::operator=(const QLCFixtureMode& mode)
...
         m_masterIntensityChannel = QLCChannel::invalid();
 
-        m_actsOnMap.clear();
-        QMapIterator<quint32, quint32> ait(mode.m_actsOnMap);
-        while (ait.hasNext())
-        {
-            ait.next();
-            m_actsOnMap.insert(ait.key(), ait.value());
-        }
+        m_actsOnMap = mode.m_actsOnMap;
```
mcallegari, line 67: Are you sure this is a deep copy like the previous code?

shaforostoff: QMap is implicitly shared. The underlying data will be deep-copied the next time either of the two is modified (strictly speaking, when non-const member is invoked on either of them).

So if you know you want to force deep copy at this point (instead of later when/if it will be actually needed), I can add m_actsOnMap.begin() call after this line. Since m_actsOnMap is non-const, a non-const begin() will be invoked which will in turn call detach() internally which will lead to a forced deep copy.

In general, I would like to convert all QMap and QHash uses in the app to std::flat_map after you allow engine code to use c++23 features. This will reduce memory fragmentation and hence its usage, and will speed up map/hash access as well. std::flat_map is not implicitly shared and a deep copy will happen already during the assignment.

shaforostoff:
I have rebased and added simplification in In Function::requestAttributeOverride which reduces complexity from O(n*log_n) to O(n)

Before my change it was doing new map lookup for each key of this map

mcallegari:
> Also, may I ask why c++99 standard compatibility has to be preserved, given that the application uses Qt5? Which QLC+ target environments are not supporting at least c++17?

QLC+ is built on a wide variety of platforms and OSes. Starting from OBS, where old OSes has been kept for a long time, the Raspberry Pi, macOS, etc..
As of today, it should be safe to use C++11 and perhaps C++17 features.
However, PLEASE don't use the `auto` type! I hate `auto` and it feels so Java-ish.
Let's use it only where it makes life easier like replacing something like `QMap<QVariant, QPair<QString, QByteArray>>`
Also, I believe `auto` makes quick code checks from the GitHub web pages (I do this a lot) harder as you always have to remember the type of a variable or jump to the header file to find out.
Agree? :slightly_smiling_face:

mcallegari:
> I don't know much about OBS - could you please provide me with some links?

OBS stands for OpenSUSE Build Service.
https://build.opensuse.org/project/show/home:mcallegari79

It is a nice way to automate builds for several Linux distro at once.

shaforostoff:
I have split the most simple changes into https://github.com/mcallegari/qlcplus/pull/1726

Now this PR is smaller in size

mcallegari:
Thanks :+1:

PR #1716, opened 2025-04-11
Refresh Grand Master slider mode on new/load document

Jurrie:
The slider mode is stored in the XML file, but is not refreshed when a document is loaded (or a new document is created). This PR fixes that.


mcallegari:
Ah yes, clearly a typo :)

PR #1720, opened 2025-04-14
Update Eurolite-LED-KLS-2500.qxf

mitscherdinger:
This time controlled by the "fixture validater": Made use of aliases, added colors wherever possible

### DIFF
--- a/resources/fixtures/Eurolite/Eurolite-LED-KLS-2500.qxf
+++ b/resources/fixtures/Eurolite/Eurolite-LED-KLS-2500.qxf
@@ -3,7 +3,7 @@
 <FixtureDefinition xmlns="http://www.qlcplus.org/FixtureDefinition">
  <Creator>
   <Name>Q Light Controller Plus</Name>
-  <Version>4.12.1 GIT</Version>
+  <Version>4.14.1</Version>
   <Author>Mitsch</Author>
  </Creator>
  <Manufacturer>Eurolite</Manufacturer>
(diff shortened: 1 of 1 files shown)


mcallegari:
This is not OK.
Please use the [online validator](https://www.qlcplus.org/fixture_validator.php) or see the error logs

mitscherdinger:
Huh!
I thought I already took "white" (#ffffff) in the fixture editor…
Guess, your critic is also true for the following lines?

  <Capability Min="208" Max="215" Preset="ColorMacro">RGBA</Capability>
  <Capability Min="216" Max="223" Preset="ColorMacro">RGBW</Capability>
  <Capability Min="224" Max="231" Preset="ColorMacro">RGAW</Capability>
  <Capability Min="232" Max="239" Preset="ColorMacro">RBAW</Capability>
  <Capability Min="240" Max="247" Preset="ColorMacro">GBAW</Capability>
  <Capability Min="248" Max="255" Preset="ColorMacro">RGBAW</Capability>
  
  These are just shades of white, so, would you recommend to use #ffffff there, too?

mcallegari:
> Guess, your critic is also true for the following lines?

Yes

> These are just shades of white, so, would you recommend to use #ffffff there, too?

I usually use "ColorMacro" or "ColorDoubleMacro" presets only when the color can be represented. 
In those case I would remove the preset completely. Thanks

mitscherdinger:
> Why every mode has physical dimensions set?
I don't know. Maybe I wanted to be 105% correct…

> Can they just use global properties?
Well, I guess. Shall I change it and send another patch?

mcallegari:
Yes, please update this PR to use only one global physical properties. Thanks

mitscherdinger:
Done. Hopefully… Don't know, if it is comitted already? Git is new to me…

mcallegari:
I think you haven't pushed the latest changes. I still see multiple physical properties

PR #1723, opened 2025-04-16
Handle the case when `make_chip_iter()` throws exception

Jurrie:
I use Podman+QEmu to test my Raspberry Pi QLC+ images on my development machine. It's a much faster development cycle. But there are no GPIO chips on my development machine, and the Raspberry Pi QLC+ image obviously has that plugin enabled. Previously I resorted to a hacky "delete GPIO plugin when running inside QEmu"-script, but this PR ensures that the situation is handled gracefully.

For anyone interested in running the Raspberry Pi QLC+ image on a non-arm64 Debian machine, the global steps are:
- `sudo apt install podman qemu-user-static`
- Extract rootfs from QLC+ `.img` file to `somedir/root`
- `sudo podman run -it --rootfs somedir/root bash`

### DIFF
--- a/plugins/gpio/gpioplugin.cpp
+++ b/plugins/gpio/gpioplugin.cpp
@@ -58,15 +58,23 @@ void GPIOPlugin::init()
     else
     {
         // autodetect chips and use first
-        for (auto& it: ::gpiod::make_chip_iter())
+        try
         {
-            qDebug() << "GPIO chip found" << QString::fromStdString(it.name());
-            if (m_chipName.empty())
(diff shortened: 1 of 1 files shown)


mcallegari:
Kind of a niche case but OK, can help

Jurrie:
> Kind of a niche case but OK, can help

Yes, maybe my case is a bit niche 😀 But I should have mentioned that this also reports things like "permission denied" when opening the GPIO device. That might benefit other users and is likely less niche. Thanks for merging!

PR #1724, opened 2025-04-16
Save XML tags in determinicstic order (better when using a VCS)

Jurrie:
Use QMap instead of QHash while saving XML. This way the tags in the resulting XML are always in the same order. When storing XML files in a version control system such as Git, this circumvents false differences between revisions.


mcallegari:
Thanks 👍

yestalgia:
Great work

PR #1725, opened 2025-04-16
Correctly name 'SonicPulse LED Bar 10'

Jurrie:
This fixes the warning `bool
QLCFixtureDefCache::addFixtureDef(QLCFixtureDef*) Cache already contains "Stairville SonicPulse LED Bar 05"` and also shows the SonicPulse LED Bar 10 in the fixtures list when adding a fixture.

The fixture in question is [this one](https://www.thomann.de/intl/stairville_sonicpulse_led_bar_10.htm).


mcallegari:
Ah nice catch! Thanks 👍

PR #1726, opened 2025-04-16
Speed up many foreach-based iterations over QMap

shaforostoff:
foreach(X x, map.values()) is semantically the same as foreach(X x, map) but does not create temporary list containter

from https://doc.qt.io/qt-6/foreach-keyword.html : With QMap and QHash, foreach accesses the value component of the (key, value) pairs automatically,
so you should not call values() on the container
(it would generate an unnecessary copy, see below).

also optimize few other obvious things:
use value(key, defaultValue) instead of calling contains() this way map is iterated just once (not twice)

call removeAll without calling contains() to reduce number of map iterations from 2 to 1. if map doesn't contain the key, removeAll() does not crash or puts everything on fire, it only iterates through the map, just like contains() would do

### DIFF
--- a/engine/audio/src/audioplugincache.cpp
+++ b/engine/audio/src/audioplugincache.cpp
@@ -103,7 +103,7 @@ void AudioPluginCache::load(const QDir &dir)
 QStringList AudioPluginCache::getSupportedFormats()
 {
     QStringList caps;
-    foreach (QString path, m_pluginsMap.values())
+    foreach (QString path, m_pluginsMap)
     {
         QPluginLoader loader(path, this);
         AudioDecoder* ptr = qobject_cast<AudioDecoder*> (loader.instance());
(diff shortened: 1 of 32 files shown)


shaforostoff:
Sorry, CI fails because of networking errors (i.e. not my fault). I will force-push tomorrow again to see if this had been resolved

mcallegari:
Thanks. I was wondering since we agreed to use C++11 features...
In the v5 code I started to use 
`for (var &foo : container) `
rather than foreach. 
Would it bring any optimization in the engine in your opinion?

shaforostoff:
> `for (var &foo : container) ` rather than foreach. Would it bring any optimization in the engine in your opinion?

It won't bring any optimization iterating over Q* containers (unless the loop modifies the container or foreach is done over container.keys() or container.values()), but will make the code shorter and easier to change type of the container in the future.

The new syntax allows modifying the items in the container (foreach does not because it creates a copy of the container; since Q* containers are implicitly shared, the copy is cheap and no deep copy is made provided the container is not changed within the loop)

mcallegari:
Merging, thanks! :+1:

shaforostoff:
> Merging, thanks! 👍

thanks! Shall I switch https://github.com/mcallegari/qlcplus/pull/1708 to use **auto** to make it shorter?
Or maybe any other adjustments?

PR #1727, opened 2025-04-16
Remember colors that are set on RGB matrix

Jurrie:
This fixes the bug described in [this forum post](https://forum.qlcplus.org/viewtopic.php?p=77014) and also shown in [this video](https://github.com/user-attachments/assets/7f285c81-3cca-45e0-b95d-f70229de19e4). This PR supersedes #1697 with a better fix.

To be honest, I do not fully understand why the code was there. And because of that, I do not fully understand what I might have broken by removing the lines. It code block _seems_ illogical in that place, and it fixes the issue, but more testing should be done before merging I think... Only... What to test and how to test? 😀 I tested with `Plasma` (where the issue surfaced), `Balls` and `Even/odd` (which doesn't use the new 5-color-max feature). I did not notice something breaking.

### DIFF
--- a/engine/src/rgbmatrix.cpp
+++ b/engine/src/rgbmatrix.cpp
@@ -242,6 +242,10 @@ void RGBMatrix::setAlgorithm(RGBAlgorithm *algo)
                     m_properties.take(it.key());
                 }
             }
+
+            QVector<uint> colors = script->rgbMapGetColors();
+            for (int i = 0; i < colors.count(); i++)
+                m_rgbColors.replace(i, QColor::fromRgb(colors.at(i)));
         }
     }
     m_stepsCount = algorithmStepsCount();
@@ -342,7 +346,8 @@ QVector<QColor> RGBMatrix::getColors() const
 
 void RGBMatrix::updateColorDelta()
 {
-    m_stepHandler->calculateColorDelta(m_rgbColors[0], m_rgbColors[1], m_algorithm);
+    if (m_rgbColors.count() > 1)
+        m_stepHandler->calculateColorDelta(m_rgbColors[0], m_rgbColors[1], m_algorithm);
 }
 
 void RGBMatrix::setMapColors(RGBAlgorithm *algorithm)
@@ -357,17 +362,23 @@ void RGBMatrix::setMapColors(RGBAlgorithm *algorithm)
     if (m_group == NULL)
         m_group = doc()->fixtureGroup(fixtureGroup());
 
-    if (m_group != NULL)
+    QVector<unsigned int> rawColors;
+    const int acceptColors = algorithm->acceptColors();
+    rawColors.reserve(acceptColors);
+    for (int i = 0; i < acceptColors; i++)
     {
-        QVector<unsigned int> rawColors;
-        for (int i = 0; i < algorithm->acceptColors(); i++)
+        if (m_rgbColors.count() > i)
         {
             QColor col = m_rgbColors.at(i);
             rawColors.append(col.isValid() ? col.rgb() : 0);
         }
-
-        algorithm->rgbMapSetColors(rawColors);
+        else
+        {
+            rawColors.append(0);
+        }
     }
+
+    algorithm->rgbMapSetColors(rawColors);
 }
 
 /************************************************************************
--- a/resources/rgbscripts/plasma.js
+++ b/resources/rgbscripts/plasma.js
@@ -27,15 +27,15 @@ var testAlgo;
     algo.apiVersion = 3;
     algo.name = "Plasma";
     algo.author = "Tim Cullingworth, Massimo Callegari";
-    algo.acceptColors = 0;
+    algo.acceptColors = 5;
     algo.properties = new Array();
     algo.rstepcount = 0;
     algo.gstepcount = 50;
     algo.bstepcount = 100;
-    algo.presetIndex = 0;
+    algo.presetIndex = 4;
     algo.properties.push(
       "name:presetIndex|type:list|display:Preset|" +
-      "values:Rainbow,Fire,Abstract,Ocean,User Defined|" +
+      "values:User Defined,Rainbow,Fire,Abstract,Ocean|" +
       "write:setPreset|read:getPreset");
     algo.presetSize = 5;
     algo.properties.push(
@@ -61,30 +61,25 @@ var testAlgo;
       if (_preset === "Rainbow")
       {
         algo.presetIndex = 0;
-        util.colorArray = [ 0xFF0000, 0x00FF00, 0x0000FF ];
       }
       else if (_preset === "Fire")
       {
         algo.presetIndex = 1;
-        util.colorArray = [ 0xFFFF00, 0xFF0000, 0x000040, 0xFF0000 ];
       }
       else if (_preset === "Abstract")
       {
         algo.presetIndex = 2;
-        util.colorArray = [ 0x5571FF, 0x00FFFF, 0xFF00FF, 0xFFFF00 ];
       }
       else if (_preset === "Ocean")
       {
         algo.presetIndex = 3;
-        util.colorArray = [ 0x003AB9, 0x02EAFF ];
       }
       else if (_preset === "User Defined")
       {
         algo.presetIndex = 4;
         algo.acceptColors = 5;
-        util.colorArray = [ 0x00FF00, 0xFFAA00, 0x0000FF, 0xFFFF00, 0xFFFFFF ];
       }
-      else { algo.presetIndex = 0; }
+      else { algo.presetIndex = 4; }
       util.initialized = false;
     };
 
@@ -95,7 +90,7 @@ var testAlgo;
       else if (algo.presetIndex === 2) { return "Abstract"; }
       else if (algo.presetIndex === 3) { return "Ocean"; }
       else if (algo.presetIndex === 4) { return "User Defined"; }
-      else { return "Rainbow"; }
+      else { return "User Defined"; }
     };
 
     algo.setSize = function(_size)
@@ -137,10 +132,11 @@ var testAlgo;
       // with the given width
       var gradIdx = 0;
       util.gradientData = new Array();
-      for (var i = 0; i < util.colorArray.length; i++)
+      var colorArray = algo.rgbMapGetColors();
+      for (var i = 0; i < colorArray.length; i++)
       {
-        var sColor = util.colorArray[i];
-        var eColor = util.colorArray[(i + 1) % util.colorArray.length];
+        var sColor = colorArray[i];
+        var eColor = colorArray[(i + 1) % colorArray.length];
         if (! sColor)
           sColor = 0;
         if (! eColor)
@@ -270,7 +266,9 @@ var testAlgo;
     {
       if (! Array.isArray(rawColors))
         return;
-      for (var i = 0; i < algo.acceptColors; i++) {
+
+      util.colorArray = new Array(rawColors.length);
+      for (var i = 0; i < util.colorArray.length; i++) {
         if (i < rawColors.length)
         {
           util.colorArray[i] = rawColors[i];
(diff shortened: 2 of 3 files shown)


hjtappe:
Sorry, I did not find the time to investigate this problem deeper yet.
Why did you not re-use / revert the PR #1697 to keep this topic bundled?

But I could not reproduce the problem with balls.js when looking into the problem a few days before. In difference to plasma.js, balls.js does not change the acceptColors through a parameter. It seems to me that changing the acceptColors causes the trouble.
https://github.com/mcallegari/qlcplus/blob/83537d2c83463434d4f4ad0d0a6926a24684b69f/resources/rgbscripts/plasma.js#L84

Removing the lines removes the call, but will probably not initialize the script correctly, when parameters are set from the GUI into the script if I understand the line correctly.

A test will be:
- the colors from the script
  https://github.com/mcallegari/qlcplus/blob/83537d2c83463434d4f4ad0d0a6926a24684b69f/resources/rgbscripts/plasma.js#L85
  are correclty read when loading the script the first time (i.e. after loading the workspace file or changing to e.g. balls.js) or changing the script to custom colors
  https://github.com/mcallegari/qlcplus/blob/83537d2c83463434d4f4ad0d0a6926a24684b69f/resources/rgbscripts/plasma.js#L284
- Changes of the five colors in the UI is correctly propagated into the script, i.e. visible in the matrix

Scripts to test with: A selection of scripts of API version 2 and 3, for version 2 with different values for acceptColors (0, 1, 2) and for version 3 with different behaviors of acceptColors:
- balls.js (v3 with static acceptColors)
- plasma.js (v3 with changing acceptColors)
- alternate.js (v2 with 2 colors)
- strobe,js (v2 with one color)
- gradient.js (v2 with 0 colors)

Jurrie:
> Sorry, I did not find the time to investigate this problem deeper yet. Why did you not re-use / revert the PR #1697 to keep this topic bundled?

Because this PR is a completely different approach to tackle the problem.
- If I was to use PR #1697 by _reverting and committing_ the changes done here, I would have cluttered the Git history with commits that should not have been merged.
- If I was to use PR #1697 by _force-pushing_ changes, the discussion in #1697 would not make sense anymore: the first comments would be about code that was never there.

> But I could not reproduce the problem with balls.js when looking into the problem a few days before. In difference to plasma.js, balls.js does not change the acceptColors through a parameter. It seems to me that changing the acceptColors causes the trouble.

I think the difference is because `balls.js` is missing the `rgbMapGetColors` function. `plasma.js` has this function, and when QLC+ calls it, it overwrites the previous colors with the defaults again. QLC+ warns about this missing function: `"/qlcplus/run/usr/share/qlcplus/rgbscripts/balls.js" is missing the rgbMapGetColors() function!`. If I remove the function from `plasma.js` also (so that it mimics `balls.js`), it starts working again. That's the gist of PR #1697. But as we agreed there, removing the function does not seem like a real solution.

> https://github.com/mcallegari/qlcplus/blob/83537d2c83463434d4f4ad0d0a6926a24684b69f/resources/rgbscripts/plasma.js#L84
> 
> Removing the lines removes the call, but will probably not initialize the script correctly, when parameters are set from the GUI into the script if I understand the line correctly.
> 
> A test will be:
> 
>     * the colors from the script
>       https://github.com/mcallegari/qlcplus/blob/83537d2c83463434d4f4ad0d0a6926a24684b69f/resources/rgbscripts/plasma.js#L85
>       
>       are correclty read when loading the script the first time (i.e. after loading the workspace file or changing to e.g. balls.js) or changing the script to custom colors
>       https://github.com/mcallegari/qlcplus/blob/83537d2c83463434d4f4ad0d0a6926a24684b69f/resources/rgbscripts/plasma.js#L284
> 
>     * Changes of the five colors in the UI is correctly propagated into the script, i.e. visible in the matrix
> 
> 
> Scripts to test with: A selection of scripts of API version 2 and 3, for version 2 with different values for acceptColors (0, 1, 2) and for version 3 with different behaviors of acceptColors:
> 
>     * balls.js (v3 with static acceptColors)
> 
>     * plasma.js (v3 with changing acceptColors)
> 
>     * alternate.js (v2 with 2 colors)
> 
>     * strobe,js (v2 with one color)
> 
>     * gradient.js (v2 with 0 colors)

Thanks! I already manually tried some tidbits and did not see anything break. But I'll try your suggested tests and report back!

Jurrie:
> I think the difference is because `balls.js` is missing the `rgbMapGetColors` function.

I just tested this by adding the function to `balls.js`; it does _not_ make a difference. So I guess I'm wrong here.
In fact: there is no difference in behavior whatsoever, so why that `rgbMapGetColors` should be there, I don't know :shrug:...

Still, I see QLC+ setting properties and overwriting colors by doing so.

And I also see that `RGBMatrix::setProperty(QString propName, QString value)` will call `RGBMatrix::setColor(int i, QColor c)` which will call `RGBMatrix::setMapColors(RGBAlgorithm *algorithm)` which will call Javascript's `rgbMapSetColors`. Since `RGBMatrix::setProperty(QString propName, QString value)` will call `RGBMatrix::setColor(int i, QColor c)` _for every color_, this means that Javascript's `rgbMapSetColors` will be called 5 times instead of 1. And I think this is not just during set up, but during actual animation. Inefficient... but not the cause of this bug.

I wrote: "To be honest, I do not fully understand why the code was there."
A little context on this: the code was added in commit [6cd67bd387a727d47ff74c4fdb11904727d42bda in engine/src/rgbmatrix.cpp:373](https://github.com/mcallegari/qlcplus/commit/6cd67bd387a727d47ff74c4fdb11904727d42bda#diff-738a196c7c1d75f2b8fd99f0ec4e95db9833e51f425db995306babb00dc37c1eR373). The commit message can be read following that link, but it doesn't clarify why this was added for me. I'm not deep enough in the code. 😀

Jurrie:
I did some tests on both `master` as on this branch.

Expected initial colors:

balls.js | plasma.js | alternate.js | strobe.js | gradient.js
-- | -- | -- | -- | --
Green | Green | Red (set by QLC+, not by the script) | Red | Rainbow
Orange | Orange | - |   |  
Blue | Blue |   |   |  
Yellow | Yellow |   |   |  
Light blue | White |   |   |  


On `master`, this is the result:

  | balls.js | plasma.js | alternate.js | strobe.js | gradient.js
-- | -- | -- | -- | -- | --
Colors read from script on first time | **NO** (just red) | YES | YES | YES | YES
Colors changed in UI are propagated to matrix | YES | YES | YES | YES | YES
Deselect + select matrix retains colors | YES | **NO** (this PR is trying to fix this)| YES | YES | YES
Colors are saved to QXW file | YES | YES | YES | YES | YES
Colors are read from QXW file on load | YES | **NO** | YES | YES | YES

On `master`, when I add the missing `rgbMapGetColors` to `balls.js`:

  | balls.js | plasma.js | alternate.js | strobe.js | gradient.js
-- | -- | -- | -- | -- | --
Colors read from script on first time | YES | YES | YES | YES | YES
Colors changed in UI are propagated to matrix | YES | YES | YES | YES | YES
Deselect + select matrix retains colors | YES | **NO** | YES | YES | YES
Colors are saved to QXW file | YES | YES | YES | YES | YES
Colors are read from QXW file on load | **NO** | **NO** | YES | YES | YES

On my branch:

  | balls.js | plasma.js | alternate.js | strobe.js | gradient.js
-- | -- | -- | -- | -- | --
Colors read from script on first time | **NO** | **NO** | YES | YES | YES
Colors changed in UI are propagated to matrix | YES | YES | YES | YES | YES
Deselect + select matrix retains colors | YES | YES | YES | YES | YES
Colors are saved to QXW file | YES | YES | YES | YES | YES
Colors are read from QXW file on load | YES | YES | YES | YES | YES

So obviously my branch currently does not fix everything. But things are already broken on `master`.

resources/rgbscripts/plasma.js
```diff
@@ -61,30 +61,25 @@ var testAlgo;
...
       else if (_preset === "Ocean")
       {
         algo.presetIndex = 3;
-        util.colorArray = [ 0x003AB9, 0x02EAFF ];
       }
       else if (_preset === "User Defined")
       {
         algo.presetIndex = 4;
         algo.acceptColors = 5;
-        util.colorArray = [ 0x00FF00, 0xFFAA00, 0x0000FF, 0xFFFF00, 0xFFFFFF ];
```
Jurrie, line 85: Don't reset the colors when the `presetIndex` property is set by QLC+. That property is always set at start, then colors get loaded, then colors get set. So when you reset colors here, _those_ colors are set again. Ergo: you loose the user defined colors.

engine/src/rgbmatrix.cpp
```diff
@@ -242,6 +242,10 @@ void RGBMatrix::setAlgorithm(RGBAlgorithm *algo)
                     m_properties.take(it.key());
                 }
             }
+
+            QVector<uint> colors = script->rgbMapGetColors();
```
Jurrie, line 246: After setting all the properties, the default colors should be loaded and displayed in the UI. If there are no default colors (and `rgbMapGetColors` returns an empty array), we default to what `RGBMatrix` defines as default (1 color: red).

Jurrie:
Right, after some mucking around, I now have a situation that passes all tests:

  | balls.js | plasma.js | alternate.js | strobe.js | gradient.js
-- | -- | -- | -- | -- | --
Colors read from script on first time | YES | YES | YES | YES | YES
Colors changed in UI are propagated to matrix | YES | YES | YES | YES | YES
Deselect + select matrix retains colors | YES | YES | YES | YES | YES
Colors are saved to QXW file | YES | YES | YES | YES | YES
Colors are read from QXW file on load | YES | YES | YES | YES | YES

I must say that the RGB matrix code is a bit of a spaghetti. (No offense to any Italian hanging around here of course! 😉) For example: when a script is loaded in the RGB matrix editor, all the properties and their default values get loaded from the script, _and the defaults are set again_. This caused the `plasma.js` script to misbehave: the default for `presetIndex` was `Rainbow` (which has `acceptColors == 0`), causing it to loose all user-set colors and resetting to the default colors. The only way around this is by having the default properties the set with the maximum `acceptColors` that the script supports. So when a script supports at max 4 user-defined colors, its properties should default to a state where `acceptColors == 4`.

@mcallegari @hjtappe WDYT about this PR now?

engine/src/rgbmatrix.cpp
```diff
@@ -357,17 +362,21 @@ void RGBMatrix::setMapColors(RGBAlgorithm *algorithm)
     if (m_group == NULL)
         m_group = doc()->fixtureGroup(fixtureGroup());
 
-    if (m_group != NULL)
+    QVector<unsigned int> rawColors;
+    for (int i = 0; i < algorithm->acceptColors(); i++)
```
shaforostoff, line 366:
```
const int acceptColors = algorithm->acceptColors();
rawColors.reserve(acceptColors);
for (int i = 0; i < acceptColors; i++)
```

this way only one vector allocation will happen, and algorithm->acceptColors() will be called just one. This is important because each RGBScript::acceptColors call locks mutex and calls javascript code

shaforostoff (changes requested):

Jurrie:
I know not much time has passed, but is there anything I can do to get this merged? Maybe explain some bits some more?

engine/src/rgbmatrix.cpp
```diff
@@ -242,6 +242,10 @@ void RGBMatrix::setAlgorithm(RGBAlgorithm *algo)
                     m_properties.take(it.key());
                 }
             }
+
+            QVector<uint> colors = script->rgbMapGetColors();
+            for (int i = 0; i < colors.count(); i++)
+            	m_rgbColors.replace(i, QColor::fromRgb(colors.at(i)));
```
shaforostoff, line 248: i see a tab here. I guess the codestyle of this project implies spaces instead of tabs

shaforostoff (approved):
I have tried this branch on my computer, seems to work fine (although I never used this area of QLC+ before). I spotted one memory leak in the rgbmatrixeditor (_not_ inroduced by this PR), and i have added a fix for it to my PR #1748

engine/src/rgbmatrix.cpp
Jurrie: 303438163be901eb27b6b7e80abfb66ba8202109

shaforostoff:
Thanks, Could you please rebase your branch on top of latest master, and squash two top commits into the main commit "Remember colors that are set on RGB matrix"? to keep the history clean

mcallegari:
Looks good to me, thanks 👍

PR #1728, opened 2025-04-16
PNGs: losslessly recompress with zopflipng to reduce binary size

shaforostoff:
https://github.com/google/zopfli

(I didn't include files for which the gain was below 20%)


mcallegari:
Thanks 👍

PR #1729, opened 2025-04-17
Prevent NPE with setEFXInterval(uint) after 'stop all functions' is hit

Jurrie:
This was first reported [in the forum](https://forum.qlcplus.org/viewtopic.php?t=18302).


mcallegari:
Makes sense, thanks 👍

PR #1732, opened 2025-04-22
[WIP] Discussing the use of  typesafe Qt slot and signal connections

Jurrie:
This PR is to check if I'm on the right track, and if this would be accepted.

_This PR is not finished, and not intended to be merged._

Qt5 introduced support for typesafe `connect()`ing signals to slots. So when this PR is merged, it breaks Qt4 compatibility.

### DIFF
--- a/engine/audio/src/audio.cpp
+++ b/engine/audio/src/audio.cpp
@@ -225,7 +225,7 @@ int Audio::adjustAttribute(qreal fraction, int attributeId)
 
 void Audio::slotEndOfStream()
 {
-    if (!stopped())
+    if (!isStopped())
         stop(FunctionParent::master());
 }
 
--- a/engine/audio/src/audio.cpp
+++ b/engine/audio/src/audio.cpp
@@ -66,8 +66,7 @@ Audio::Audio(Doc* doc)
     setRunOrder(Audio::SingleShot);
 
     // Listen to member Function removals
-    connect(doc, SIGNAL(functionRemoved(quint32)),
-            this, SLOT(slotFunctionRemoved(quint32)));
+    connect(doc, &Doc::functionRemoved, this, &Audio::slotFunctionRemoved);
 }
 
 Audio::~Audio()
@@ -365,8 +364,7 @@ void Audio::preRun(MasterTimer* timer)
         m_audio_out->setFadeIn(elapsed() ? 0 : fadeIn);
         m_audio_out->setLooped(runOrder() == Audio::Loop);
         m_audio_out->start();
-        connect(m_audio_out, SIGNAL(endOfStreamReached()),
-                this, SLOT(slotEndOfStream()));
+        connect(m_audio_out, &AudioRenderer::endOfStreamReached, this, &Audio::slotEndOfStream);
     }
 
     Function::preRun(timer);
--- a/engine/src/chaser.cpp
+++ b/engine/src/chaser.cpp
@@ -57,8 +57,7 @@ Chaser::Chaser(Doc *doc)
     setName(tr("New Chaser"));
 
     // Listen to member Function removals
-    connect(doc, SIGNAL(functionRemoved(quint32)),
-            this, SLOT(slotFunctionRemoved(quint32)));
+    connect(doc, &Doc::functionRemoved, this, &Chaser::slotFunctionRemoved);
 
     m_startupAction.m_action = ChaserNoAction;
     m_startupAction.m_masterIntensity = 1.0;
@@ -627,7 +626,7 @@ void Chaser::preRun(MasterTimer* timer)
     {
         QMutexLocker runnerLocker(&m_runnerMutex);
         createRunner(elapsed());
-        connect(m_runner, SIGNAL(currentStepChanged(int)), this, SIGNAL(currentStepChanged(int)));
+        connect(m_runner, &ChaserRunner::currentStepChanged, this, &Chaser::currentStepChanged); // TODO: This was connecting SIGNAL to SIGNAL - is that even correct?
     }
 
     Function::preRun(timer);
--- a/fixtureeditor/editchannel.cpp
+++ b/fixtureeditor/editchannel.cpp
@@ -61,7 +61,7 @@ EditChannel::EditChannel(QWidget *parent, QLCChannel *channel)
 
     QAction *action = new QAction(this);
     action->setShortcut(QKeySequence(QKeySequence::Close));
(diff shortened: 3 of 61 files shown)


engine/audio/src/audio.cpp
```diff
@@ -225,7 +224,7 @@ int Audio::adjustAttribute(qreal fraction, int attributeId)
 
 void Audio::slotEndOfStream()
 {
-    if (!stopped())
+    if (!isStopped())
```
Jurrie, line 227: This change is done because there is another method named `stopped` with a different parameter list. One can use lambda's, but renaming to `isStopped()` is cleaner IMO. There is already an `isPaused()` and `isRunning()`.

engine/src/chaser.cpp
```diff
@@ -627,7 +626,7 @@ void Chaser::preRun(MasterTimer* timer)
     {
         QMutexLocker runnerLocker(&m_runnerMutex);
         createRunner(elapsed());
-        connect(m_runner, SIGNAL(currentStepChanged(int)), this, SIGNAL(currentStepChanged(int)));
+        connect(m_runner, &ChaserRunner::currentStepChanged, this, &Chaser::currentStepChanged); // TODO: This was connecting SIGNAL to SIGNAL - is that even correct?
```
Jurrie, line 629: In some places the original code connects a `SIGNAL` to a `SIGNAL`. Since both `SIGNAL`s and `SLOT`s are just functions, I assume it works. But was it intended this way...?

fixtureeditor/editchannel.cpp
```diff
@@ -117,8 +116,7 @@ void EditChannel::init()
         m_presetCombo->setCurrentIndex(m_channel->preset());
     }
 
-    connect(m_presetCombo, SIGNAL(activated(int)),
-            this, SLOT(slotPresetActivated(int)));
+    connect(m_presetCombo, static_cast<void (QComboBox::*)(int)>(&QComboBox::activated), this, &EditChannel::slotPresetActivated);
```
Jurrie, line 119: The cast is necessary here because there are two methods `QComboBox::activated`, and we have to let the compiler know which one we intend to target as signal.

hotplugmonitor/src/hotplugmonitor.cpp
```diff
@@ -56,12 +56,10 @@ void HotPlugMonitor::connectListener(QObject* listener)
     QByteArray removed = QMetaObject::normalizedSignature("slotDeviceRemoved(uint,uint)");
 
     if (listener->metaObject()->indexOfMethod(added.constData()) != -1)
-        connect(instance(), SIGNAL(deviceAdded(uint,uint)),
-                listener, SLOT(slotDeviceAdded(uint,uint)));
+        connect(instance(), SIGNAL(deviceAdded(uint,uint)), listener, SLOT(slotDeviceAdded(uint,uint))); // TODO: Is there a type-safe way to connect() this? The type of `listener` is not known at compile time... Maybe introduce an interface 'IHotplugListener'?
```
Jurrie, line 59: Any thoughts on this?

plugins/E1.31/e131plugin.cpp
```diff
@@ -173,8 +173,7 @@ bool E131Plugin::openOutput(quint32 output, quint32 universe)
         E131Controller *controller = new E131Controller(m_IOmapping.at(output).iface,
                                                         m_IOmapping.at(output).address,
                                                         output, this);
-        connect(controller, SIGNAL(valueChanged(quint32,quint32,quint32,uchar)),
-                this, SIGNAL(valueChanged(quint32,quint32,quint32,uchar)));
+        connect(controller, &E131Controller::valueChanged, this, [this](quint32 _t1, quint32 _t2, quint32 _t3, uchar _t4){ valueChanged(_t1, _t2, _t3, _t4); });
```
Jurrie, line 176: This lambda is necessary because the fuction `E131Plugin::valueChanged` has 5 parameters while the signal only has 4. The fifth one is an optional parameter, but the compiler doesn't understand that.

plugins/hid/hidplugin.cpp
```diff
@@ -88,8 +87,7 @@ void HIDPlugin::closeInput(quint32 input, quint32 universe)
     {
         removeFromMap(input, universe, Input);
         dev->closeInput();
-        disconnect(dev, SIGNAL(valueChanged(quint32,quint32,quint32,uchar)),
-                   this, SIGNAL(valueChanged(quint32,quint32,quint32,uchar)));
+        disconnect(dev, SIGNAL(valueChanged(quint32,quint32,quint32,uchar)), this, SIGNAL(valueChanged(quint32,quint32,quint32,uchar))); // TODO: The slot has one more parameter than the signal. We can connect() using a lambda, but can not easily disconnect.
```
Jurrie, line 90: This one is tricky... We could `connect()` using a lambda (same reason as https://github.com/mcallegari/qlcplus/pull/1732/files#r2054325313). But `disconnect()`ing using a lambda is not possible of course. [There is a way](https://stackoverflow.com/a/14829520) but I think that would make things only less intuitive. Maybe continue to use the type-unsafe way here?

Jurrie:
> While I started to use this type of syntax in the v5 code, I'm not sure I want to embark on this activity for the engine too, for the following reasons:
> 
>     * everything, I repeat -everything- needs to be tested again and I don't have time to do that
> 
>     * for some reason, some signals from the engine doesn't work in the UI using the "new" method. No idea why
> 
>     * it doesn't quite provide any improvement in terms of performance or readibility
> 
> 
> Looks to me like an exercise more than an advancement. Sorry to disappoint but right now this is my opinion

Alright, I understand. Maybe it's better suited for v5 indeed. Since that's in development people should expect it to be a little less stable.

> P.S. If you want tasks on QLC+ just tell me. I've got plenty!

I'm hesitant 😀 I do some coding on QLC+ to get my own itches scratched and to learn a bit of C++ / Qt in the process. I'm a Java-dev by trade. This type-safe way of `connect()`ing signals and slots was something I stumbled upon while debugging RGBMatrix et al (for https://github.com/mcallegari/qlcplus/pull/1727). And since Java is type-safe, it appealed to me, and I thought I'd dig into this some more.

So when I take on a task I can't promise I get it done. It might not trigger my interest, or it might be that I don't have the time. I know I sound like a lazy bum. But the thing is that the stuff I do my spare time should be stuff I find pleasant to do. I hope you understand.

That said: is that list of tasks available somewhere? I'll browse through it to see if anything interests me 😀

PR #1742, opened 2025-04-25
Fix high RAM usage related to RGB scripts

shaforostoff:
RAM usage goes down from 1.2GB to 150MB when loading Project One.qxw

[qxw.zip](https://github.com/user-attachments/files/19917381/qxw.zip)


mcallegari:
Hey @shaforostoff thanks for this.
Can you please explain a bit your findings?

While reviewing the change, I see some interesting things, but I think it will crash as it did before my latest changes.
If you create a script from the RGBScriptCache (so in the UI thread) and then use it from the MasterTimer thread, it will crash.
You can easily reproduce this with a large matrix (eg. 20x20 rgb panel) and the plasma effect.
I'm gonna check your PR myself

mcallegari:
And unfortunately it crashes.
Please use the attached sample. 
- Open the RGB matrix editor and hit preview
OR
- Go to VC and press the button. 
It should crash within a few seconds.
[plasma.zip](https://github.com/user-attachments/files/19921025/plasma.zip)

shaforostoff:
> It should crash within a few seconds. [plasma.zip](https://github.com/user-attachments/files/19921025/plasma.zip)

Thanks for providing an easy way to reproduce the crash! I have pushed a new version which doesn't crash with this plasma.qxw (i opened the file, then pressed Play and moved slider in the virtual console to reproduce the crash).

I tried to explain each change in commit messages, if you still have questions I will gladly answer them

engine/src/rgbscriptscache.cpp
```diff
@@ -39,18 +39,18 @@ QStringList RGBScriptsCache::names() const
...
     {
-        return *mScript;
+        return mScript;
     }
     else
     {
         mScript->load(filename);
-        return *mScript;
+        return mScript;
     }
```
shaforostoff, line 54: the huge leak was happening here

mcallegari:
hey @shaforostoff I just tested your latest changes and found that this sequence freeze QLC+:
- open plasma project
- go to RGB matrix editor (preview is OK)
- hit the "play" button
- change the RGB script

shaforostoff:
> hey @shaforostoff I just tested your latest changes and found that this sequence freeze QLC+:
> 
> * open plasma project
> * go to RGB matrix editor (preview is OK)
> * hit the "play" button
> * change the RGB script

Thanks for pointing this out!!
This is now fixed by having a dedicated thread which runs QJSEngine.

mcallegari:
@shaforostoff I have done some tests, even with multiple RGB panels and two matrices activated at the same time and it works pretty well! Thank you for this.
While attempting to fix the thread safety issue the first time I attempted several ways, including using QMetaObject::invokeMethod but for some reason it still crashed for me :(
Glad you found a way to make it work. Sometimes Qt can really be obscure...

Just a couple of consideration on this PR: 
- please stick to the code style and add a space after if/else/for/while
- I would separate audio triggers and xypad changes into another PR. Just to keep history clean

Thanks again!

PR #1743, opened 2025-04-29
Do not updateStepColor() with 1-too-high step count

Jurrie:
During the "RGB Matrix to sequence" logic, while looping through the RGB Matrix, everything is ok. But the last step of the loop will try to set `currentStep` to a value that is 1 too high. This will in turn cause QLC+ to try to set colors to values out of range 0 <= c < 256, causing errors in the console.


mcallegari:
@Jurrie thanks. Does this fix a specific report from the forums?

Jurrie:
> @Jurrie thanks. Does this fix a specific report from the forums?

No it does not. I tried to reproduce https://www.qlcplus.org/forum/viewtopic.php?t=18273 and found this while doing so.

PR #1745, opened 2025-04-30
Port QML UI to Qt6

mcallegari:
Also, migrate Windows builds to ninja and make use of official Qt versions and windeployqt

PR #1746, opened 2025-04-30
Update m_columnSpin when m_compCombo or m_16bitCheck change

Jurrie:
We now include the 16bit checkbox in the calculation of the maximum amount of columns.
We also set the value of m_columnSpin to the maximum if its current value is over the new maximum.

### DIFF
--- a/ui/src/addrgbpanel.cpp
+++ b/ui/src/addrgbpanel.cpp
@@ -21,6 +21,8 @@
 #include <QDebug>
 #include <QSettings>
 
+#include <math.h>
+
 #include "addrgbpanel.h"
 #include "doc.h"
 
@@ -53,7 +55,14 @@ AddRGBPanel::AddRGBPanel(QWidget *parent, const Doc *doc)
     connect(m_uniCombo, SIGNAL(currentIndexChanged(int)),
             this, SLOT(slotUniverseChanged()));
     connect(m_compCombo, SIGNAL(currentIndexChanged(int)),
-            this, SLOT(slotComponentsChanged(int)));
+            this, SLOT(slotComponentsChanged()));
+#if (QT_VERSION < QT_VERSION_CHECK(6, 7, 0))
+    connect(m_16bitCheck, SIGNAL(stateChanged(int)),
+            this, SLOT(slotComponentsChanged()));
+#else
+    connect(m_16bitCheck, SIGNAL(checkStateChanged(Qt::CheckState)),
+            this, SLOT(slotComponentsChanged()));
+#endif
     connect(m_addressSpin, SIGNAL(valueChanged(int)),
             this, SLOT(slotAddressChanged()));
     connect(m_columnSpin, SIGNAL(valueChanged(int)),
@@ -96,12 +105,19 @@ void AddRGBPanel::slotUniverseChanged()
     checkAddressAvailability();
 }
 
-void AddRGBPanel::slotComponentsChanged(int index)
+void AddRGBPanel::slotComponentsChanged()
 {
-    if (index == 6) // RGBW
-        m_columnSpin->setMaximum(128);
-    else
-        m_columnSpin->setMaximum(170);
+    int dmxChannelsPerPixel = 3;
+    if (m_compCombo->currentIndex() == 6) // RGBW
+        dmxChannelsPerPixel = 4;
+
+    if (m_16bitCheck->isChecked())
+        dmxChannelsPerPixel *= 2;
+
+    m_columnSpin->setMaximum(floor(UNIVERSE_SIZE / dmxChannelsPerPixel));
+
+    if (m_columnSpin->value() > m_columnSpin->maximum())
+        m_columnSpin->setValue(m_columnSpin->maximum());
 }
 
 void AddRGBPanel::slotAddressChanged()
--- a/ui/src/addrgbpanel.h
+++ b/ui/src/addrgbpanel.h
@@ -80,7 +80,7 @@ private:
 
 protected slots:
     void slotUniverseChanged();
-    void slotComponentsChanged(int index);
+    void slotComponentsChanged();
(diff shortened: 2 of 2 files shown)


mcallegari:
Hi @Jurrie sorry for the confusion.
The aim was to resolve a Codacy indication about an include but in the end it's the same as your original PR.
I think qFloor vs std::floor doesn't make any difference.
Actually, the standard C++ method is preferred, considering the amount of deprecations happening so fast in Qt.
I'll revert my changes.

ui/src/addrgbpanel.cpp
```diff
@@ -53,7 +54,9 @@ AddRGBPanel::AddRGBPanel(QWidget *parent, const Doc *doc)
     connect(m_uniCombo, SIGNAL(currentIndexChanged(int)),
             this, SLOT(slotUniverseChanged()));
     connect(m_compCombo, SIGNAL(currentIndexChanged(int)),
-            this, SLOT(slotComponentsChanged(int)));
+            this, SLOT(slotComponentsChanged()));
+    connect(m_16bitCheck, SIGNAL(stateChanged(int)),
```
mcallegari, line 58: Which Qt version have you used to test this signal? Qt 5.x?
According to documentation, Qt 6.7+ emits [checkStateChanged](https://doc.qt.io/qt-6/qcheckbox.html#checkStateChanged)

Jurrie: I just used the defaults, so I think it's Qt 5 (since the GitHub actions seem happy, and their name suggests they use Qt 5). I'm happy to test with Qt 6, but how to build? I checked [the wiki](https://github.com/mcallegari/qlcplus/wiki) but didn't find a page that explains it. (In fact, [this page](https://github.com/mcallegari/qlcplus/wiki/QLC-Plus-5-build-HOWTO) explicitly states that Qt 6 is not supported yet.)

mcallegari:
The page you referred is to build v5 with Qt6. On master that is still true, but I worked for weeks to port it.
See my [huge PR](https://github.com/mcallegari/qlcplus/pull/1745)

Build instructions for v4 are still valid if you point to Qt6 instead of Qt5.
On which OS do you build?

Jurrie:
> The page you referred is to build v5 with Qt6. On master that is still true, but I worked for weeks to port it. See my [huge PR](https://github.com/mcallegari/qlcplus/pull/1745)

Impressive PR indeed! I knew you were busy with Qt 6 support but didn't realize there was a PR already.

> Build instructions for v4 are still valid if you point to Qt6 instead of Qt5. On which OS do you build?

I use Ubuntu LTS. If you could tell me how to point cmake to Qt 6, then I'll check if this PR works using Qt 6.

mcallegari:
Download the official Qt version and do:
```
mkdir build && cd build
cmake -DCMAKE_PREFIX_PATH="/home/your_user/Qt/6.9.0/gcc_64/lib/cmake/" ..
make
```

Jurrie:
> Download the official Qt version and do:
> 
> ```
> mkdir build && cd build
> cmake -DCMAKE_PREFIX_PATH="/home/your_user/Qt/6.9.0/gcc_64/lib/cmake/" ..
> make
> ```

Oh shoot, you were just ahead of me. I am using the packages in the Ubuntu repository instead. I already do that for Qt 5; that's why I struggled with Qt 6: the package names are sometimes different.

But it seems to compile and run ok. (Just a tiny change in `/qlcplus/engine/src/script.cpp`.)
Want me to wrap up some build instructions using vanilla Ubuntu and system Qt 6packages? Or isn't that useful?

mcallegari:
> But it seems to compile and run ok. (Just a tiny change in /qlcplus/engine/src/script.cpp.)

Wait. The "old" Script class chould not be built with Qt6. You're doing something wrong.
Please start from a clean checkout

> Want me to wrap up some build instructions using vanilla Ubuntu and system Qt 6packages? Or isn't that useful?

Well, technically the official Linux release is still based on Qt5 but Ok, instructions can be updated for Qt6.
To be honest though, I still don't have a Linux distro where Qt6 is the default.

ui/src/addrgbpanel.cpp
Jurrie: In Qt 6.9.0 `stateChanged(int)` is indeed deprecated in favor of `checkStateChanged(Qt::CheckState)`.

Jurrie: This should not be 6.9.0, but also not 6.0.0. It should be 6.7.0 as the documentation states it is introduced in that version.

mcallegari:
@Jurrie  Why have you overwritten my change?
I checked before doing it:
https://doc.qt.io/archives/qt-6.0/qcheckbox.html#stateChanged

stateChanged was present in Qt 6.0.0

Jurrie:
> @Jurrie Why have you overwritten my change? I checked before doing it: [doc.qt.io/archives/qt-6.0/qcheckbox.html#stateChanged](https://doc.qt.io/archives/qt-6.0/qcheckbox.html#stateChanged)
> 
> stateChanged was present in Qt 6.0.0

It should be `stateChanged` for all versions of Qt 5, and Qt 6 up to (excluding) 6.7.0, right? That's how I read [the docs on checkStateChanged](https://doc.qt.io/qt-6/qcheckbox.html#checkStateChanged) which you pointed me to in [your comment](https://github.com/mcallegari/qlcplus/pull/1746#discussion_r2071205623). And from Qt 6.7.0 and up it should be `checkStateChanged`?

So the code should read something like:

```
#if (QT_VERSION < QT_VERSION_CHECK(6, 7, 0))
    // Use stateChanged(int)
#else
    // Use checkStateChanged(Qt::CheckState)
#endif
```

Or am I completely misinterpreting the Qt documentation?

mcallegari:
You're right. I misread the name in the documentation.

PR #1749, opened 2025-05-03
Fix hanging connections because maxPendingConnections is reached

Jurrie:
The bug in question:
Open QLC+ using `--web`, start a browser and navigate to `http://localhost:9999`, then start refreshing bypassing the browsers cache using CTRL+SHIFT+R. After some attempts you will see that the browser's connection starts hanging.

The reason for this is that every connection is added to the list of "pending connections", and this list is finite. By default, it's 30 positions. So after 30 connections (with every page request using multiple connections), things get clogged up.

The solution:
Since QLC+ doesn't use pending connections in `QHttpServer` (there is no call to `m_tcpServer->nextPendingConnection()`), it is no problem to remove support for that.

Either that, or we should call nextPendingConnection() for every connection we add with addPendingConnection(QTcpSocket*). But I wasn't sure where to do that call.

### DIFF
--- a/webaccess/src/qhttpserver/qhttpserver.cpp
+++ b/webaccess/src/qhttpserver/qhttpserver.cpp
@@ -141,6 +141,4 @@ void CustomTcpServer::incomingConnection(qintptr handle)
             this, SIGNAL(webSocketDataReady(QHttpConnection*,QString)));
     connect(connection, SIGNAL(webSocketConnectionClose(QHttpConnection*)),
             this, SIGNAL(webSocketConnectionClose(QHttpConnection*)));
-
-    addPendingConnection(socket);
 }


mcallegari:
Intersting one!
Might this be related too?
https://www.qlcplus.org/forum/viewtopic.php?t=18239

Jurrie:
> Interesting one!
> Might this be related too?
> [qlcplus.org/forum/viewtopic.php?t=18239](https://www.qlcplus.org/forum/viewtopic.php?t=18239)

Could very well be. The forum message talks about two different clients, each with a browser. That's two times the connection count of a single user. When a single user "behaves like a normal user" and doesn't CTRL+SHIFT+R like a madman, you probably don't notice. But double that count, and it might be enough to trigger the bug.

I also think the bug is more prevalent now that [the favicons are in](https://github.com/mcallegari/qlcplus/pull/1698). The browser will try to side-load these, using yet another connection.

On a side note: I do notice this bug regularly when doing shows. My use-case is that I'm on an Android tablet, switching between another app and the browser. The browser will reload every time I switch. That has probably triggered this bug. When it occurs I notice that QLC+ will still run the light show and even respond to OSC messages. So the app in general is not crashing, just the HTTP web console side. And a stop+start of the QLC+ process fixes everything.

> An alternative is [Qt HTTP Server](https://doc.qt.io/qt-6/qthttpserver-index.html), but they licensed it as GPLv3 so not compatibe with QLC+ code

As explained [here](https://www.apache.org/licenses/GPL-compatibility.html). I see... Too bad 🙁

shaforostoff:
> Interesting one! Might this be related too? https://www.qlcplus.org/forum/viewtopic.php?t=18239
> 
> > On a side note, the QHttpServer project [nikhilm/qhttpserver#72 (comment)](https://github.com/nikhilm/qhttpserver/issues/72#issuecomment-1430785252). There is [nikhilm/qhttpserver#64](https://github.com/nikhilm/qhttpserver/pull/64), which seems to fix a memory leak. Maybe that PR should be applied in QLC+? Or maybe there is a maintained alternative to QHttpServer somewhere?
> 
> An alternative is [Qt HTTP Server](https://doc.qt.io/qt-6/qthttpserver-index.html), but they licensed it as GPLv3 so not compatibe with QLC+ code

this change doesn't change anything, because m_request is always created by passing QHttpConnection as its parent. Both are QObjects, so when QHttpConnection is destructed, QHttpRequest is deleted as well because it is a child of connection in QObject terms. On the countrary, this change is harmful because deleting QObjects directly via delete operator instead of calling deleteLater can cause a crash, because deleteLater proceeds with deletion only after all signal-slot connections have been processed. If QObject is deleted via delete operator and there are still signal-slot connections pending, they will operate on the already freed pointer.

shaforostoff:
I have tried this branch and it works well, I am voting for merging it

mcallegari:
> I have tried this branch and it works well, I am voting for merging it

So is this change good or bad? Didn't seem so from your first review

shaforostoff:
This QLC+ PR is good. I was referring to qhttpserver PR https://github.com/nikhilm/qhttpserver/pull/64 which we should _not_ port to our code.

shaforostoff (approved):

mcallegari:
> This QLC+ PR is good. I was referring to qhttpserver PR [nikhilm/qhttpserver#64](https://github.com/nikhilm/qhttpserver/pull/64) which we should _not_ port to our code.

Ah OK!
Let's merge this one

PR #1755, opened 2025-05-24
allow any serial port device.

plutphil:
Hi,
I'm trying to use an arduino as a dmx adapter. Therefore i would like to make this pull request. I guess i need to change the code a bit. It would be nice that the com port number or path is visible in ui. Is it possible to use just a usb serial converter with an rs485 module?


mcallegari:
Rejected. You need to build the UART plugin which is used on the Raspberry Pi.

plutphil:
> Rejected. You need to build the UART plugin which is used on the Raspberry Pi.

does this work on windows?

PR #1756, opened 2025-05-25
add two Stairville fixtures

theowl337:
Added two fixtures:
- Stairville BEL6 IP Bar HEX (only 6 channel mode)
- Stairville BSW-350 LED (only 32 channel mode)


theowl337:
Thanks for reminding me.
I'll be sure to add the other modes and validate them.

PR #1722, opened 2025-04-15
The QTreeWidget for Slider dialog → Level tab now allows sorting

Jurrie:
This was a feature request in [the forum](
https://forum.qlcplus.org/viewtopic.php?t=18266).

PR #1763, opened 2025-06-08
ui/virtualconsole: fix sorting in the level tab of the slider dialog

jpue:
#1722 (or 4bbba7ba85716ce3076739db013a25633a9dcaac, respectively) introduced the possibility to sort the fixture-channel-list on the `Level` page of Virtual Console Slider properties dialog.

However, this was done by comparing the channel names as a string and thus introduced known "sorting anomalies": Channel 1 - Channel 10 - Channel 11 - Channel 2 - ... (tested with the `Lixada Mini Gobo Moving Head` fixture).

In my opinion, this is particularly unintuitive for non-programmer users of QLC+.

This PR attempts to fix this problem by basing the sort on the numeric channel id instead of the string name.

(I am not very familiar with Qt Framework and its internals, so if there is a better solution, please treat this PR as an issue report and ignore my code.)

### DIFF
--- a/ui/src/CMakeLists.txt
+++ b/ui/src/CMakeLists.txt
@@ -103,6 +103,7 @@ add_library(${module_name}
     simpledeskengine.cpp simpledeskengine.h
     speeddial.cpp speeddial.h
     speeddialwidget.cpp speeddialwidget.h
+    treewidgetitem.cpp treewidgetitem.h
     universeitemwidget.cpp universeitemwidget.h
     videoeditor.cpp videoeditor.h videoeditor.ui
     videoprovider.cpp videoprovider.h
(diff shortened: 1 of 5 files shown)


mcallegari:
I wonder, since the UI presents the channel number, what if we always present a 3 digit number?
001: pan
002: pan fine
...

jpue:
That would definitely be easier to implement, but only make sense if fixtures cannot have more than 999 channels. Of course, a DMX universe has “only” 512 channels, but I have not tested large cross-universe RGB panels, for example.

In any case, I have implemented the change you suggested. I can force-push it here, if you like.

mcallegari:
Yeah please push here. Thanks
If anyone stumbles into a 999+ channel fixture then we'll add another digit here :wink:

mcallegari:
Thanks :+1:

PR #1768, opened 2025-06-19
Fixtureeditor/mode editor: disable "source" channel item in ActsOn co…

edogawa23:
…mbobox list

Apparently the docs on Fixture editor's ActsOn setting are not being read/found or misunderstood constantly by new contributors of fixture definitions. One reason might be that the Mode editor screenshot in the manual is from an older version and doesn't even include the "Acts On:"column so the text explaining the functionality can easily be overseen or just be ignored, IMHO.
So this PR tries to reduce this common error for new users by disabling the channel the combobox is clicked on.

This is just a simple workaround, but at least it might help avoiding the same mistake again and again.
A more complete solution would be to preselect and enable only possible target channels for a certain channel type, but that's beyond my programming capabilities I fear...


mcallegari:
Hey @edogawa23 I just pushed a change that improves this PR. Thanks for submitting
https://github.com/mcallegari/qlcplus/commit/058270df23ded45fb16a47e1b6dc8200faf60e78

PR #1771, opened 2025-06-23
GenericFader: protect map to fix crash

shaforostoff:
reproducible with 12 Tri - 12 RW - 11 Strobo - 12 Moving Heads.qxw

### DIFF
--- a/engine/src/genericfader.cpp
+++ b/engine/src/genericfader.cpp
@@ -92,6 +92,7 @@ void GenericFader::add(const FadeChannel& ch)
 {
     quint32 hash = channelHash(ch.fixture(), ch.channel());
 
+    QWriteLocker l(&m_channelsLock);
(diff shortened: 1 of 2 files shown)


mcallegari:
Interesting, thank you!
I'm wondering why coverage build crashes on Scene test?
Also, if there is a synchronization issue related to adding/removing faders, why not using a mutex?

shaforostoff:
> Interesting, thank you! I'm wondering why coverage build crashes on Scene test? Also, if there is a synchronization issue related to adding/removing faders, why not using a mutex?

The scene test is fixed now. QReadWriteLock is actually a special kind of mutex. It allows simultaneous read access from multiple threads. If you're sure two threads will never attempt reading the map at the same time, of course I can replace it with a simple QMutex

mcallegari:
@shaforostoff can you please rebase with master?
I think some latest changes slipped into this PR. Thanks

mcallegari:
Thanks a lot! 👍

PR #1778, opened 2025-07-09
Update vcxypad.cpp

gino1631:
fixed a bug where MIDI note off also register as key press that's why the button seem to flash instead of latching, also fixed other bugs where feedback is not registering properly


mcallegari:
Please describe the use case, the controller you used to test the change, the observed behaviour and the expected behaviour.
If available, a workspace to reproduce the issue.
Like this, a PR will likely never be accepted without a context.

Issue #1779, opened 2025-07-13
Some overdriven MP3's cause audible "crackling" when played


sharky5102:
**Describe the bug**
I received some MP3's and added them to a QLC+ project. When playing, high-volume parts of the mp3's caused audible "crackling", as if clipped, but worse. I dove into the issue and found that when I load the mp3 into, say, Audacity, the mp3 track actually contains samples that are overdriven (>1.0f and <-1.0f). Apparently, the default for libsndfile is to not check for such overdrive, and the samples "wrap around" (over +32767) to -32768, causing audible crackles. Other mp3 players apparently do this correctly

**To Reproduce**
1. Add an mp3 which is overdriven. Unfortunately the mp3's I have are copyright-protected, so I cannot attach them

**Expected behavior**
We should be clipping the samples at +32767 and -32767

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows 10
 - QLC+ Version Latest 4.x from git (23debfddb3b53bee2c92336a330f2c50b9f05fd0), but also on latest release 4.14.3
- hardware setup (mandatory for audio/video issues): Just windows jack output. Heard it via line out and headphones on bluetooth

**Additional context**
I have the pull request for you!

sharky5102:
I have verified that this one-liner fixes the crackling on my MP3s

PR #1780, opened 2025-07-13
Enable libsndfile clipping to avoid integer overflow

sharky5102:
Fixes https://github.com/mcallegari/qlcplus/issues/1779

Some mp3's (possibly other formats?) seem to overdrive the signal at
>1.0f or <-1.0f .. during conversion to PCM (16-bit signed integer), by
default no overflow or underflow is detected or corrected. This makes these samples "wrap around" and cause audible "crackle" artifacts. Enabling SFC_SET_CLIPPING = True, enables a different float conversion in libsndfile which clamps the output to the 16-bit signed integer range, avoiding the crackles.


mcallegari:
Thanks 👍

PR #1785, opened 2025-07-21
add 3 fixture

Lorys89:


mcallegari:
Rejected. These don't even pass the [online validator](https://www.qlcplus.org/fixture_validator.php).

PR #1797, opened 2025-08-21
remove legacy Qt4 code

jpue:
57a4d7adb0328dd5d0be3b65b431e43496cb2af0 removed most of the code needed to support Qt4, but for some reason, this code (located under `engine/audio`) was kept in place.
However, since it is only compiled and used when building with a version prior to Qt5, I think that it is no longer used on any platform currently supported by QLC+.
Therefore, it can now be safely removed (while the transition from Qt5 to Qt6 is already underway) to make the source code tree a little clearer.


mcallegari:
No, that code stays even if not used.
It doesn't harm anyone or any build process.
For example portaudio is still a valid audio backend so, in case, it can be easily re-included.
It's not related to Qt4.

PR #1801, opened 2025-09-10
virtualconsole/vcclockproperties: disable schedule controls if not in regular clock mode

jpue:
Inspired by [this](https://www.qlcplus.org/forum/viewtopic.php?t=18603) question in the forum, this PR disables (greys out) the schedule controls of the `VCClockProperties` window if the clock widget is not in regular clock mode.

This informs the user that the schedule is not used in those modes (stopwatch and countdown).

I think this way of visualising the behaviour is a good compromise between not changing the UI at all (as it is currently) and completely hiding the controls in those scenarios (which might lead the user to assume that the schedule is going to be deleted after saving the properties in non-regular-clock mode).

While working on the `vcclock*` classes, I also added the `const` keyword to all functions which do not change the state of the member variables.

(Tested on both Qt5 and Qt6 builds of QLC+ (Ubuntu))

### DIFF
--- a/ui/src/virtualconsole/vcclock.cpp
+++ b/ui/src/virtualconsole/vcclock.cpp
@@ -123,7 +123,7 @@ VCClock::ClockType VCClock::clockType() const
     return m_clocktype;
 }
 
-QString VCClock::typeToString(VCClock::ClockType type)
+QString VCClock::typeToString(VCClock::ClockType type) const
 {
     if (type == Stopwatch)
         return "Stopwatch";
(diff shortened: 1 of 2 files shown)


mcallegari:
Hey @jpue thanks for this.
Inspired by this PR, I just [pushed a patch](https://github.com/mcallegari/qlcplus/commit/680c4859a7d5cb821c7f695e9a95d4cd1ec5c785) to run Functions also in Countdown mode.
It required a bit of UI review but it's more useful this way

jpue:
Ok, nice, that is indeed more useful than my visual workaround. Thank you!

Issue #1784, opened 2025-07-21
Persist feedback when changing input source of VC controls


p-rogalski:
I'm currently developing a VC which mirrors my button matrix midi controller (Akai APC mini mk2). Sometimes I want to relocate functions on the button matrix, but this is quite tedious:

- Move the button to the target position on the VC
- Change input source assignment
- Set up feedback from scratch because it is reset when changing the input source <-- _This is the tedious part because one has to click through multiple dropdowns every time the input source get's changed_

**Proposal:**

What if we persist feedback values when the input source changes? If I'm correct it should suffice to save the old feedback values from `m_inputSource` and apply them again when the new source is set in

https://github.com/mcallegari/qlcplus/blob/8a22b7b840a2f222ac623a88136bd45a37dfeb99/ui/src/inputselectionwidget.cpp#L112-L116

Something like

```cpp
void InputSelectionWidget::setInputSource(const QSharedPointer<QLCInputSource> &source)
{
    if (!m_inputSource.isNull())
    {
        // Save feedback
    }
    m_inputSource = source;
    if (!m_inputSource.isNull())
    {
        // Restore feedback
    }
    updateInputSource();
}
```

@mcallegari Would this work?

yestalgia:
Another use case for something like this is when I rock up to a venue and want to add more Artnet universes at the start of my project. 

Why would I want to add the universes to the start of the project? Aside from neatness, the Artnet plugin defaults nicely to selecting the output universe (Art: 3) to the corresponding QLC+ universe (Universe: 3). 

Currently, I have to add universes to the end of the project and start the whole MIDI assignment over again.

Easy reassigning to another universe would improve the ability to add universes at the top of the list. It might even open the possibility to enable the + icon when a non-final universe is selected, automatically reassigning the inputs for the VC to the next universe, perhaps even remapping the fixtures. 

Regardless I think persisting the input source is a good idea.

yestalgia:
@p-rogalski I've assigned @Binary-Vanguard-12138  to work on this issue. 

Can you please test the fix that has been done in #1805 and report back? If possible, please try a number of scenarios including all widget types + flash buttons for sliders.

Any issues comment on that PR. 

You can find the build here: https://github.com/mcallegari/qlcplus/actions/runs/17732644285

p-rogalski:
@yestalgia 
I would love to test the fix, but I am currently on vacation and won't have access to my computer for a couple of months. Sorry :(

[...]

PR #1805, opened 2025-09-15
#1784 feat/persist-feedback-when-changing-input-source-of-vc-controls

Binary-Vanguard-12138:
Summary of Changes:
  - Save feedback values before changing input source value
  - Set feedback values that was saved after changing input source value

Related Issues: #1784


mcallegari:
Thanks

PR #1816, opened 2025-10-06
plugins/dmxusb: add support for usbdmx legacy device

jpue:
(see [here](https://www.qlcplus.org/forum/viewtopic.php?t=18668))

I integrated the AI generated code and refactored it slightly to match the rest of the `dmxusb` plugin/QLC+ source code.

Under Ubuntu (v4/Qt6), it now compiles without error and generally runs.
However, I do not own this DMX device and therefore cannot test the new code in practice.

Currently, only DMX output is supported, while the hardware apparently also supports DMX input.

### DIFF
--- a/plugins/dmxusb/src/usbdmxlegacy.cpp
+++ b/plugins/dmxusb/src/usbdmxlegacy.cpp
@@ -27,8 +27,8 @@
 #define USBDMXLEGACY_CMD_SET_LAST_TX_HIGH_ADDRESS_BIT  0x4F
 
 
-UsbdmxLegacy::UsbdmxLegacy(DMXInterface *interface, quint32 outputLine)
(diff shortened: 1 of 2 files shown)


mcallegari:
Thanks 👍

jpue:
When I saw that the Qt5 build failed, I realised that I had forgotten to push the corresponding patch in October, as I had already created this PR and did not want to trigger the Actions build again.

Only for the Qt5 build, in `UsbdmxLegacy::writeUniverse`,

```
uchar &oldVal = reinterpret_cast<uchar&>(m_portsInfo[0].m_universeData[i]);
```

must be replaced with

```
QByteRef oldVal = m_portsInfo[0].m_universeData[i];
```
.

PR #1828, opened 2025-10-10
Update IMG-Stageline-PARL-4Set.qxf

mitscherdinger:
## Description

**Summary of Changes:**
It's better to default this channel to 255 cause otherwise it stays dark until you change it manually.

## Checklist
  - [ * ] Linux
  - [ ] Windows
  - [ ] macOS

## Additional Notes

Just a small fix…

---

Thank you for your excellent software!


mcallegari:
I don't think this is needed.
The concept of "shutter closed" comes from discharge lamp fixtures that have a real mechanical shutter.
Then for historical reasons they brought it also to LED fixtures which don't actually need it but some consoles consider it as normal control.
Also, you don't want your light at full dimmer by default!
So I'll leave it as it is.

PR #1829, opened 2025-10-10
Update Cameo-CLMPAR3.qxf

mitscherdinger:
Just a little fix:
Height and Depth were switched.
Added 25° of beam angle.


mcallegari:
Thanks :+1:

PR #1831, opened 2025-10-11
fixtures: Stairville: WGF-2000: Fix typos

cmuellner:
Fix typos in the definition of the Stairville Water Ground Fog 2000:
* Model name is WGF-2000 (as defined by the manual)
* Replace "Van Intensity" with "Fan Intensity"
* Weight is 24 kg (as specified in the manual)


cmuellner:
Adding correction of weight.

mcallegari:
Thanks :+1: 
I'll regenerate the fixture map too since model name changed

PR #1714, opened 2025-04-10
engine: speedup loadXML methods by using QStringLiteral

shaforostoff:
reduce number of temporary allocations when parsing xml

On my old i3-3110M, with 1615 fixtures in ~/.qlcplus/fixtures main/qlcplus -w -p --open ~/test.qxw -platform vnc:mode=websock

Before:
heaptrack reports 1278772 temporary allocations via QString -> QArrayData::allocate2 QLCInputChannel::loadXML took 1.1 seconds accumulative (~30000 calls) QLCInputProfile::loadXML took 35 ms accumulative (38 calls)

After:
heaptrack reports 325342 temporary allocations via QString -> QArrayData::allocate2 QLCInputChannel::loadXML took 0.79 seconds accumulative (~30000 calls) QLCInputProfile::loadXML took 31 ms accumulative (38 calls)

On Raspberry Pi 4, with only 3 custom fixtures
main/qlcplus -w -p --open ~/test.qxw -platform vnc:mode=websock

Before:
QLCInputProfile::loadXML took 94 ms cold start and 57 ms hot start accumulative (38 calls)

After:
QLCInputProfile::loadXML took 89 ms cold start and 53 ms hot start accumulative (38 calls)

PR #1835, opened 2025-10-14
Change remaining #define-d QStrings to QStringLiterals

jpue:
#1714 replaced many occurrences of `#define […] QString("[…]")` with the `QStringLiteral` macro to reduce the number of temporary allocations when working with constant strings. This resulted, according to the original PR, in a measurable performance improvement.

As [this](https://github.com/mcallegari/qlcplus/pull/1714#issuecomment-2800760349) comment suggests, this _could_ be applied to all `#define`-d `QString`s, but was not done in the original PR.

This PR completes the previously mentioned work from #1714.


mcallegari:
Thanks! 👍

PR #1841, opened 2025-10-19
qmlui: fix more Qt6 warnings

jpue:
This PR fixes some more QML signal handlers that emit the following warning when called:

```
qrc:/[...] Parameter "[...]" is not declared. Injection of parameters into signal handlers is deprecated. Use JavaScript functions with formal parameters instead.
```

### DIFF
--- a/qmlui/qml/ColorToolFull.qml
+++ b/qmlui/qml/ColorToolFull.qml
@@ -128,8 +128,8 @@ Rectangle
                 toolColorChanged(currentRGB.r, currentRGB.g, currentRGB.b, currentWAUV.r, currentWAUV.g, currentWAUV.b)
             }
 
-            onPressed: setPickedColor(mouse)
-            onPositionChanged: setPickedColor(mouse)
+            onPressed: (mouse) => setPickedColor(mouse)
+            onPositionChanged: (mouse) => setPickedColor(mouse)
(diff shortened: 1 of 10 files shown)


jpue:
When implementing these changes, I followed the steps below to ensure that I made the changes correctly.

1. Notice the warnings on an unpatched build (steps to reproduce)
- DMX Address Tool: Start a new QLC+ instance, open the tool and click on one of the “dip switches”.
- External Control Delegate: Start a new QLC+ instance, switch to the virtual console and add a button, go to its external control settings, add an external controller input, then click on the “Control” combo box and select the only available option, “Pressure”.

2. Look at the definition of the UI elements and the function signature of their signal handlers
- DMX Address Tool: [DMXAddressWidget](https://github.com/mcallegari/qlcplus/blob/master/qmlui/qml/DMXAddressWidget.qml#L38)
- External Control Delegate: [CustomComboBox](https://github.com/mcallegari/qlcplus/blob/master/qmlui/qml/CustomComboBox.qml#L56)

3. Add the “formal parameters” to the “JavaScript functions” and repeat step 1 to verify the changes

However, I can also simply remove these two changes from the PR if you prefer.

mcallegari:
All good then, thanks :+1:

Issue #1808, opened 2025-09-27
Guideline for using the C++11 override and final specifiers


jpue:
While reading through QLC+ code, I recently came across [this](https://github.com/mcallegari/qlcplus/blob/master/engine/src/scenevalue.h#L73) line in the engine code.

When I took a closer look at the `override` and `final` specifiers (introduced with C++11), I came to the conclusion that it might be advantageous to…

- add the `override` specifier to all (virtual) member function of derived classes
(when compiling with gcc, adding `-Werror=suggest-override` helps to identify the relevant functions).
→ forces an error at compile time, if the signature of the overriding function differs from the signature of the function to be overridden in the base class

- add the `final` specifier to all classes, from which no other class is derived
(when compiling with gcc, adding `-Wsuggest-final-types` helps to identify the relevant classes).
→ forces an error at compile time, if a newly introduced class attempts to derive from an existing (base) class, giving the developer the opportunity to change the base class accordingly (e.g., declare the destructor as virtual)

However, before opening a corresponding PR, I would like to know the general guideline/opinion of this project on the use of those C++ features.

mcallegari:
See this discussion
https://github.com/mcallegari/qlcplus/pull/1708#issuecomment-2800740653

In general, at the moment I don't have the capacity to deal with this type of changes.

jpue:
Ok, thanks for the honest answer.

PR #1842, opened 2025-10-19
introduce the C++11 override and final specifiers

jpue:
As discussed in #1808 (and mentioned [here](https://github.com/mcallegari/qlcplus/blob/master/engine/src/scenevalue.h#L73)), it might be beneficial to start using the `override` and `final` specifiers (introduced with C++11) in the QLC+ source code.

This PR add these specifiers in the following ways:
- add the `override` specifier to all (virtual) member function of derived classes
(when compiling with gcc, adding `-Werror=suggest-override` helps to identify the relevant functions).
→ forces an error at compile time, if the signature of the overriding function differs from the signature of the function to be overridden in the base class

- add the `final` specifier to all classes, from which no other class is derived
(when compiling with gcc, adding `-Wsuggest-final-types` helps to identify the relevant classes).
→ forces an error at compile time, if a newly introduced class attempts to derive from an existing (base) class, giving the developer the opportunity to modify the base class accordingly (e.g., declare the destructor as `virtual`)

These changes do not affect the behaviour, appearance or functionality of the program and serve as additional safety precautions for developers.

### DIFF
--- a/engine/audio/plugins/mad/audiodecoder_mad.h
+++ b/engine/audio/plugins/mad/audiodecoder_mad.h
@@ -43,7 +43,7 @@ extern "C"
  * @{
  */
 
-class AudioDecoderMAD : public AudioDecoder
+class AudioDecoderMAD final : public AudioDecoder
 {
     Q_OBJECT
     Q_INTERFACES(AudioDecoder)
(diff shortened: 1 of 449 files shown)


mcallegari:
@jpue sorry for the long delay. Can you please rebase this to master? Thanks

jpue:
> @jpue sorry for the long delay. Can you please rebase this to master? Thanks

No problem, I have just rebased this PR and adapted it to the other code changes that have been made since then.

mcallegari:
Thanks for this :+1:

PR #1848, opened 2025-10-28
ui/dmxmonitor: update gel color immediately

jpue:
2. Add a `generic dimmer`, switch to the `Simple Desk` page and set the dimmer channel to `255`.
3. Open the `DMX Monitor`, switch to the `2D View` and add the fixture.
4. In the DMX monitor, click on the added fixture to open the properties sub-window.
5. a) Click on the `Gel color` colour picker, choose a colour other than white and click `OK`.
or
b) If a colour (other than white) has been chosen, click the `X` button to reset it to white/no gel colour.

**Expected behaviour**
In the `2D View`, the gel colour should updated immediately. Currently, however, it only changes after modifying the intensity channel value (e.g. via the `Simple Desk`).

**Problem Analysis**
When clicking the gel colour picker or the reset button, `MonitorFixturePropertiesEditor::slotGelColorClicked()` or `MonitorFixturePropertiesEditor::slotGelResetClicked()` is called, respectively.
There, the `m_getColor` attribute of the corresponding `MonitorFixtureItem` is set correctly.
However, this change is not immediately displayed because `MonitorFixtureItem::slotUpdateValues()` is only called when the DMX value changes, not when the gel colour is modified.

**Proposed Solutions**
Once all variables have been set correctly, call `MonitorFixtureItem::slotUpdateValues()` from `MonitorFixturePropertiesEditor::slotGelColorClicked()` and `MonitorFixturePropertiesEditor::slotGelResetClicked()`.

The problem with this is that `slotUpdateValues()` is private within `MonitorFixtureItem`.
Instead of declaring this method public for all classes, `MonitorFixturePropertiesEditor` was declared a friend class of `MonitorFixtureItem`, giving it access to all private members of `MonitorFixtureItem`.

### DIFF
--- a/ui/src/monitor/monitorfixtureitem.h
+++ b/ui/src/monitor/monitorfixtureitem.h
@@ -86,6 +86,8 @@ class MonitorFixtureItem : public QObject, public QGraphicsItem
     Q_OBJECT
     Q_INTERFACES(QGraphicsItem)
 
+    friend class MonitorFixturePropertiesEditor;
+
 public:
     MonitorFixtureItem(Doc *doc, quint32 fid);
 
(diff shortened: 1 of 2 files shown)


mcallegari:
I'll check it later today, thanks!
I love surgical fixes 😉

mcallegari:
Thanks :+1:

PR #1852, opened 2025-11-01
ui/{addresstool,addrgbpanel,efxeditor}: fix tab order

jpue:
The `Address Tool`, `Add RGB Panel` dialog and the `EFX` function editor have strange tab orders due to the order in which their elements are defined in the `.ui` files.

This PR uses the `<tabstops>` XML element to set those tab orders to more reasonable ones.

This is not a new feature, but an improvement in usability/user experience.
_(The EFX Editor in particular cost me some valuable time before a show when I needed to quickly change some values.)_

### DIFF
--- a/ui/src/efxeditor.ui
+++ b/ui/src/efxeditor.ui
@@ -727,6 +727,38 @@
    </item>
   </layout>
  </widget>
+ <tabstops>
+  <tabstop>m_speedDial</tabstop>
+  <tabstop>m_testButton</tabstop>
+  <tabstop>m_nameEdit</tabstop>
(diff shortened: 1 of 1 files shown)


ui/src/addresstool.ui
```diff
@@ -114,7 +114,7 @@
      <item row="2" column="1" colspan="4">
       <widget class="QCheckBox" name="m_reverseHorizCheck">
        <property name="text">
-        <string>Reverse Horizontally</string>
```
mcallegari: Please revert this as it impacts on several language translations

mcallegari:
Thanks :+1:

PR #1862, opened 2025-11-08
webaccess: fix synchronisation of RGB matrix colours

jpue:
#1422 (b3094b5826cd42072257c2f158719a5da5a5dcd1) introduced the option to use up to five colours in RGB matrix functions/widgets to create more interesting light effects. However, two sections in the `webaccess` code were overlooked during these changes, resulting in the malfunction described below.

- Start QLC+ (v4, new workspace) with the command line parameter `-w` to enable the `webaccess`.
- Prepare a simple workspace and virtual console to test the RGB matrix functionality (add some fixtures, add them to a fixture group, create a new RGB matrix function, add the fixture group to the RGB matrix function, add a RGB Matrix widget to the virtual console, link the RGB matrix function to the widget, switch to `Operate` mode).
- Open a web browser and navigate to `http://127.0.0.1:9999` to load the `webaccess` page.
- _In the web browser_, click on one of colour pickers of the RGB Matrix widget and select a colour.
→ The colour is sent correctly to the local/native UI.
- _In the local/native UI_, click on one of colour pickers of the RGB Matrix widget and select a colour.
→ The colour is not sent to the web interface. Only after a page refresh (during which the currently selected colour is statically embedded in the page), the new colours will be displayed on the web interface.

**Expected behaviour**
When a colour is selected in the local virtual console, it should be sent to the web interface and displayed there.

**Problem Analysis**
The aforementioned PR #1422 changed a lot of `engine`, `ui` and `webaccess` code to support five matrix colours instead of just a start and an end colour.

To achieve this, the signals sent by the [`VCMatrix`](https://github.com/mcallegari/qlcplus/commit/b3094b5826cd42072257c2f158719a5da5a5dcd1#diff-248ac438bb862bfb02af2334a6eb34514a22a5917a7d5a71bf77ef1e9cbb58af) class were changed.
However, on the server side of `webaccess`, it was forgotten to connect these new signals to the new slots.

Furthermore, commit b3094b5826cd42072257c2f158719a5da5a5dcd1 also introduced a “copy-and-paste” mistake in the [client-side code]( https://github.com/mcallegari/qlcplus/commit/b3094b5826cd42072257c2f158719a5da5a5dcd1#diff-67e18f7837a0458674698977a3119313f2969fe68c7b14caebb0bf070c12a2a2), which further prevented the colours from being displayed.

**Proposed Solution**
Connect the signals correctly to the slots and call the correct functions in the JavaScript code

### DIFF
--- a/webaccess/src/webaccess.cpp
+++ b/webaccess/src/webaccess.cpp
@@ -2082,7 +2082,7 @@ QString WebAccess::getMatrixHTML(VCMatrix *matrix)
                 "id=\"msl" + QString::number(matrix->id()) + "\" "
                 "oninput=\"matrixSliderValueChange(" + QString::number(matrix->id()) + ");\" ontouchmove=\"matrixSliderValueChange(" + QString::number(matrix->id()) + ");\" "
                 "style=\"width: " + QString::number(matrix->height() - 20) + "px; "
-                "margin-top: " + QString::number(matrix->height() - 10) + "px; margin-left: 25px; \""
+                "margin-top: " + QString::number(matrix->height() - 10) + "px; margin-left: 25px;\" "
                 "min=\"1\" max=\"255\" step=\"1\" value=\"" + QString::number(matrix->sliderValue()) + "\">\n";
         str +=  "</div>";
     }


mcallegari:
Thanks. I wonder if we actually need all that code duplication!
I'd replace all `setMatrixColor*Value` with one method with an additional index parameter:

```
function setMatrixColorValue(idx, id, color) {
  var combo = document.querySelector("#mc" + idx + "i" + id);
  combo.value = color;
}
```
Perhaps even `matrixColor*Change` signals can be improved (and I would add a 'd')

```
function matrixColorChanged(idx, id) {
  var colorObj = document.querySelector("#mc" + idx + "i" + id);
  var colorMsg = id + "|MATRIX_COLOR_CHANGE|COLOR_" + idx + "|" + hexToUint(colorObj.value);
  websocket.send(colorMsg);
}
```

mcallegari:
Thanks. This is OK for now but I will optimize also the signals/slots.
There's too much code duplication for no reason when a simple `int index` parameter can be used.

PR #1863, opened 2025-11-08
qmlui/vcclock: prevent underflow/negative time if countdown is started at 00:00:00.0

jpue:
If a VC clock is started in countdown mode from the time `00:00:00.0` in QLC+ 5, the internal time value is decreased before checking whether it is equal to `0`. As a result, it jumps to a large negative number (underflow) that is counted up.

This PR introduces a check before starting the countdown to prevent the undesirable behaviour described above.

### DIFF
--- a/qmlui/qml/virtualconsole/VCClockItem.qml
+++ b/qmlui/qml/virtualconsole/VCClockItem.qml
@@ -109,12 +109,15 @@ VCWidgetItem
             {
                 anchors.fill: parent
                 acceptedButtons: Qt.LeftButton | Qt.RightButton
-                onClicked:
+                onClicked: (mouse) =>
                 {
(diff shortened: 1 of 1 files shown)


qmlui/qml/virtualconsole/VCClockItem.qml
```diff
@@ -109,12 +109,15 @@ VCWidgetItem
             {
                 anchors.fill: parent
                 acceptedButtons: Qt.LeftButton | Qt.RightButton
-                onClicked:
+                onClicked: (mouse) =>
```
mcallegari: mouse is not used within the signal so I don't think it is needed

mcallegari:
Thanks 👍

PR #1876, opened 2025-11-21
Add beattracking class

Wazzledi:
Added first version of the beattracking class

## Description

**Summary of Changes:**

This pull request adds the beattracking class used for automatic beat tracking and tempo estimation, as discussed in the forum [Link](https://www.qlcplus.org/forum/viewtopic.php?t=17016). It is based on the algorithm presented in [this](https://aubio.org/phd/thesis/brossier06thesis.pdf) dissertation by Paul M. Brossier.

**Related Issues:**

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS 

## Testing

**Test Cases:**

No test cases are implemented yet.

**Test Results:**

## Additional Notes

---

Thank you for your contribution to QLC+!

### DIFF
--- a/engine/audio/src/beattracking.cpp
+++ b/engine/audio/src/beattracking.cpp
@@ -35,7 +35,6 @@ const double filterCoeffB[] = {0.15998789, 0.31997577, 0.15998789};
 BeatTracking::BeatTracking(int channels, QObject *parent)
     : QObject(parent)
 {
-    
     m_channels = channels;
     m_sampleRate = BEAT_DEFAULT_SAMPLE_RATE;
     m_windowSize = BEAT_DEFAULT_WINDOW_SIZE;
@@ -55,19 +54,19 @@ BeatTracking::BeatTracking(int channels, QObject *parent)
     m_continuityDerivation = targetLag/8;
 
     m_fftInputBuffer = new double[m_windowSize];
-    #ifdef HAS_FFTW3
+#ifdef HAS_FFTW3
     m_fftOutputBuffer = fftw_malloc(sizeof(fftw_complex) * m_windowSize);
     m_planForward = fftw_plan_dft_r2c_1d(m_windowSize, m_fftInputBuffer, reinterpret_cast<fftw_complex*>(m_fftOutputBuffer), 0);
-    #endif
+#endif
 
     m_raileighFilterBank = getRaileighFilterBank(m_onsetWindowSize, targetLag);
 
     m_currentPredictionState = PredictionState::ACF;
 
     qDebug() << "Channels: " << m_channels
-                << "Sample rate:" << m_sampleRate
-                 << "Window size:" << m_windowSize
-                 << "Hop size:" << m_hopSize;
+             << "Sample rate:" << m_sampleRate
+             << "Window size:" << m_windowSize
+             << "Hop size:" << m_hopSize;
 }
 
 BeatTracking::~BeatTracking()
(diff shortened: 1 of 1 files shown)


mcallegari:
Thanks for this.
I have a few questions:
- can you explain how it works?
- since it is based on a research, can I assume the code is not optimized?
- what is the entry point of the audio buffer? processAudio?
- are there hardcoded tuning parameters that might be used to tweak the algorithm?

Wazzledi:
The code is based on the research, but is also implemented in the aubio library. So, existing code exists for this algorithm, which helped me implement this efficiently. However, it must be tested if it performs well enough in low spec environments. I tried to implement it as efficient as possible. But i did not evaluate the performance on different devices yet.

The entry point is the processAudio method. It takes the same parameters as the processAudio method in the AudioCapture class. From line 132 to 141, the method performs the same audio mixdown as the AudioCapture class does (might be removed, depending on where the algorithm is included).

Yes, there are hardcoded parameters. They can be changed in the beattracking.h file (line 32-36). But, these parameters are the recommended parameters from the dissertation and the existing code base. Some parameters might be useful to add. I will add some more settable parameters in the future.

Explanation:

1. The audio is split into frames of 1024 entries. They overlap with a size of 512. Then the FFT is calculated on each window.
2. The magnitudes of the FFT frame are calculated, and a basic onset function is applied (right now spectral flux). The dissertation discussed different onset functions which might be implemented in the future.
3.  The onset function is filtered and peak picked with the proposed Real-Time Peak Picking procedure. These peak picked values are stored in m_tOnsetValues for tempo estimation and beattracking.
4. Once 512 peak picked values are stored, an auto correlation is calculated on the values to get reoccurring distances between onsets. 
5. On top of this auto correlation, a raileigh filter is applied to favor results near the "RAILEIGH_TARGET_BPM". From this auto correlation, we pick the index of the highest value to be our calculated lag. This is our estimated tempo in number of frames.
6. Now the algorithm has two different states. The first one is the PredictionState::ACF, which calculates the estimated tempo by the method above (using reileigh filter). When 3 continuous lags are detected, the algorithm switches in a continuity mode labeled as  PredictionState::CONTINUITY.
7. When the algorithm is in the continuity mode it calculates a gaussian weighting on top of the auto correlation function calculated in 4. (without raileigh filter), so that a peak around the last estimated tempo is picked. Every time the lag is chosen with the continuity mode, the algorithm checks if the value chosen in 5. derives significantly from the continuity value. If it deviates significantly, the algorithm goes back to PredictionState::ACF.
8. Once a lag is chosen by either mode, the algorithm calculates the current phase of the frame depending on the estimated tempo. With this information, we can predict the frames where the next beats will occur. Normally, there would be two beat detection methods too, but I did not implement the continuity mode for this function yet.
9. Since we have estimated the next occurring beats, we check every frame if this specific frame has a beat. If so, we can trigger a beat in the engine (See line 335). 

Let me know if anything is unclear.

What needs to be implemented in the future:
- The context dependent method for beat tracking
- Silence gate as proposed in the dissertation
- Maybe mapping FFT frequencies to Mel frequencies (not mentioned in the research, but used in some other research algorithms)

mcallegari:
Thanks a lot for the explanation.
One last question. Can you explain what "The context dependent method for beat tracking" means?
For silence gate, can this work?
https://github.com/mcallegari/qlcplus/blob/master/engine/audio/src/audiocapture.cpp#L201

mcallegari:
Another question: does it consider the sample rate fixed to 44.1Khz?
I'm seeing fixed numbers indicating that

Wazzledi:
> Thanks a lot for the explanation. One last question. Can you explain what "The context dependent method for beat tracking" means?

As is explained, the whole implementation does use two states. These two states apply to the tempo estimation and the beattracking i.e. finding the phase of the current frame in the predicted beats. In the first state, we do not have any additional information to calculate the tempo and the beat phase. In the second state (the context dependent method), we have the previously identified tempo and beat phase, and we search near this information. Right now, I only implemented this second method for tempo estimation and not the beat phase detection.

> For silence gate, can this work? 

I did not read through the silence gate part of the research yet. I saw that you implemented some sort of silence filtering. I assume it will be something like that. But I have to read this part of the dissertation again to truly answer this question.

> Another question: does it consider the sample rate fixed to 44.1Khz?
I'm seeing fixed numbers indicating that

Yeah, I see. That is indeed an issue. I forgot to move this to a parameter. Feel free to change this.

PR #1880, opened 2025-11-23
correct fixture types, part 2: single color fixtures (also set to typ…

edogawa23:
…e Color Changer as this seems to be common sense)

OK, I've looked at how some other single color LED fixtures like e.g. Cameo TS (40, 100 and 200 WW variants) are done, and it seems to be the widely accepted and preferred way to go.
So I changed the remaining items on my list and set them all to type "Color Changer" now.

This fixes the 2D/3D fixture rendering as discussed in the forum thread mentioned in my previous "part 1" PR.

### DIFF
--- a/resources/fixtures/American_DJ/American-DJ-Encore-FR150z.qxf
+++ b/resources/fixtures/American_DJ/American-DJ-Encore-FR150z.qxf
@@ -8,7 +8,7 @@
  </Creator>
  <Manufacturer>American DJ</Manufacturer>
  <Model>Encore FR150z</Model>
- <Type>Dimmer</Type>
+ <Type>Color Changer</Type>
  <Channel Name="Master Dimmer" Preset="IntensityMasterDimmer"/>
  <Channel Name="Strobe" Default="40">
   <Group Byte="0">Shutter</Group>
--- a/resources/fixtures/American_DJ/American-DJ-Encore-FR50z.qxf
+++ b/resources/fixtures/American_DJ/American-DJ-Encore-FR50z.qxf
@@ -8,7 +8,7 @@
  </Creator>
  <Manufacturer>American DJ</Manufacturer>
  <Model>Encore FR50z</Model>
- <Type>Dimmer</Type>
+ <Type>Color Changer</Type>
  <Channel Name="Master Dimmer" Preset="IntensityMasterDimmer"/>
  <Channel Name="Strobe" Default="40">
   <Group Byte="0">Shutter</Group>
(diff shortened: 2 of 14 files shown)


mcallegari:
I don't understand. Why a fixture like Cameo F2 T PO should be a color changer? I see no color channels or color wheels there.

edogawa23:
It's about the forum post I linked in my previous PR.
The only reason is to make these fixtures look ok in 2D/3D preview.
Currently the Fixture type is set to "Dimmer" which causes this.
I tried to ask you what type would be better to choose, there's only "Other" or "Color Changer" available, the others don't make sense. Your answer was confusing and bringing in multihead fixtures.

I'd happily choose "Other" but anyways, I then looked and found a few other fixtures like the Cameo TS 40/100/200 WW which also are set to Color Changer, so I thought that's the reasonable thing to do.

mcallegari:
I've read the forum post and the fact some fixtures are set as "Dimmer" is not always a mistake you need to correct.

The Cameo fixture I mentioned above is one example. When a fixture doesn't produce a color, then it's not a color changer.
Can we at least agree on this?

In the 3D preview, dimmers are represented as a generic PAR can, and there you can set a gelatine to change the light beam color.
A multi-head dimmer can also be this (and here again you'll see a few generic PAR cans)
https://www.huss-licht-ton.de/images/product_images/large_images/11/12641_0.jpg

If you instead think warm/cold white are colors, then we're talking about a totally different thing and at the moment the 2D/3D previews are not handling CTO at all.

I hope my reply doesn't confuse you this time.

TL;DR: this PR is not OK

edogawa23:
I understand all you say very well. I also don't think WW + CW should count as one color.

Now in turn, you didn't answer what type these fixtures shold be set to instead. If Dimmer, they show multiple lamp heads/par can representations in 2/3D, if set to any other than 1 ch mode. This case seems not to be covered by existing types, so maybe bring in an additional one (1 color multi channel fixtures)?

If not, I don't mind if this PR is rejected but I still think this should be resolved in a way or another. If "Other" is more appropriate, I'm perfectly happy with that. :) Just be aware that there are several fixtures in the library that should be revisited because they produce only WW or CW or UV, and are set to type "Color Changer", and obviously got accepted in the past. 
type dimmer
type not dimmer

See how the lower four fixtures are represented well, the upper three less so. I've changed my local Cameo F2 profile to be of type Other, just to show it's also fine. It also can preview the strobe setting, while the Dimmer type fixtures just brighten the 3rd channel. I definitely don't want that appearance in preview.

I attach the workspace with an .xml extension appended so I can upload it here:
[1-color fixtures - different types.qxw.xml](https://github.com/user-attachments/files/23755912/1-color.fixtures.-.different.types.qxw.xml)

PR #1884, opened 2025-11-26
webaccess/css: improve consistency across different browsers

jpue:
When using the webaccess in different browsers, sliders look very different.

webaccess

This PR fixes this by adding CSS rules to apply the custom slider style in Firefox as well.

While at it, this PR also formats all CSS files consistently, fixes some CSS syntax errors and replaces some prefixed (vendor-specific) CSS properties (which were necessary back in 2012-2014 when the code was originally written) with newer, standardised properties.
Of course, this could affect backwards compatibility. However, all popular browsers released after around 2015 support the features required here. I can provide a detailed analysis of this, if desired.
Overall, this not only makes the code more readable, but also facilitates future changes/improvements.

Finally, all CSS files have been successfully checked with the [W3C CSS Validation Service](https://jigsaw.w3.org/css-validator/).

### DIFF
--- a/webaccess/res/simpledesk.css
+++ b/webaccess/res/simpledesk.css
@@ -65,3 +65,12 @@ input[type="range"]::-webkit-slider-thumb {
  width: 20px;
  height: 34px;
 }
+
+input[type="range"]::-moz-range-thumb {
+ appearance: none;
+ background-color: #999999;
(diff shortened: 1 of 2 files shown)


mcallegari:
Thanks :+1:

PR #1885, opened 2025-11-26
webaccess: various small improvements

jpue:
_by-product of #1879_

- RGB matrix: allow stopping RGB Matrix functions (previously not possible since the webaccess matrix slider was limited to a minimum value of `1` instead of `0`)
- RGB matrix: update the algorithm list in webaccess when changing the algorithm (both locally and via the web interface) using custom buttons
- webaccess: fix a JavaScript error visible in the browser’s developer tools when moving a slider as sliders were always incorrectly treated as knobs
- simpledesk: check parameters to avoid a `NULL`-pointer dereference and correct generated HTML code
- keypad: fix WebSocket error handling
- remove unnecessary `type="text/javascript"` attribute from `script` tags
- simplify logic statements within `if` statements

Furthermore, I can add the following changes (here or in a separate PR), if desired:
- HTML-encode all user-defined strings embedded in the HTML code to prevent HTML syntax errors and injection attacks; however, this means losing the ability to customise the webaccess UI beyond the official options using the HTML tags `<b>`, `<i>` and `<br>`
- split the DMX keypad (currently a standalone, static HTML file) into a separate CSS and JS file and integrate the HTML code as a `static` function into C++ code source of `webaccess/simpledesk`, enabling the use of some common code as well as the translation of the DMX keypad

### DIFF
--- a/webaccess/res/keypad.html
+++ b/webaccess/res/keypad.html
@@ -93,7 +93,7 @@ function connect()
 
   websocket.onerror = function () {
     console.error("QLC+ connection encountered error. Closing socket");
-    ws.close();
+    websocket.close();
   };
 
   websocket.onmessage = function(ev) {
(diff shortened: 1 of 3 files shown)


mcallegari:
Thanks :+1: 

Just one little note.
Sometimes I prefer "condition == false" instead of "!condition" because it is more readable. 
If I don't see the '!' then I might read the opposite meaning of a condition.
Maybe it's just me getting old though...

PR #1887, opened 2025-11-30
fixtureeditor,ui: make keyboard shortcuts non-translatable

jpue:
_fixes #1886_

This PR removes all translation code (`tr(...)`) from the `setShortcut` calls in [`fixtureeditor::App::initActions()`](https://github.com/mcallegari/qlcplus/blob/master/fixtureeditor/app.cpp#L241) and [`ui::App::initActions()`](https://github.com/mcallegari/qlcplus/blob/master/ui/src/app.cpp#L679) and hardcodes the key sequence strings.

It also cleans up the translation files.


mcallegari:
Looks good, thanks :+1:

PR #1888, opened 2025-12-01
ui: fix audio volume adjustment of a video on intensity change

jpue:
_inspired by https://www.qlcplus.org/forum/viewtopic.php?t=18545 (however, the issue described in this forum post is a Qt problem and cannot be easily fixed by QLC+ alone, see c348c2b4ae3967f39db212e6f96b833f98bfecda)_

1. Start QLC+ (v4, new workspace, compiled with Qt6 (!)) and add a new `Video` function. The selected video does not matter as long as it can be played by QLC+ and contains an audio track.
2. Switch to the `Virtual Console`, add a frame and, within it, a button linked to the `Video` function and a slider in submaster mode.
3. Switch to `Operate` mode and start the video by clicking the button.
4. When changing the slider value, the audio volume does not change unless the slider has a value of `0`, which mutes the audio completely.

**Expected behaviour**
The audio volume should change according to the value of the slider.
(Ideally, the video should also fade out according to the value of the slider, but this feature has been removed with Qt6.)

**Problem Analysis**
In the function [`Video::adjustAttribute`](https://github.com/mcallegari/qlcplus/blob/master/engine/src/video.cpp#L328), the slider value (`[0, 1]`) is mapped to a range of `[-100, 0]`, which is then sent to [`VideoWidget::slotBrightnessAdjust`](https://github.com/mcallegari/qlcplus/blob/master/ui/src/videoprovider.cpp#L313) to adjust, in case of Qt5, the brightness of the video.

However, this `value` has the wrong range for the Qt6 audio volume (`m_audioOutput->setVolume(value + 100)` → range: `[0, 100]`), which requires a value in the range `[0, 1]`.

Furthermore, this value is assumed to be on a linear scale, whereas humans perceive volume on a logarithmic scale. Therefore, the raw slider value needs to be converted to enable smooth volume control. ([source Qt5](https://doc.qt.io/archives/qt-5.15/qmediaplayer.html#volume-prop), [source Qt6](https://doc.qt.io/qt-6/qaudiooutput.html#volume-prop)).

**Proposed Solutions**
Scale the slider value to the correct range and convert it to a linear scale (fixes 2d857c8af552048e4620aab71aaccb730ce58d1b and c348c2b4ae3967f39db212e6f96b833f98bfecda)

### DIFF
--- a/engine/src/video.cpp
+++ b/engine/src/video.cpp
@@ -333,8 +333,7 @@ int Video::adjustAttribute(qreal fraction, int attributeId)
     {
         case Intensity:
         {
-            int b = -100 - (int)((qreal)-100.0 * getAttributeValue(Intensity));
-            emit requestBrightnessAdjust(b);
+            emit requestBrightnessVolumeAdjust(getAttributeValue(Intensity));
             emit intensityChanged();
         }
         break;
--- a/engine/src/video.h
+++ b/engine/src/video.h
@@ -164,7 +164,7 @@ signals:
     void requestPlayback();
     void requestPause(bool enable);
     void requestStop();
-    void requestBrightnessAdjust(int value);
+    void requestBrightnessVolumeAdjust(qreal value);
 
 private:
     /** URL of the video media source */
--- a/ui/src/videoprovider.cpp
+++ b/ui/src/videoprovider.cpp
@@ -117,8 +117,8 @@ VideoWidget::VideoWidget(Video *video, QObject *parent)
             this, SLOT(slotSetPause(bool)));
     connect(m_video, SIGNAL(requestStop()),
             this, SLOT(slotStopVideo()));
-    connect(m_video, SIGNAL(requestBrightnessAdjust(int)),
-            this, SLOT(slotBrightnessAdjust(int)));
+    connect(m_video, SIGNAL(requestBrightnessVolumeAdjust(qreal)),
+            this, SLOT(slotBrightnessVolumeAdjust(qreal)));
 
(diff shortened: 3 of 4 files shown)


mcallegari:
@jpue thanks for this.
Please keep in mind these two things regarding this topic:
- we basically have no more officla builds on Qt5 except for Linux. It's not going to be removed any time soon but let's consider only Qt6 codepath is to be maintained. Linux distros will eventually make Qt6 the default quite soon.
- nice people of Qt company, removed the video brightness adjustment in Qt6, so in v4 it is no longer possible to fade a video. Luckily in v5 the video rendering is handled by QML so I've been able to mimic brightness adjustment with a black overlay (not an ideal solution either)
I remember I studied a solution also for v4 but it looked quite complex and time consuming. IIRC it was about using a GL shader applied to the video overlay. I guess chatgpt could help here if we really want to achieve it.

Question out of topic: I realized I still don't know your name 🙂
I'm asking basically to credit your changes in the changelog and website. If you prefer to use your nickname no problem 🙂

jpue:
Thanks for merging.

I tested a Qt5-based QLC+ build of this branch yesterday and also noticed the build error, but have not had time to push the fix and/or mark this PR as a draft yet.

I accidentally added another bug through this PR related to converting floats to integers. My apologies. I will provide a fix as soon as possible.

In any case, I am going to open a new PR because I have some great news:

I spent some more time working on the brightness adjustment/video fading support for Qt6 and it turns out the solution was simpler than you outlined, but still not trivial.

PR #1889, opened 2025-12-01
engine/{scriptv4,scriptrunner}: fix v4 to v5 script conversion and v5 scriptrunner

jpue:
_Reported: https://www.qlcplus.org/forum/viewtopic.php?t=18935_

2. Add a new `Script` function and paste the following code:
```
systemcommand:cmdline arg:abc arg:def arg:"gh i" arg:"j'k'l m" arg:n"op"q
systemcommand:cmdline arg:"--geometry=45x7" arg:"--hide-menubar" arg:"--" arg:"mpv" arg:"/path/to/my/file.mp3"
```
with `cmdline` being the following helper bash script:
```
#!/bin/bash
echo -ne "$#|" >> "output.txt"
while [[ $# -gt 0 ]]; do
  echo -ne "$1|" >> "output.txt"
  shift
done
echo >> "output.txt"
```
(change the paths as necessary).

3. Run the QLC+ script and verify that the contents of `output.txt` match:
```
5|abc|def|gh i|j'k'l m|n"op"q|
5|--geometry=45x7|--hide-menubar|--|mpv|/path/to/my/file.mp3|
```

4. Save the workspace, close QLC+ (v4) and open/import the workspace into QLC+ v5. Open the script and verify that it matches:
```
Engine.systemCommand("../cmdline abc def 'gh i' 'j'k'l m' n"op"q");
Engine.systemCommand("../cmdline '--geometry=45x7' '--hide-menubar' '--' 'mpv' '/path/to/my/file.mp3'");
```
5. Check the script with the syntax checker.
**Bug A**: The (double) quotes in the last argument of the first line were not escaped during the script conversion from v4 to v5. This is an issue and prevents the script from running.
6. Temporarily fix the problem by placing backslashes in front of the quotes. (Result: `n\"op\"q`). The syntax checker should now report: “No errors found.”
7. Run the QLC+ script and verify that the contents of `output.txt` match:
```
5|abc|def|gh i|j'k'l m|n"op"q|
0|
```
The first line should be correct since it is equivalent to the output from QLC+ v4.
However, the second line (the example from the linked forum thread) is wrong, as it does not pass any of the specified arguments to the bash script.
**Bug B**: If an argument does not contain spaces, but is still enclosed in single quotes (as are all arguments in scripts converted from v4), they are skipped when the script is executed.

**Problem Analysis**
_Bug A_
The conversion code only checks whether an argument begins with quotes and then threats it accordingly. However, quotes contained within an argument are not take into account in any way. [source](https://github.com/mcallegari/qlcplus/blob/master/engine/src/scriptv4.cpp#L485)

_Bug B_
The issue lies in the `ScriptRunner::systemCommand` function. Assuming that the current input for this function is the second line of the above test script, during the first round of [this](https://github.com/mcallegari/qlcplus/blob/master/engine/src/scriptrunner.cpp#L447) loop, the first argument begins with a single quote and is therefore written (with only the first character, the opening quotation mark, removed) to the cleaned `multiPartArg` string. In the next iteration (when `token` contains the second argument, which also begins with a single quote), `multiPartArg` is cleaned and the second argument is written to it. This continues for all arguments.
Therefore, at the end of the loop, no argument has been written into `programArgs` and the system command is launched without any arguments.

**Proposed Solutions**
_Bug A_
Add a `replace` call to also escape quotes in the middle of an argument

_Bug B_
Add another `if`-case before `if (token.startsWith("'"))` (line 450) and change the old line 450 to `else if`
The new `if`-case checks whether the `token` (argument) starts and ends with a single quote (and does not contain a space, which is implicitly given since all `token` were `split` from `command` by a space).
If this is true, simply remove the single quotes at the beginning and end, and then add the resulting string to the `programArgs` list.

### DIFF
--- a/engine/src/scriptrunner.cpp
+++ b/engine/src/scriptrunner.cpp
@@ -447,7 +447,11 @@ bool ScriptRunner::systemCommand(QString command)
     for (int i = 1; i < tokens.size(); i++)
     {
         QString token = tokens.at(i);
-        if (token.startsWith("'"))
+        if (token.startsWith("'") && token.endsWith("'"))
+        {
+            programArgs << token.removeFirst().removeLast();
+        }
+        else if (token.startsWith("'"))
         {
             multiPartArg.clear();
             multiPartArg.append(token.mid(1));


mcallegari:
Thanks :+1:

PR #1895, opened 2025-12-06
ui/scripteditor: improve systemcommand helper

jpue:
_by-product of #1889_

2. Add a new `Script` function, click the “+” button and select “System Command”. Select some executable (i.e. `/usr/bin/true`) and (important!) add some arguments (i.e. `abcd`). Click `OK` to insert the command.
3. Click on the check mark to have QLC+ check the syntax of the script. A syntax error is found in the newly added `systemcommand` line.

**Expected behaviour**
Commands inserted by internal helpers must not contain any syntax errors.

**Problem Analysis**
In the [`ScriptEditor::slotAddSystemCommand()` function](https://github.com/mcallegari/qlcplus/blob/master/ui/src/scripteditor.cpp#L374), all arguments specified in the dialogue box are structured into one line using the format string `"arg:%1 "`. However, the space after the last argument (at the end of the line) is not removed and the syntax checker complains about it.

**Proposed Solution**
`Trim` the final `formattedArgs` string

### DIFF
--- a/ui/src/scripteditor.cpp
+++ b/ui/src/scripteditor.cpp
@@ -396,6 +396,7 @@ void ScriptEditor::slotAddSystemCommand()
     {
         formattedArgs.append(QString("arg:%1 ").arg(arg));
     }
+    formattedArgs = formattedArgs.trimmed();
 
     m_editor->moveCursor(QTextCursor::StartOfLine);
     m_editor->textCursor().insertText(QString("%1:%2 %3\n")


mcallegari:
Thanks 👍 
I guess this doesn't apply to v5 right?

jpue:
No, v5 is generally not affected by this because (a) it does not have those kinds of helper dialogues (it merely inserts an empty version of the command at the current cursor position in the script) and (b) the entire script syntax is similar to JavaScript, which is very tolerant of whitespace characters around commands:

` Engine . systemCommand ( "/usr/bin/sleep 100" ) ; `

does not generate any syntax errors and runs as expected.

I have just tested this again.

However, commands such as

- `Engine.systemCommand(" /usr/bin/sleep 100");`,
- `Engine.systemCommand("/usr/bin/sleep 100 ");` or
- `Engine.systemCommand("/usr/bin/sleep  100");`

do **not** work since in those cases, the `ScriptRunner::systemCommand` function parses the argument string in a way that `QProcess` (or ultimately the kernel) does not accept.

The annoying thing is that it fails silently, without displaying any syntax or runtime error message.

To investigate this, it is best to output the variables `programName` and `programArgs` at the end of `ScriptRunner::systemCommand` to see what is being passed.

I do not know whether there is a specification for how those system commands should be handled, but I would guess that shells like bash do their best to make them work somehow.

If we would like to replicate this behaviour, in the function `ScriptRunner::systemCommand`, a pre-processing (removing whitespaces at the beginning and end) could be applied

```
    // program name and the arguments
    QStringList tokens = command.trimmed().split(" ");
    if (tokens.count() == 0)
```
as well as a post-processing of the QStringList `programArgs`, removing all empty arguments (which are created when command line arguments are separated by more than one space):

```
    QProcess *newProcess = new QProcess();
    programArgs.removeAll(QStringLiteral());
    newProcess->setProgram(programName);
```
I can contribute this via a PR, if you like. However, I think only these two lines are really needed to silently fix many issues with the `systemCommand` input.

PR #1906, opened 2025-12-23
miscellaneous improvements

jpue:
_I know that it is generally a bad idea to put multiple different changes into one PR, but these are all such small changes that I figured splitting them up would be much more confusing. They are separated into commits anyway._

This PR contains some improvements and fixes for small issues I found over the past few months:

- engine: mark the class `SceneValue` class as `final`
The `TODO` has been in there since 2017 (6ee06b743ddac621f84b3f5cd65cbd7e926f1486). C++11 support is nowadays pretty much guaranteed on all platform supported by Qt6 anyway (see #1808 and #1842).

- ui/addfixture: fix “ISO C++ forbids variable length array”
When compiling QLC+ with very strict compiler flags (for example (`gcc`): `-pedantic -pedantic-errors -std=c++26`), this compiler error is generated since `maxChannels` is not “an integral constant expression” ([source](https://en.cppreference.com/w/cpp/language/array.html)). Declaring `maxChannels` as a `const` variable solves the error and is possible because the variable is never assigned a new value after initialisation.

- change more #define-d strings to QStringLiterals
(see #1714 and #1835)

- change many links in comments and documentation from `http` to `https`

- resources/fixtures-tool: fix “SyntaxWarning: invalid escape sequence '\d'”

Since [Python 3.6](https://docs.python.org/3.6/whatsnew/3.6.html#deprecated-python-behavior),
> A backslash-character pair that is not a valid escape sequence now generates a DeprecationWarning.

and since [Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes) (*),

> A backslash-character pair that is not a valid escape sequence now generates a SyntaxWarning, instead of DeprecationWarning.

That is why, when running `make check` on a system with Python 3.12 (or newer) install, the following warning is produced:
```
qlcplus/build/resources/fixtures/./scripts/fixtures-tool.py:321: SyntaxWarning: invalid escape sequence '\d'
numversion_tok = re.findall('\d+', version_tag.text)
```
(*) also presents the solution:

> For example, re.compile("\d+\\.\d+") now emits a SyntaxWarning ("\d" is an invalid escape sequence, use raw strings for regular expression: re.compile(r"\d+\\.\d+")).

and continues stating that

> In a future Python version, SyntaxError will eventually be raised, instead of SyntaxWarning.


mcallegari:
Thanks! :+1:

PR #1925, opened 2026-01-11
Update workflows.yml to also trigger qlcplus-qt6-git

edogawa23:
This is to make GH commits trigger the new v4 Qt6 package rebuilds.


mcallegari:
Thanks 👍

PR #1935, opened 2026-01-18
engine,ui: move info page creation from engine to ui

jpue:
_fixes #1915_

**Describe the bug / To Reproduce / Expected behaviour / Problem Analysis**
see linked issue

**Proposed Solution**

As proposed in the linked issue, the most pragmatic solution to this “problem” is to move the functions for creating the info pages from the engine code into the v4 `ui` code, where the translation tools can pick up the corresponding strings.

This is easily possible since the `Fixture` class provides “getter” methods for all attributes that contain the necessary information and can be used as a substitute for direct access to member variables.

In addition, some NULL-pointer case are now explicitly handled and `const` attributes have been added to some local variables (pointers to internal `engine` objects) to mechanically prevent modification from the `ui` code.

For the time being, the corresponding unit test (`Fixture_Test::status()`) has simply been disabled, as there is no suitable place for it within the `ui` tree. Since this test did nothing more than check that the code did not crash, I do not consider this a problem.


mcallegari:
Awesome, thank you :+1: 
In this way the engine gets even lighter for version 5

PR #1934, opened 2026-01-17
use initializer list

jpue:
This PR improves the C++ code quality and performance by using the constructor’s initializer list wherever possible, rather than assigning variables within the constructor body.

PR #1941, opened 2026-01-28
various code improvements

jpue:
This PR contains (code quality) improvements and fixes I found over the past few weeks and months:

- qmlui: finish removing Qt5 support
These pieces seem to have been forgotten when removing Qt5 support for QLC+ 5 (2d74c99c60bf519bcbc207952dcec75c3bcbeee9).

- webaccess: add missing override keywords
This improves the new, refactored and enhanced webaccess code for QLC+ v4/v5 (introduced by #1936).

- engine/qlci18n_test: finish removing qmake
This appears to have been forgotten when removing the `qmake` support (and with it, all `.pro` files) (3ed0ab005d6337973d5fed7d160bf49e59a0ea76).

- engine: use initializer list
This commit continues the work started in #1934 for some more `engine` classes that were previously overlooked.

- {engine,ui,qmlui,plugins}: make more functions const
These commits mark more functions that do not change the state of their own object as `const`, simplifying the use of `const` objects, references, and pointers. The `override` keyword helped to ensure that no overwrites by subclasses were broken.

Another inconsistency in the QLC+ codebase is the use of `#include <qmath.h>` versus `#include <QtMath>`. As far as I understand this, using `qmath.h` is the legacy method (from Qt4 times) and `QtMath` is the more modern approach (for Qt5+).

Similarly, there is the question on whether to use `#include <math.h>` or `#include <cmath>` for the mathematical functions of the standard library. My understanding is that it is best to use the `cmath` version in C++ code, since it takes full advantage of the additional features of C++ (compared to C), such as the `const` keyword, more build-in types, templates and namespaces (source: https://cplusplus.com/reference/clibrary/).

If desired, I will add a commit that unifies those includes after I receive a response about the QLC+ guideline on this subject.

### DIFF
--- a/plugins/E1.31/e131plugin.cpp
+++ b/plugins/E1.31/e131plugin.cpp
@@ -72,7 +72,7 @@ void E131Plugin::init()
     std::sort(m_IOmapping.begin(), m_IOmapping.end(), addressCompare);
 }
 
-QString E131Plugin::name()
+QString E131Plugin::name() const
 {
     return QString("E1.31");
 }
@@ -82,7 +82,7 @@ int E131Plugin::capabilities() const
     return QLCIOPlugin::Output | QLCIOPlugin::Input | QLCIOPlugin::Infinite;
 }
 
-QString E131Plugin::pluginInfo()
+QString E131Plugin::pluginInfo() const
 {
     QString str;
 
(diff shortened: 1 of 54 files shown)


mcallegari:
Thanks 👍

PR #1945, opened 2026-02-03
[Translation] refine German translation

ClarkLiam:
## Description

**Summary of Changes:**
This PR refines the German translation across multiple translation files for increased clarity and precise meanings.

If possible it would be nice if you could review these translation changes @FloEdelmann @jpue @Loewe111

### DIFF
--- a/plugins/uart/UART_de_DE.ts
+++ b/plugins/uart/UART_de_DE.ts
@@ -7,7 +7,7 @@
         <location filename="uartplugin.cpp" line="111"/>
         <source>This plugin provides output for UART devices.</source>
         <oldsource>This plugin provides input/output for UART devices.</oldsource>
-        <translation>Dieses Plugin bietet Ein-/Ausgabe über UART-Geräte.</translation>
+        <translation>Dieses Plugin bietet Ausgabe über UART-Geräte.</translation>
     </message>
 </context>
 </TS>


ClarkLiam:
Thank you @FloEdelmann for the feedback 
On a side note: I find it interesting that you also use the hyphen between stuff like Software-Version, because I'd do the same by default and in the translations it's a bit of both...

plugins/dmxusb/src/DMX_USB_de_DE.ts
```diff
@@ -134,7 +134,7 @@
     <message>
         <location filename="dmxusbopenrx.cpp" line="160"/>
         <source>System Timer Accuracy</source>
-        <translation>System-Timer-Genauigkeit</translation>
+        <translation>System-Takt-Präzision</translation>
```
Loewe111: I would leave it at `Timer`, I think `System-Timer` is clearer than `System-Takt`, i would leave `Präzision`, so `System-Timer-Präzision` (same on line 175)

FloEdelmann:
> you also use the hyphen between stuff like Software-Version, because I'd do the same by default and in the translations it's a bit of both...

Well, that's correct German punctuation :smiley: 
But many people don't know that and/or have adapted to the English style which combines words through space instead of hyphen.

kripton (approved):
Comments from others are valid, fine otherwise. Thanks

mcallegari:
Thanks everyone :+1:

PR #1952, opened 2026-02-08
improve the (v5) webaccess

jpue:
After the extensive rework of the `webaccess` code to create a QLC+ 5 webaccess, I also `const`-ified many functions (which also required some minimal changes to the `qmlui` and `ui` code) and added some `nullptr` checks to make the code more robust (However, I am not sure whether I used the new `CommonRequestResult` correctly for the error case).

Furthermore, two small changes to the JavaScript code add support for VC sliders in percentage and inverted mode.

In addition, the web VC cue list now behaves more like the native one, immediately populating the `bottomLabel` when starting a cue list.

Finally, in the web simple desk, sliders became briefly non-override (not red) as soon as they were dragged to `0`, until the WebSocket message of the client re-set them to override. This flickering effect has now also been fixed.


mcallegari:
Thanks @jpue 👍

Issue #1953, opened 2026-02-08
webaccess-qml: implement more vcclock functionality


jpue:
Currently, only the “standard clock” mode of the `vcclock` works in the new v5 `webaccess`.

Before restoring the missing functionality, I would like to ask how this was planned to be implemented from an architectural point of view.

In the old v4 webaccess, the web client was very simple: It sent out messages about the interaction with the UI (clock start/stop/reset) and relied on the main application to send updated time information via a WebSocket.

At present, only the “standard clock” mode works this way in the v5 webaccess.
The other modes (stopwatch and countdown) receive their initial values from the client via the `vc.json` file, but then the local program and the web app work independently of each other and both implement the corresponding countdown/countup logic on their own. (Not even that works correctly, since the local program sends some time information in milliseconds, while the web app currently interprets everything as seconds.)

Question 1: How (if at all) should those v5 components communicate with each other? Who implements what?

Question 2: The initial values mentioned above are taken from the `VCClock` C++ class, but the actual values shown in the local UI are only stored in the `VCClockItem` QML code. Any suggestions/requests on how these values can be made available for web access in a clean way?

PR #1954, opened 2026-02-10
improve static webaccess code/resources

jpue:
This PR contains several minor fixes that I have collected for the `webaccess`. It mainly focuses on the static/statically served parts of the code and resources (`.js`, `.css` and `.html` files).

- fix HTML and CSS validation errors and improve syntax by formatting the HTML and JavaScript code
- add file type specifiers for file uploads (projects and fixtures)
- keypad: fix the (previously broken) usage of the `resetUniverse` API, add a “clear command” button and also send command when pressing `Enter` in the command input field
- Test_Web_API: fix malformed output of the “getChannelsValues” API function (code inspired by `simpledesk.js`)

### DIFF
--- a/webaccess/res/simpledesk.js
+++ b/webaccess/res/simpledesk.js
@@ -96,7 +96,7 @@ function drawPage(data) {
  var cVars = data.split("|");
  var payloadSize = cVars.length - 2;
  var stride = payloadSize % 4 === 0 ? 4 : 3;
- for (i = 2; i < cVars.length; i+=stride) {
+ for (let i = 2; i < cVars.length; i+=stride) {
      var chNum = parseInt(cVars[i]);
      code += "<div class='sdSlider' style='width: 36px; height: 372px; background-color: #aaa; margin-left: 2px;'>";
      code += getSliderTopCode(cVars[i + 2]);


mcallegari:
Hey @jpue this is surprising.
Can you believe the whole rework and v5 web interface has been written by AI?
Obviously I supervised the whole process and took a little bit but I'm pretty satisfied by the result.
What do you think about the new v5 web interface?
I am surprised it didn't respect standard specifications though!

jpue:
_Please excuse the late reply._

I really like the new v5 web interface, as the visual and functional differences between the web interface and the local user interface are much smaller than in v4, which significantly improves usability. I also appreciate the WebSocket connection indicator in the top right corner, which is very handy when working on/with the web interface.

To be honest, I have not used generative AI tools very often, but I am not surprised that they can do such things nowadays. In my opinion, with such extensive rewrites, it does not matter whether the code was written by humans or AI tools, as long as (importantly!) we understand what the code does and can therefore verify that it does what we expect it to do. On this basis, the code can be refined, edge cases can be eliminated, and missing features can be added.

For this reason, I have created a QLC+ workspace with a virtual console equipped with various widgets and have started testing the new web interface with that. I am sure you have already conducted numerous tests, but further testing certainly cannot hurt, can it?
The results can be found in #1952 and #1953.

On the other hand, _this_ PR makes changes to the old webaccess v4 (mainly shifting spaces to make the generated HTML code more consistent and thus easier to read) as well as to static files (keyboard, Test_Web_API) that were not touched at all during the creation of the v5 web interface – the issues fixed here existed before.

mcallegari:
I completed the review of this PR. Thanks as always for your contribution.
Before merging though, I just have one comment: this PR adds many more translated strings (so more work for translators)
Did you use existing strings from the UI so that translators can copy & paste?

jpue:
The strings related to the web VC cue list were copied (more or less) from the tooltips of the native UI (see [here](https://github.com/mcallegari/qlcplus/blob/master/ui/src/virtualconsole/vccuelist.cpp#L207)). However, I needed to add new strings for the web VC frame, sometimes based on the icon’s file name, sometimes similar to the strings used in the `qmlui`, because the corresponding native UI does not contain any tooltips for its buttons.

In total, this PR “only” adds 10 new string, of which 4-8 can more or less be copy-pasted from the `qmlui` and `ui` translations.

mcallegari:
Thanks 👍

PR #1958, opened 2026-02-13
Fix crash when deleting script runner

daves003:
## Description

Avoid deleting the script runner immediately in the slot connected to finished() .
QThread is not quite done executing, so this crashed. Using deleteLater() is explicitly mentioned by Qt to avoid this problem and only delete the thread when it's fully done.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

### DIFF
--- a/engine/src/scriptv4.cpp
+++ b/engine/src/scriptv4.cpp
@@ -364,7 +364,10 @@ void Script::postRun(MasterTimer* timer, QList<Universe *> universes)
 
 void Script::slotRunnerFinished()
 {
-    delete m_runner;
+    // we can't directly delete m_runner here because the finished() signal of
+    // QThread is fired just before the thread has actually finished executing. This
+    // would crash in ~QThread. deleteLater() ensures the thread is actually done.
+    m_runner->deleteLater();
     m_runner = NULL;
 }
 


mcallegari:
Thanks
@jpue since you worked on #1933, is this somehow related or conflicting? Or perhaps you already observed/solved this?

jpue:
No, I have never encountered any issues related to such crashes while working on the `Script` function in the past months. Therefore, I did not fix anything.

@daves003 Can you provide a concrete example of this issue or _Steps To Reproduce_? Or did you just notice it during code review – be it manual, using a linter, or an AI tool?

Furthermore, #1933 just adapts to `Script` code so that the underlying superclass `Function` is correctly terminated and deleted _after_ everything `Script`-related has already been shut down.

I did a quick test on a branch where both this commit and the one from #1933 had been applied and this seems to work fine as well.

daves003:
I just came across the crash when building some really short looping scripts. So instead of the AppImage I built it locally in debug mode and that showed a `this = 0x0` in the script runner.

Repro that crashes consistently for me:
- New project using QML UI
- Create a script with just this line: `Engine.waitTime(20);`
- Create a chaser with the script as the only step
- Set hold time to 20ms
- Make sure the run order is set to loop
- Preview the chaser
- Should crash almost immediately


#1933 also looks not directly related to me. I've quickly tested a version with both of the changes and it seems to be running fine.

jpue:
Thank you for the information.

Using the steps, I was able to reproduce the behaviour described above.

However, your fix does not solve this issue for me – QLC+ still crashes in various different `Script`-related place (might take a couple of seconds), seemingly at random, whereas previously, it was very localised to destructor of the `ScriptRunner` class, according to some quick debugging.

I was unable to find the root cause in the time available.

daves003:
Interesting, it looked stable for me so far. I'll keep using this branch and see if I can catch crashes in other places.

mcallegari:
Merged, thanks :+1:

PR #1962, opened 2026-02-20
qmlui,ui: improve nested submasters

jpue:
_fixes #1810_

see linked issue

**Problem Analysis / Proposed Solutions**
This issue affects both v4 (`ui`) and v5 (`qmlui`). However, nested submasters work for the most part in v4, whereas they do not work at all in v5.

_For the following explanation, I am assuming that the workspace from the linked issue is being used, or at least the same Virtual Console setup. “Inner” submaster refers to the one inside the frame widget, while “outer” submaster describes the one directly on the Virtual Console._

**`ui` (v4)**
Most of the code to make nested submaster work is already in place. Therefore, the only issue is the edge case that an outer submaster slider can override inner ones.

This can be easily fixed by calling the method `void VCFrame::updateSubmasterValue()` at the end of `void VCFrame::adjustIntensity(qreal val)` (method inherited from `VCWidget` that is used from outside the `VCFrame` class/by other `VCFrame` objects to set the intensity of that object) to restore the restrictions already set by inner submasters.

While at it, some `NULL`-pointer checks were added for additional safety (regarding pointer dereferences).

**`qmlui` (v5)**
To implement this functionality in v5, a new property was added to the `VCFrame` class: `m_submasterValue`, which holds the latest value of a submaster _inside_ the corresponding frame and is set to `1` (neutral element of multiplication of real numbers) by default. The existing `intensity` property (inherited from `VCWidget`) contains the value set for the frame by an _outer_ submaster.

By storing both of those values separately, the intensity value to be set on widgets contained within the frame can be recalculated at any time. This would not be possible if one of the values changes and only the “combined” value (product of those two values) would have been stored.

To catch an intensity change from outside the frame, the method `VCFrame::adjustIntensity(qreal intensity)` is added, which overrides the `void VCWidget::adjustIntensity(qreal val)` method inherited from `VCWidget`. It is similar to the method of the same name in `ui`, but removes the call to `VCFrame::updateSubmasterValue()` and uses the modern `nullptr` instead of `NULL` to check for null-pointers. Furthermore, a filter similar to the one of `ui/VCFrame::updateSubmasterValue()` (which skips all submaster sliders) is added, since setting intensity values on submaster sliders limits their value range and thereby causes unwanted side effects.

The existing `void VCFrame::slotSubmasterValueChanged(qreal)` method is changed accordingly to store new `submaster` values and to use this information when calculating the actual intensity value.

### DIFF
--- a/qmlui/virtualconsole/vcframe.cpp
+++ b/qmlui/virtualconsole/vcframe.cpp
@@ -50,6 +50,7 @@ VCFrame::VCFrame(Doc *doc, VirtualConsole *vc, QObject *parent)
     , m_pagesLoop(false)
     , m_PIN(0)
     , m_validatedPIN(false)
+    , m_submasterValue(1)
 {
     setType(VCWidget::FrameWidget);
     setTotalPagesNumber(1);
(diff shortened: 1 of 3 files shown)


mcallegari:
Thanks a lot as always :+1: 
I'm gonna fix the "submaster not applied on project load" in v5 now

PR #1974, opened 2026-03-05
more code improvements

jpue:
This PR finishes the work from #1842 (introduce `override` and `final`) and #1941 (make more functions `const`).

This should not change the functionality at all, but should better communicate to both the compiler and programmers _which functions modify which objects_, leading to better optimisations and safer code (especially when parallelised).

Furthermore, in some places, uses of `enum` have been made stricter and more explicit and cases of pass-by-value of `QString` objects have been changed to `const QString&` to avoid unnecessary object copies (although Qt seems to already optimise this internally using a COW mechanism).

Finally, thanks to some checks with the `cppcheck` linter, some handwritten code (loops over container) was identified that could be replaced with STL algorithms (`audio/beattracker`) and a few cases where local variables shadowed outer variables were also fixed.

### DIFF
--- a/engine/audio/src/audiocapture.cpp
+++ b/engine/audio/src/audiocapture.cpp
@@ -312,25 +312,22 @@ void AudioCapture::processData()
 #endif
 
     // 5) Fill per-band magnitudes and compute power
+    double pwrSum = 0.;
+    double maxMagnitude = 0.;
+    for (int barsNumber : m_fftMagnitudeMap.keys())
     {
-        double maxMagnitude;
-        double pwrSum;
-        for (int barsNumber : m_fftMagnitudeMap.keys())
-        {
-            maxMagnitude = fillBandsData(barsNumber); // fills & returns max per-band
-            pwrSum = 0.;
-            for (int n = 0; n < barsNumber; n++)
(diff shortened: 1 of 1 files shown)


mcallegari:
@jpue this will take a while. 207 files to review :cry:

mcallegari:
Push it here, thanks
GitHub remembers what I have already reviewed

jpue:
Sorry for the late reply, I have been quite busy this week.

As this PR is starting to get a bit messy: should I rework it to avoid the “correction” commits or are you going to squash-commit it at the end anyway?

mcallegari:
I squashed and merged, thanks.
I'll deal myself with the audiocapture code block as I need to release 5.2.1
Thank you for this huge code review

qmlui/simpledesk.h
```diff
@@ -194,7 +194,7 @@ protected slots:
 
 private:
     FadeChannel *getFader(QList<Universe *> universes, quint32 universeID,
```
mcallegari: @jpue this broke Simple Desk!
Now every time a fader is moved a new generic fader is registered in the universe.
This is a major regression and I need to release 5.2.1
I need to review all the places where you put const where it was not needed
Oh my god this is a disaster

PR #1978, opened 2026-03-13
Updated qmlui and webaccess Catalan and Spanish translations.

sbenejam:
Updated qmlui and webaccess Catalan and Spanish translations.


mcallegari:
Thanks Santiago :+1:

PR #1993, opened 2026-04-19
ui/virtualconsole: fix background colour issues

### LINKED ISSUE #1986: [v4] VC edit mode issues with backgrounds
**Describe the bug**
The newest QLC+ 4 version (4.14.4) introduced some serious issues regarding the backgrounds of:
* virtual console
* frames
* any other widget

When the edit mode is enabled, the backgrounds of those widgets change unexpectedly, reducing their visibility and pauperizing the editing experience. Also, it seems that a layer is always present (even in live mode) for some widgets, such as the slider, covering the background that has been set for that widget. A thin border that has the expected color can still be seen.

**Steps to reproduce the behavior:**
1. Change the VC background,
2. Import sliders, labels and other widgets
3. Change their background
4. Switch to live/edit mode: theose issues are extremely noticeable

Video: https://drive.google.com/file/d/1QpxSue_hwJh63SVV5xtC-G-Z-lM2HhPH/view?usp=sharing

**Expected behavior**
In 4.14.3, this issue was not present. Editing the virtual console in 4.14.4 has became discouraging.

 - OS: Windows 11
 - QLC+ Version 4.14.4
 
> [!note]
> **No theme is applied** via a `qss` file

jpue:
_Reported at https://www.qlcplus.org/forum/viewtopic.php?t=19491, fixes #1986_

QLC+ 4.14.4 unintentionally introduced breaking changes to the background colour/background image of the `Virtual Console` (see forum or linked issue for details).

**Expected behaviour**
The existing functionality of QLC+ 4.14.3 and earlier versions should not be changed (see forum or linked issue for details).

**Problem Analysis**
Since there have been no changes to the relevant parts of the QLC+ source code, the breaking changes are likely located in the Qt framework.

The following versions/platforms seem to be affected: 
- official QLC 4.14.4 on Windows (11) / Qt 6.10.2
- current `master` branch build on Ubuntu 25.04 (Qt 6.8.3), 25.10 (Qt 6.9.2) and 26.04 beta (Qt 6.10.2)

The following versions/platforms do NOT seem to be affected:
- official QLC 4.14.3 on Windows / Qt 6.8.1
- current `master` branch build on Ubuntu 22.04 (Qt 5.15.3)

**Proposed Solutions**
As described in [this](https://wiki.qt.io/How_to_Change_the_Background_Color_of_QWidget#Using_the_Palette) Qt example, the [autoFillBackground](https://doc.qt.io/qt-6/qwidget.html#autoFillBackground-prop) property of `QWidget` seems to be important for a solid background colour to be drawn.
Although the constructor of `VCWidget` also sets this to the correct value (`true`), this property is reset to `false` for the VC main `VCFrame` called `m_contents` at some point after initialisation.
To be honest, I could not find the exact cause of this, but re-setting it to `true` when a new colour or image is set, as well as setting this property for the parent widget (`m_scrollArea`) when it is created ([source for this idea](https://stackoverflow.com/questions/79532988/how-to-make-a-qscrollarea-inherit-its-color-scheme-in-a-qtabwidget-as-if-it-was)), fixes the issues and restores the previous behaviour.

For (Linux) Qt5 builds (which were not affected anyway), this changes nothing.

Please note that I only developed this on Linux, so it still needs to be tested on Windows and macOS builds.

When this PR is merged, the [Qt version of Linux v4 builds in the GitHub Actions](https://github.com/mcallegari/qlcplus/blob/master/.github/workflows/build.yml#L72) can be bumped to `6.10.2` again to match the other platforms.

### DIFF
--- a/ui/src/virtualconsole/virtualconsole.cpp
+++ b/ui/src/virtualconsole/virtualconsole.cpp
@@ -1257,6 +1257,7 @@ void VirtualConsole::slotBackgroundColor()
     {
         if (m_selectedWidgets.isEmpty() == true)
         {
+            contents()->setAutoFillBackground(true);
             contents()->setBackgroundColor(color);
         }
         else
@@ -1287,6 +1288,7 @@ void VirtualConsole::slotBackgroundImage()
     {
         if (m_selectedWidgets.isEmpty() == true)
         {
+            contents()->setAutoFillBackground(true);
             contents()->setBackgroundImage(path);
         }
         else
@@ -1304,6 +1306,7 @@ void VirtualConsole::slotBackgroundNone()
 
     if (m_selectedWidgets.isEmpty() == true)
     {
+        contents()->setAutoFillBackground(true);
         contents()->resetBackgroundColor();
     }
     else
@@ -1609,6 +1612,7 @@ void VirtualConsole::initContents()
     m_contentsLayout->addWidget(m_scrollArea);
     m_scrollArea->setAlignment(Qt::AlignCenter);
     m_scrollArea->setWidgetResizable(false);
+    m_scrollArea->setAutoFillBackground(true);
 
     resetContents();
 }


mcallegari:
Hi @jpue thanks for this.
Unfortunately this patch is not enough to fix the issue.
I am testing on Windows, since it is the most used platform and where the issue is more evident.
Building with Qt 6.11 this is what I get with your patch

Design mode:
{4233E34C-9901-452C-82A6-C41D42A1F9AD}

Operate mode:
{81B97544-5F92-419D-BC63-13ADAE372B38}

3 cases:
- Virtual console background shows only the border in design and full color in operate mode
- a VC frame shows the border in design mode and the border again in operate mode
- a VC Button seems to be OK, since the background painting is customized

I have alreasy spent some time before your patch and decided to release 4.14.5 on Qt 6.9.x until a fix is found

jpue:
Ok, fine.

I am still wondering whether this is an issue that could even be fixed in the QLC+ source code, or rather a change/regression in the Qt library.

It is up to you, of course, if we keep this open as a reminder of an ongoing issue or close it, since it does not actually the problem.

As you mentioned QLC+ 4.14.4 being released with Qt 6.9.x: Unfortunately, I have not had a chance to test QLC+ 4.14.4 (it was just released before a larger show I did), but I remember there were audio issues on Windows with Qt 6.8.2 and higher.
Your bug report ([QTBUG-136057](https://qt-project.atlassian.net/browse/QTBUG-136057)) is still linked on the [QLC Plus 5 build HOWTO](https://github.com/mcallegari/qlcplus/wiki/QLC-Plus-5-build-HOWTO) wiki page.
Although the Qt developers mention a patch they have merged into Qt 6.9.1 (and backported to 6.8.4), the issue is not marked as resolved.
So is the issue still present?

mcallegari:
The v5 video tutorials are made on Windows and the last one I did about audio/video was all fine with Qt 6.10.
I think most of the audio issues have gone in recent Qt versions.
There still might be corner cases with specific sample rates or mono files. Can be resolved by converting the audio files though.
In any case this issue remains, so sooner or later we'll have to sort it out (unless it's a Qt regression....)

Issue #1950, opened 2026-02-06
5.2.0: Build error on Fedora rawhide with gcc 16.0.1


ycollet:
I try to build qlcplus 5.2.0 on Fedora rawhide with gcc 16.0.1.
I met the following error during build:
```
/builddir/build/BUILD/qlcplus-5.2.0-build/qlcplus-QLC-_5.2.0/plugins/peperoni/unix/peperonidevice.cpp:262:56: error: bitwise operation between different enumeration types ‘libusb_request_type’ and ‘libusb_request_recipient’ is deprecated [-Werror=deprecated-enum-enum-conversion]
  262 |                             LIBUSB_REQUEST_TYPE_VENDOR |
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
  263 |                             LIBUSB_RECIPIENT_INTERFACE |
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~  
/builddir/build/BUILD/qlcplus-5.2.0-build/qlcplus-QLC-_5.2.0/plugins/peperoni/unix/peperonidevice.cpp:276:56: error: bitwise operation between different enumeration types ‘libusb_request_type’ and ‘libusb_request_recipient’ is deprecated [-Werror=deprecated-enum-enum-conversion]
  276 |                             LIBUSB_REQUEST_TYPE_VENDOR |
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
  277 |                             LIBUSB_RECIPIENT_INTERFACE |
      |                             ~~~~~~~~~~~~~~~~~~~~~~~~~~  
/builddir/build/BUILD/qlcplus-5.2.0-build/qlcplus-QLC-_5.2.0/plugins/peperoni/unix/peperonidevice.cpp: In member function ‘virtual void PeperoniDevice::run()’:
/builddir/build/BUILD/qlcplus-5.2.0-build/qlcplus-QLC-_5.2.0/plugins/peperoni/unix/peperonidevice.cpp:399:60: error: bitwise operation between different enumeration types ‘libusb_request_type’ and ‘libusb_request_recipient’ is deprecated [-Werror=deprecated-enum-enum-conversion]
  399 |                                 LIBUSB_REQUEST_TYPE_VENDOR |
      |                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~^
  400 |                                 LIBUSB_RECIPIENT_INTERFACE |
[...]

PR #1994, opened 2026-04-19
fix builds with gcc 16

jpue:
_Originally reported by @kripton, https://bugs.gentoo.org/show_bug.cgi?id=959507 / https://www.qlcplus.org/forum/viewtopic.php?t=18575_

_Although #1950 also first occurs on `gcc` 16 builds, it is unrelated to the errors described here since the underlying issue of #1950 originates from `libusb`, a dependency of QLC+, and not from the QLC+ source code itself._

Build the `master` branch of QLC+ using `gcc` 16; both `v4`/`ui` and `v5`/`qmlui` fail at some point

**Expected behaviour**
The builds should run without error (and in the best case without warnings).

**Problem Analysis**
`gcc` 16 introduces the `-Wsfinae-incomplete` warning (which is enabled by default) and with that, more checks in _SFINAE_ situations ([source 1](https://gcc.gnu.org/pipermail/gcc-patches/2025-June/686777.html), [source 2](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-Wsfinae-incomplete)).
In this case (QLC+), it looks like the includes in some Qt/CMake generated `mocs_….cpp` files occur in an order where some classes have been defined incompletely (`class ABC;`), leading to _SFINAE_ inside the Qt code, whilst they are later defined completely within the same compilation unit (by including the corresponding `.h` file).

Moreover, `gcc` complains about a variable that, although defined and initialised, is not used anywhere in the code. It can be safely removed to fix this additional error.

**Proposed Solution**
Include some additional header files to define the corresponding classes directly completely

**DISCLAMER:** I am not knowledgeable enough about the inner workings of C++ compilers to judge whether this is a genuine issue in the QLC+ source code or (as yet) a compiler error.
I know that nowadays, it is generally very hard to blame the compiler for this, which are probably one of the best (unit-)tested software components out there.
Both the original report of this issue (which was made on Gentoo Linux) and my own tests (on Ubuntu 26.04 Beta) used preview builds of `gcc` 16, which is currently still in development.
Therefore, there is a possibility that this is still a compiler bug/regression. For this reason, this PR is still marked as “draft” and serves merely as a reference for anyone currently encountering the issue.

### DIFF
--- a/engine/src/inputoutputmap.h
+++ b/engine/src/inputoutputmap.h
@@ -27,13 +27,13 @@
 
 #include "qlcinputprofile.h"
 #include "grandmaster.h"
+#include "../../plugins/interfaces/qlcioplugin.h"
 
 class QXmlStreamReader;
 class QXmlStreamWriter;
 class QElapsedTimer;
 class QLCInputSource;
 class AudioCapture;
-class QLCIOPlugin;
 class OutputPatch;
 class InputPatch;
 class Universe;
--- a/engine/src/inputoutputmap.h
+++ b/engine/src/inputoutputmap.h
@@ -27,7 +27,7 @@
 
 #include "qlcinputprofile.h"
 #include "grandmaster.h"
-#include "../../plugins/interfaces/qlcioplugin.h"
+#include "qlcioplugin.h"
 
 class QXmlStreamReader;
 class QXmlStreamWriter;
--- a/qmlui/app.h
+++ b/qmlui/app.h
@@ -26,6 +26,9 @@
 #include <QObject>
 #include "doc.h"
 
+#include "fixtureeditor/modeedit.h"
+#include "treemodelitem.h"
+
 class MainView2D;
 class ShowManager;
 class SimpleDesk;
--- a/engine/src/bus.h
+++ b/engine/src/bus.h
@@ -25,6 +25,8 @@
 #include <QString>
 #include <QHash>
 
+#include "../../plugins/interfaces/qlcioplugin.h"
+
 class QXmlStreamReader;
 class BusEntry;
 
--- a/engine/test/channelsgroup/channelsgroup_test.cpp
+++ b/engine/test/channelsgroup/channelsgroup_test.cpp
@@ -166,17 +166,17 @@ void ChannelsGroup_Test::saveLoad()
     QXmlStreamReader xmlReader(&buffer);
     xmlReader.readNextStartElement();
 
-    ChannelsGroup grp2(m_doc);
-    QVERIFY(grp2.loadXML(xmlReader) == true);
-    QCOMPARE(grp2.id(), quint32(7));
-    QCOMPARE(grp2.name(), QString("SaveGroup"));
-    QCOMPARE(grp2.getChannels().size(), 1);
-    SceneValue scv = grp2.getChannels().first();
+    ChannelsGroup group2(m_doc);
+    QVERIFY(group2.loadXML(xmlReader) == true);
+    QCOMPARE(group2.id(), quint32(7));
+    QCOMPARE(group2.name(), QString("SaveGroup"));
+    QCOMPARE(group2.getChannels().size(), 1);
+    SceneValue scv = group2.getChannels().first();
     QCOMPARE(scv.fxi, fxi->id());
     QCOMPARE(scv.channel, quint32(0));
-    QVERIFY(!grp2.inputSource().isNull());
-    QCOMPARE(grp2.inputSource()->universe(), quint32(2));
-    QCOMPARE(grp2.inputSource()->channel(), quint32(3));
+    QVERIFY(!group2.inputSource().isNull());
+    QCOMPARE(group2.inputSource()->universe(), quint32(2));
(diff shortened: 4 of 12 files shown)


qmlui/app.h
```diff
@@ -26,6 +26,9 @@
 #include <QObject>
 #include "doc.h"
 
+#include "fixtureeditor/modeedit.h"
```
mcallegari: This is not clear to me.
ModeEdit is not referenced anywhere in the App class.

jpue: I actually let myself be confused by the new compiler warnings and did not investigate the issue further, as the builds were running successfully with my original patch.

I have now done that and placed all the `#include`s in the appropriate places.

The result has been rebased and force-pushed.

engine/src/inputoutputmap.h
```diff
@@ -27,13 +27,13 @@
 
 #include "qlcinputprofile.h"
 #include "grandmaster.h"
+#include "../../plugins/interfaces/qlcioplugin.h"
```
mcallegari: I prefer to add an include path in the cmake file and here just include "qlcioplugin.h" without relative path

jpue: Changed it.

kripton:
Thanks for working on this! In the meantime, I will remove `-Wall -Wextra -Werror" from the variables file so the compilation should no longer fail, even if it will print the warning with GCC 16.

mcallegari:
Right now on Arch the build fails with this error

```
[  105s] [  4%] Building CXX object engine/src/CMakeFiles/qlcplusengine.dir/qlcplusengine_autogen/mocs_compilation.cpp.o
[  109s] In file included from /usr/src/packages/BUILD/qlcplus5-git/src/qlcplus5-git-5.2.2+git.20260503.95b340a/build/engine/src/qlcplusengine_autogen/Z5NZOC3L2Z/moc_qlcioplugin.cpp:9,
[  109s]                  from /usr/src/packages/BUILD/qlcplus5-git/src/qlcplus5-git-5.2.2+git.20260503.95b340a/build/engine/src/qlcplusengine_autogen/mocs_compilation.cpp:42:
[  109s] /usr/src/packages/BUILD/qlcplus5-git/src/qlcplus5-git-5.2.2+git.20260503.95b340a/build/engine/src/qlcplusengine_autogen/Z5NZOC3L2Z/../../../../../plugins/interfaces/qlcioplugin.h:90:7: error: defining ‘QLCIOPlugin’, which previously failed to be complete in a SFINAE context [-Werror=sfinae-incomplete=]
[  109s]    90 | class QLCIOPlugin : public QObject
[  109s]       |       ^~~~~~~~~~~
[  109s] In file included from /usr/include/qt6/QtCore/qobject.h:18,
[  109s]                  from /usr/include/qt6/QtCore/QObject:1,
[  109s]                  from /usr/src/packages/BUILD/qlcplus5-git/src/qlcplus5-git-5.2.2+git.20260503.95b340a/build/engine/src/qlcplusengine_autogen/EWIEGA46WW/../../../../../engine/src/bus.h:24,
[  109s]                  from /usr/src/packages/BUILD/qlcplus5-git/src/qlcplus5-git-5.2.2+git.20260503.95b340a/build/engine/src/qlcplusengine_autogen/EWIEGA46WW/moc_bus.cpp:9,
[  109s]                  from /usr/src/packages/BUILD/qlcplus5-git/src/qlcplus5-git-5.2.2+git.20260503.95b340a/build/engine/src/qlcplusengine_autogen/mocs_compilation.cpp:2:
[  109s] /usr/include/qt6/QtCore/qmetatype.h:343:64: note: here.  Use ‘-Wsfinae-incomplete=2’ for a diagnostic at that point
[  109s]   343 |         static auto check(U *) -> std::integral_constant<bool, sizeof(U) != 0>;
[  109s]       |                                                                ^~~~~~~~~
[  111s] cc1plus: all warnings being treated as errors
[  111s] make[2]: *** [engine/src/CMakeFiles/qlcplusengine.dir/build.make:86: engine/src/CMakeFiles/qlcplusengine.dir/qlcplusengine_autogen/mocs_compilation.cpp.o] Error 1
[  111s] make[1]: *** [CMakeFiles/Makefile2:3501: engine/src/CMakeFiles/qlcplusengine.dir/all] Error 2
[  111s] make: *** [Makefile:156: all] Error 2
```

Can it be related to GCC 16?

jpue:
Yes, this is exactly the error this PR solves.

jpue:
> Thanks for working on this! In the meantime, I will remove `-Wall -Wextra -Werror" from the variables file so the compilation should no longer fail, even if it will print the warning with GCC 16.

Actually, it should also build without warnings with GCC 16 (even with `-Wall -Wextra -Werror` enabled), as long as you do not run into #1950.

If there remains an issue, please let me know.

mcallegari:
> Yes, this is exactly the error this PR solves.

Good! Any reason this is still a draft?
Not ready to be merged yet?

mcallegari:
Thanks :+1:

Issue #1595, opened 2024-07-08
Cuelist with Submaster doesn't work until QLC+ is reloaded


CarnivalBen:
**Describe the bug**
If you set up the virtual console with a cuelist and a submaster inside a solo frame, the submaster does not work until the workspace is saved, QLC+ is restarted and the workspace re-loaded.

**To Reproduce**
1. Open QLC+
2. Add a generic dimmer fixture.
3. Add a scene that sets the test fixture to 100%.
4. Add a chaser that contains the new scene with hold/duration set to infinite.
5. Switch to Virtual Console.
6. Add a Solo Frame.
7. Add a Cuelist inside the solo frame, in the properties set the cuelist chaser to the one created in stop 4, and set the slide fader to crossfade.
8. Add a slider inside the same solo frame, in the properties switch to submaster mode.
9. Open the fixture monitor.
10. Switch to operate mode.
11. Click Play on the cuelist, note the fixture monitor shows the fixture at max brightness.
12. Adjust the submaster slider - The fixture brightness does not change as it should.
13. Save the workspace.
14. Close, reopen QLC+ and the saved workspace.
15. Perfom steps 9 to 12 again... this time the fixture brightness is affected by the submaster, as you'd expect.

**Expected behavior**
I would expect the submaster to work without having to reload the workspace.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - QLC+ Version: 4.13.0
- hardware setup: high spec laptop - doesn't make a difference to the issue as I've tried it on other hardware with the same result.

CarnivalBen:
Also tested it with QLC+ v4.13.1 - Issue remains.

mcallegari:
Not reproduced.
Submaster worked immediately.

yestalgia:
I was able to reproduce your bug based on your video. Thank you for taking the time. 

I completed the test using Windows 11, 4.13.1 GIT.

cmuellner:
I tested this, and it is still reproducible on current master.

Steps are:
- Create a 1-channel fixture
- Create Scene 0 with fixture at 255
- Create a chaser with this scene and infinite duration
- In virtual console, create solo frame
- Add cue list to solo frame and connect cue list to chaser
- Set cue list side fader to crossfade mode
- Add a slider and set to submaster
- Switch to operate mode
- Start cue list (dimmer jumps to 255)
- Use slider to adjust the dimmer (works)
- Stop operate mode
- Change mode of slider to playback mode
[...]

PR #2009, opened 2026-05-07
chaserrunner: apply submaster updates to all crossfade steps

### LINKED ISSUE #2004: Submaster does not update all fixtures when crossfade is running
During the investigation of #1595, I came across an issue with a submaster when crossfade is running.

This can be reproduced with these steps:

  1. Create two 1-channel fixtures.
  2. Create Scene 0 with Fixture A at 255.
  3. Create Scene 1 with Fixture B at 255.
  4. Create a Chaser with exactly those two scenes as steps.
  5. Set the chaser so it does not auto-advance immediately (infinite duration).
  6. In the Virtual Console, create a Solo Frame.
  7. Inside that same frame, add a Cue List and assign the chaser to it.
  8. Set the cue list side fader mode to Crossfade.
  9. Inside the same frame, add a Slider and set it to Submaster.
  10. Switch to Operate mode.
  11. Start the cue list.
  12. Move the cue list crossfade slider to an in-between position, for example 25%.
  13. Confirm both fixtures are now contributing output.
  14. Pull the submaster slider to 0.

  Expected result:

  - Both fixtures go to 0.

  Actual result (current git master):

  - Only one fixture goes to 0.
  - The other stays on.

This may be a rare bug, since using the submaster while crossfading is unlikely.

cmuellner:
Track the master and step intensities of each running step and reapply master-intensity changes to every active crossfade step.

Add an engine regression in chaserrunner_test so this bug can be proposed independently from #1595.

Fixes: #2004.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

New testcase in engine/test/chaserrunner/chaserrunner_test.cpp.

### DIFF
--- a/engine/src/chaserrunner.cpp
+++ b/engine/src/chaserrunner.cpp
@@ -423,24 +423,38 @@ void ChaserRunner::adjustStepIntensity(qreal fraction, int requestedStepIndex, i
     int stepIndex = requestedStepIndex;
     if (stepIndex == -1)
     {
-        stepIndex = m_lastRunStepIdx;
         // store the intensity to be applied at the next step startup
         m_pendingAction.m_masterIntensity = fraction;
-    }
 
-    foreach (ChaserRunnerStep *step, m_runnerSteps)
-    {
-        if (stepIndex == step->m_index && step->m_function != NULL)
+        foreach (ChaserRunnerStep *step, m_runnerSteps)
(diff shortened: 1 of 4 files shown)


mcallegari:
Looks good to me
Thanks a lot :+1:

PR #2020, opened 2026-05-14
engine/sequence: fix crash after removing fixture

jpue:
_Reported at https://www.qlcplus.org/forum/viewtopic.php?t=19616_

1. Open QLC+ v4 with an empty workspace.
2. Add a fixture (for example, a `Generic Dimmer`).
3. Create a sequence with a single step containing the fixture.
4. Remove the fixture.
5. Open the sequence (in the editor) and click on the step.
6. QLC+ crashes.

**Expected behaviour**
QLC+ should not crash, and the values associated with the deleted fixture should be removed.

**Problem Analysis**
The observed crash is caused by the `Q_ASSERT` in `SceneEditor::slotSetSceneValues`.
The issue could be fix (or _masked_) by replacing this assertion with an if-clause (`if (fixture == NULL) { continue; }`), which skips the fixture if it does not exist.

However, this would not fix the underlying problem.
The real problem is that, unlike _scenes_, sequences do not remove fixture values from their steps if the corresponding fixture is deleted.

**Proposed Solutions**
In the `Sequence` class, override the `slotFixtureRemoved` function inherited from `Function` with a Sequence-specific one, which cleans up the sequence steps when the fixture is deleted.

I am not sure whether implementing the solution with a range-based for-each loop works in every case, since it iterates over the data structure that it (or more precisely, the called `unset` function) modifies itself. At least during my tests (also with more than one fixture), it seemed to work reliably.


engine/src/sequence.cpp
```diff
@@ -146,6 +146,25 @@ void Sequence::applyDumpValues(const QList<SceneValue> &dumpedValues, int target
...
+{
+    if (fxID == Fixture::invalidId())
+        return;
+
+    for (ChaserStep& step : m_steps)
+    {
+        for (SceneValue& value : step.values)
+        {
+            if (value.fxi == fxID)
+                step.unSetValue(value);
```
cmuellner: You iterate over `step.values` and call `step.unSetValue(v)`, which essentially does `step.values.removeAt(values.indexOf(v));`.

However, `step.values` is a QList and modifying a QList invalidates its iterators, so this loop is relying on undefined behavior.

I would copy the code from `ChannelsGroup::slotFixtureRemoved()` or `Scene::slotFixtureRemoved()`, which uses a `QMutableListIterator` and tracks the removal with `hasChanged`.

engine/src/sequence.cpp
```diff
@@ -146,6 +146,25 @@ void Sequence::applyDumpValues(const QList<SceneValue> &dumpedValues, int target
     }
 }
 
+/*****************************************************************************
+ * Fixtures
+ *****************************************************************************/
+
+void Sequence::slotFixtureRemoved(quint32 fxID)
+{
+    if (fxID == Fixture::invalidId())
```
cmuellner: I would drop this check.
Similar code (e.g. `ChannelsGroup::slotFixtureRemoved()` or `Scene::slotFixtureRemoved()`) does not include the check and if someone does call this function with `Fixture::invalidId()`, the iterator loop will be a no-op because no valid SceneValue should match that fixture id.

jpue:
Thanks for the feedback.
I have adapted the code accordingly and rebased it as well.

mcallegari:
Looks good to me, thanks :+1:

cmuellner:
I don't know the structure here well enough, but I wonder if this still misses change tracking and notification (`hasChanged` and `emit changed(this->id())` on removal).

PR #2021, opened 2026-05-14
Add flash/unFlash support for Collection functions

Paul-Goulpie:
### Summary

This PR implements the `flash()` and `unFlash()` virtual methods on the `Collection` function type, which were previously not overridden. As a result, assigning a Virtual Console button in **Flash** mode to a Collection had no effect. The fix propagates the flash state down to all child functions of the collection.

A label update for the Flash button in the Virtual Console properties UI was also shipped alongside, clarifying that Collections are now supported.

---

### Problem

`Collection` did not override `Function::flash()` / `Function::unFlash()`. When a VC button configured in Flash mode triggered a Collection, the base-class implementation was called but none of the child functions received the flash signal — resulting in no visible DMX output.

---

#### Engine — `engine/src/collection.{h,cpp}`

- Override `flash(MasterTimer*, bool shouldOverride, bool forceLTP)`: iterates over all child function IDs, resolves each `Function*` from the document, and calls `flash()` on it before delegating to `Function::flash()` for bookkeeping.
- Override `unFlash(MasterTimer*)`: same propagation pattern in reverse, guarded by an early-return when the collection is not currently flashing.
- Both methods follow the exact same guard pattern already used by `Scene`, `Chaser`, and `RGBMatrix`.

#### Tests — `engine/test/collection/collection_test.{cpp,h}`

Three new test cases covering:

| Test | What it verifies |
|---|---|
| `flashUnflash` | Flash propagates to all child scenes; unFlash restores all child states and clears DMX sources |
| `flashAlreadyFlashing` | Calling `flash()` a second time while already flashing is a no-op (no duplicate DMX source registration) |
| `unFlashNotFlashing` | Calling `unFlash()` when not flashing is a no-op |

#### i18n — all language files (`ui/src/`, `qmlui/`)

Updated the Flash button label in the VC Button Properties dialog from `"Flash"` to `"Flash (also Collections)"` across **20 translation files** (ca_ES, cz_CZ, de_DE, es_ES, fi_FI, fr_FR, it_IT, ja_JP, nl_NL, pl_PL, pt_BR, ru_RU, uk_UA) in both the Qt Widgets UI and QML UI stacks.

#### UI — `ui/src/virtualconsole/vcbuttonproperties.ui` · `qmlui/qml/virtualconsole/VCButtonProperties.qml`

Updated the static label string for the Flash action option to reflect the newly supported function type.

---

### Testing

- All three new unit tests pass (`collection_test`).
- Manually verified: a VC button in Flash mode bound to a Collection containing two Scenes now correctly flashes both scenes simultaneously and releases them on button release.

---

### Checklist

- [x] Engine logic implemented and guarded against double-flash / double-unflash
- [x] Unit tests added for all edge cases
- [x] i18n strings updated in all supported languages (both UI stacks)
- [x] No regression on existing `Collection_Test` test cases

### DIFF
--- a/engine/src/collection.cpp
+++ b/engine/src/collection.cpp
@@ -467,3 +467,39 @@ void Collection::setBlendMode(Universe::BlendMode mode)
 
     Function::setBlendMode(mode);
 }
+
+/****************************************************************************
+ * Flashing
+ ****************************************************************************/
+
+void Collection::flash(MasterTimer *timer, bool shouldOverride, bool forceLTP)
+{
+    if (flashing() == true)
+        return;
+
+    for (int i = 0; i < m_functions.count(); i++)
+    {
+        Function* function = doc()->function(m_functions.at(i));
+        Q_ASSERT(function != NULL);
+        function->flash(timer, shouldOverride, forceLTP);
+    }
+
+    Q_ASSERT(timer != NULL);
+    Function::flash(timer, shouldOverride, forceLTP);
+}
+
+void Collection::unFlash(MasterTimer *timer)
+{
+    if (flashing() == false)
+        return;
+
+    for (int i = 0; i < m_functions.count(); i++)
+    {
+        Function* function = doc()->function(m_functions.at(i));
+        Q_ASSERT(function != NULL);
+        function->unFlash(timer);
+    }
+
+    Q_ASSERT(timer != NULL);
+    Function::unFlash(timer);
+}
--- a/engine/src/collection.h
+++ b/engine/src/collection.h
@@ -179,6 +179,16 @@ public:
 public:
     /** @reimp */
     void setBlendMode(Universe::BlendMode mode) override;
+
+    /*********************************************************************
+     * Flash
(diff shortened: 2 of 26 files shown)


mcallegari:
Rejected.
Flashing of any Function can be done with a VC slider + flash button.
Next time, perhaps, before vibe coding 200 lines of code, let's discuss this somewhere first.
Thank you.

Paul-Goulpie:
> Rejected. Flashing of any Function can be done with a VC slider + flash button. Next time, perhaps, before vibe coding 200 lines of code, let's discuss this somewhere first. Thank you.

Hi @mcallegari ,

Thank you for taking the time to review this.
I understand the rejection — the VCSlider + flash button already covers this use case at the intensity level, making a dedicated Collection::flash() override unnecessary.
I also take note of your point about the process. I should have opened a discussion first to validate the approach before writing any code. I'll make sure to do that for any future contribution.

Sorry for the noise.

PR #2022, opened 2026-05-15
gitignore: ignore CMake and Qt generated files

cmuellner:
An in-tree CMake build leaves generated metadata, Qt autogen directories, and install staging files as untracked changes.

Ignore those generated paths so build output does not obscure real source changes in git status. Narrow build* to build*/ so tracked files such as .github/workflows/build.yml are not hidden by ignore rules.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,8 +17,10 @@
 
 .qmake.stash
 .qtc_clangd
+.qt/
 *.pro.user
 *.pro.user.*
+*.qrc.depends
 moc_*.cpp
 qrc_*.cpp
 Makefile
@@ -26,6 +28,7 @@ Makefile.*
 *-build-*
 ui_*
 object_*
+*_autogen/
 engine/src/qlcconfig.h
 *.swp
 *.qmlc
@@ -53,3 +56,13 @@ coverage/
 # CMake build
 CMakeLists.txt.user
 build*
+.ninja_deps
+.ninja_log
+CMakeCache.txt
+CMakeFiles/
+CPackConfig.cmake
+CPackSourceConfig.cmake
+cmake_install.cmake
+cmake_uninstall.cmake
+install_manifest.txt
+/install/


mcallegari:
I'm not sure this one is OK.
Can you please tell me how you build QLC+?
I never had the need to add those files to gitignore.

Also, this statement:

> Narrow build* to build*/ so tracked files such as .github/workflows/build.yml are not hidden by ignore rules.

is not true: I've changed build.yml many times and it's not excluded by gitignore.

cmuellner:
I'm building with:

```
WARNING_FLAGS="-Wall -Wextra -Werror"
CFLAGS="$WARNING_FLAGS"
CXXFLAGS="$WARNING_FLAGS"

cmake   -B . \
        -G Ninja \
        -Dqmlui=OFF \
        -DBUILD_SHARED_LIBS:BOOL=OFF \
        -DCMAKE_BUILD_TYPE=Release \
        -DCMAKE_PREFIX_PATH=/usr/lib64/cmake/Qt6 \
        -DCMAKE_C_FLAGS="$CFLAGS" \
        -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
        .

cmake --build . -j$(nproc)

DESTDIR="$PWD/install" cmake --install .
```

cmuellner:
You are right, thanks for catching the issue with `build*/`. I forgot one important detail of how gitignore rules are applied: since `.github/workflows/build.yml` is already tracked, it is not hidden by the existing `build*` rule. The `build*` -> `build*/` change was therefore wrong, and I’ll revert it.

For the other entries: for in-tree builds with Unix Makefiles or Ninja, I just double-checked that these additional rules match generated CMake/Qt files. I also noticed that I missed two generated Ninja files, `.ninja_deps` and `.ninja_log`, so I’ll add those as well.

So I’ll update the PR to add only those generated CMake/Qt/Ninja files and leave the existing `build*` behavior untouched.

If you prefer not to support generated in-tree build/install artifacts in `.gitignore`, I’m also fine closing the PR.

mcallegari:
Yeah, to be honest I prefer out of tree builds as per wiki instructions and actions workflows.
Also Qt creator does create folders to build. It's easier to cleanup as well.

PR #2030, opened 2026-05-22
replace tab stops with spaces and remove trailing whitespace characters

jpue:
This PR tries to standardise the appearance of code across different editors, development environments and/or platforms by enforcing an indentation of four spaces instead of (one) tab stop(s).
Moreover, this PR cleans up the code a little more by removing trailing whitespace characters.

This PR deliberately leaves some files untouched to avoid unwanted changes (learning from the mistakes I made in the “`const` PR” some weeks ago):
- Markdown files (since they use trailing spaces to separate paragraphs)
- most files under `debian` and `resources/schemas` as well as many files under `platforms` and `resources` (I am not familiar with the syntax/formatting rules for these special files)
- some translation files and some old Japanese documentation (`resources/docs/html_ja_JP`) (so as not to break encoding stuff)
- the C++ code in `plugins/hid` and `plugins/peperoni`, whose indentation is mostly based on tabs
(Replacing all of them would have cluttered this PR and made it very confusing to review.)


mcallegari:
Thanks :+1:

PR #2035, opened 2026-05-25
Add qtbase translations

schorschii:
## Description

**Summary of Changes:**
Added loading of qtbase translation files in order to translate e.g. dialog standard buttons like Yes, No and Cancel, which are only displayed in English otherwise:
Bildschirmfoto vom 2026-05-25 20-13-36

## Checklist

  - [x] Linux
  - [x] Windows
  - [ ] macOS

## Testing

**Test Cases:**

**Test Results:**

## Additional Notes

### DIFF
--- a/create-exe.sh
+++ b/create-exe.sh
@@ -48,7 +48,7 @@ else
 fi
 
 # remove uneeded stuff
-rm -rf generic networkinformation qmltooling renderplugins tls translations
+rm -rf generic networkinformation qmltooling renderplugins tls
 
 
 # Create Installer
--- a/platforms/windows/qlcplus4Qt5.nsi
+++ b/platforms/windows/qlcplus4Qt5.nsi
@@ -98,6 +98,7 @@ Section
 	File /r ModifiersTemplates
 	File /r Plugins
 	File /r RGBScripts
+	File /r translations
 	File /r Web
 
 	WriteRegStr HKCR ".qxw" "" "QLightControllerPlus.Document"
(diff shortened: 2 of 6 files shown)


jpue:
I have actually been looking into this on my [`qt_translation`](/jpue/qlcplus/tree/qt_translation) branch for several months now.
For comparison, I just rebased the branch and pushed the latest version of my changes, which also implement this for QLC+ 4.

There are two reasons why I have not opened a PR for it yet:
1. I am not sure how the language files provided by Qt should be distributed via the installer. It works in my development environment, but would the paths be correct on a production machine as well?
Furthermore, I also do not know if it is even technically legal to include these Qt files in our own software.
2. My solution for loading the `qtbase` translation for QLC+ 5 (`qmlui`) is quite similar to yours. However, QLC+ 5 allows for dynamic language switching/re-translations without restarting the program, which (in my tests) has never worked with such solutions (see also [this](https://forum.qt.io/topic/163786/dialog-standardbuttons-localization) thread).
Were you able to accomplish that?

schorschii:
> I am not sure how the language files provided by Qt should be distributed via the installer. It works in my development environment, but would the paths be correct on a production machine as well?

On Linux it works since the qtbase files are installed in the system. I now also tested Windows and this seems to work too. After executing the windeployqt command, a directory "translations" is created containing the qt_XX.qm files. The only difference: they are called "qt_XX" and not "qtbase_XX". Since "qt_XX" works on Linux too, I just changed it so it works for both OS.

Edit: added the translation dir into the Windows installer.

> Furthermore, I also do not know if it is even technically legal to include these Qt files in our own software.

Why not? I mean they share the same license as the Qt libs itself.

> However, QLC+ 5 allows for dynamic language switching/re-translations without restarting the program, which (in my tests) has never worked with such solutions (see also [this](https://forum.qt.io/topic/163786/dialog-standardbuttons-localization) thread). Were you able to accomplish that?

Unfortunately, no. But I think it's way better than not translated at all.

mcallegari:
Thanks 👍

PR #2046, opened 2026-06-14
main: add qtbase translations

jpue:
_follow-up to #2035_

As mentioned in #2035, I had already implemented the changes of that PR for QLC+ 4.

Since that PR already changed the v4 installer accordingly, this PR only needs to add the code to load the Qt translation files.

I also added the case distinction between operating systems (macOS vs. non-macOS) from 87a7cdedc00b01cf9f882176d1194d38229bcc43, but I am unable to test this due to the lack of such hardware.

### DIFF
--- a/engine/src/qlci18n.cpp
+++ b/engine/src/qlci18n.cpp
@@ -23,6 +23,7 @@
 #include <QString>
 #include <QDebug>
 #include <QDir>
+#include <QLibraryInfo>
 
 #include "qlcconfig.h"
 #include "qlcfile.h"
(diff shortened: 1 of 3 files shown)


jpue:
> Since that PR already changed the v4 installer accordingly, this PR only needs to add the code to load the Qt translation files.

On the other hand, the AppImage creation script (`create-appimage.sh`) has not been updated yet.

Therefore, the required Qt translation files are not included in those versions of QLC+ 5, for example not in the official AppImage binary of the fresh QLC+ 5.2.2 release.

mcallegari:
Thanks! :+1:

Issue #1881, opened 2025-11-24
[Feature] full beat tracking


mcallegari:
Thanks to @Wazzledi we now have a full blown beat tracker. (merged with #1876)
However, it needs to be fine tuned to work properly.

For example, using this video
https://www.youtube.com/watch?v=6oz0ivczNSY
I get BPM detected: 180 and beats are emitted at a random rate.

Perhaps something not right during the integration.

mcallegari:
@Wazzledi let's use this as an open discussion to fine tune the tracker. Thanks

Wazzledi:
> Perhaps something not right during the integration.

I looked into it and the integration looks fine. With my test tracks, it works as intended. However, it might be good, if the provided BPM is also sent to the engine. This would make the BPM label consistent instead of jumping around. 

> I get BPM detected: 180 and beats are emitted at a random rate.

Yeah, when I tested it with an online metronome, I also did not get good results. I did not have time yet to look into this issue. Maybe this will be fixed, when I implement the silence gate.

mcallegari:
OK, thanks, in the meantime I'll fix the build with Qt5 on OBS.

mcallegari:
@Wazzledi I asked chatgpt to review the BeatTracking class.
Attached the outcome.
I'm wondering how it can actually work for you, considering all the major issues found in the code.
Some of them are plain wrong formulas.

I'll apply the fixes later today and see if the tracking improves.

[BeatTracking_Code_Review_CodeBlocks.pdf](https://github.com/user-attachments/files/23743902/BeatTracking_Code_Review_CodeBlocks.pdf)

Wazzledi:
@mcallegari Yeah, die not expect to have it perfectly implemented in the first iteration. That's why I asked you to test it and give feedback.

I looked through the result. Yes there are some dump mistakes. Let me go through it quickly.

1. Memcpy. Yes this is a mistake and must be fixed. Right now too few values are copied. Funny enough that chatgpt gave me this code line :D 
2. The last magnitudes where stored previously but I mistakenly removed it by cleaning the code. Yes it must be stored persistently for the next frame. Right now the onset envelope is just depending on the overall magnitude of the frame.
3. Mistake occured during cleaning. Must be changed. 
4. Yeah, could be changed. But the size of the values is fixed so the bounds will never be exceeded. However, for security reasons it would be good to implement this.
5. Same as 4. But the division by zero bound must be implemented.

Performance:
[...]

PR #2062, opened 2026-07-06
engine: fix beat tracker integration and add comb-ACF tempo tracker (issue #1881)

varghele:
## Description

**Summary of Changes:**

As discussed in #1881, this fixes the beat tracker integration defects behind the reported symptoms and adds a tempo-induction beat tracker (`beattrackerauto.cpp`) as the active implementation.

*Integration fixes:*
- `BeatTracking` now takes `(sampleRate, channels)` and converts total samples to frames in its mixdown — this removes an out-of-bounds read and a 2x time-base distortion (the previous hard-coded stereo mixdown on mono capture is what doubled the BPM and produced erratic beats; the issue's test video is a 100 BPM metronome and read ~198 with the bug reproduced, 99.6 correctly wired).
- New `AudioCapture::beatBpmChanged(int)` signal: the UI now shows the tracker's own tempo estimate instead of re-deriving BPM from the wall-clock spacing of `beatDetected()` signals, which amplified every hop of emission jitter into a BPM change.
- Bugfix found along the way: `beattracker.h` reused `BEATTRACKING_H` as its include guard, colliding with `beattracking.h` — a translation unit including both silently dropped one of them.

*New tracker (in one paragraph):* an onset front end computes, per 512-sample hop (~86/s), the positive log-magnitude spectral difference over a ~93 ms window, split into three bands (<200 Hz / 200-4000 Hz / >=4 kHz) with per-band peak saturation and rising-edge detection, so kicks and snares contribute near-equally (no backbeat half-lock) while hats stay weaker (no subdivision double-tempo). Every 2 s an 8 s onset window is autocorrelated and every candidate on a fractional 50-240 BPM grid is scored by a harmonic comb (ACF at 1-4x the period); a temporal belief filter picks a stable base, an octave-raise step resolves the half-tempo trap, and the reported BPM is the median of the last 3 analyses, gated on confidence. Beats are emitted from the predicted tempo+phase grid ("steady beats"), not per audible onset. Full documentation, tuning rationale and validation record: `engine/audio/src/beattrackerauto.md`.

It is self-contained plain C++ (own radix-2 FFT): no Qt or fftw3 dependency, so beat tracking also works in builds without fftw3. CPU: ~1.3% of one core.

*Behavior notes (documented):* first estimate after ~4 s of music; re-lock after a live tempo change takes ~12-16 s (deliberate stability trade-off); material fully dominated by 8th-note energy can read an octave up.

*Kept for the consolidation:* the previous spectral-flux tracker stays selectable behind the `NEW_TRACKER` define for A/B comparison (see the benchmark below) — happy to remove it in a follow-up commit if a hard replacement is preferred.

**Related Issues:**
Related to #1881.

## Checklist

  - [x] Linux (Ubuntu 24.04, GCC 13.3, Qt 6.4)
  - [x] Windows (MinGW 13.1, Qt 6.8)
  - [ ] macOS (untested — the tracker is plain C++ with no platform-specific code; CI should confirm)
- [x] Documentation: no new user-facing options; the algorithm is documented in-repo (`engine/audio/src/beattrackerauto.md`).

## Testing

**Test Cases:**

- **Unit test** (`engine/test/beattrackerauto`, runs in `make check`): synthetic kick tracks through `processAudio()` in capture-sized blocks — tempo within 2% at 90/120/140/174 BPM, regular post-warmup beat spacing, no estimate/beats on silence, interleaved stereo matches mono. 9/9 pass on both platforms.
- **Three-way benchmark tool** (`engine/test/beattrackerbench`): synthesizes 8 percussion scenarios (four-on-floor, 8th hats, backbeat, 8th-note bassline, swung hats, ±3% drift, 2.5 s dropout, +30% tempo step) x 8 tempi with known ground truth and runs all three implementations head-to-head. Quick mode runs inside `make check`; `--full` prints the complete table. The benchmark is deterministic (fixed seed) — it reproduces this exact table on Windows and Linux:

  | tracker | correct (4% tolerance) |
  |---|---|
  | BeatTrackerAuto (default) | **63/64** |
  | BeatTracker (spectral flux) | 54/64 |
  | BeatTracking (ACF) | 48/64 |

  The new tracker's single miss is a 90 BPM 8th-hats clip reading an octave up — the documented known limit (the ACF tracker flips on the same clip).
- **The issue's own test material**: the #1881 video (a 100 BPM metronome) reads 100.00 BPM with 0.600 s beat intervals; a 140 BPM rock drum-track video reads 139.75. Confirmed at file level and in live microphone tests on Windows (phone speaker -> laptop mic: locks at 100 within seconds, holds steady) and Linux (virtual source).
- **Full engine test suite**: complete run on Linux (95 suites, 3416 assertions, 0 failures, fixture validation OK); on Windows all suites pass except pre-existing path-normalization/name-casing cases that fail identically on the untouched branch.

**Testing note:** while verifying on Linux we also noticed that a test suite aborting (e.g. a non-executable `test.sh`) is silently swallowed by `unittest.sh` — `make check` still reports success while skipping the remaining suites. Pre-existing and unrelated to this PR; happy to file it separately.

### DIFF
--- a/engine/audio/src/CMakeLists.txt
+++ b/engine/audio/src/CMakeLists.txt
@@ -9,6 +9,7 @@ add_library(${module_name}
     audioplugincache.cpp audioplugincache.h
     audiorenderer.cpp audiorenderer.h
     beattracker.cpp beattracker.h
+    beattrackerauto.cpp beattrackerauto.h
     beattracking.cpp beattracking.h
 )
 set_property(TARGET ${module_name} PROPERTY POSITION_INDEPENDENT_CODE ON)
--- a/engine/audio/src/audiocapture.cpp
+++ b/engine/audio/src/audiocapture.cpp
@@ -24,7 +24,9 @@
 #include <qmath.h>
 
 #include "audiocapture.h"
-#ifdef NEW_TRACKER
+#if defined(AUTO_TRACKER)
+  #include "beattrackerauto.h"
+#elif defined(NEW_TRACKER)
   #include "beattracker.h"
 #else
   #include "beattracking.h"
@@ -78,13 +80,16 @@ AudioCapture::AudioCapture (QObject* parent)
     m_plan_forward = fftw_plan_dft_r2c_1d(m_bufferSize, m_fftInputBuffer,
                                           reinterpret_cast<fftw_complex*>(m_fftOutputBuffer), 0);
 #endif
- #ifdef NEW_TRACKER
+ #if defined(AUTO_TRACKER)
+    m_beatTracker = new BeatTrackerAuto(m_sampleRate, m_channels);
+    m_lastTrackerBpm = 0;
+ #elif defined(NEW_TRACKER)
     m_beatTracker = new BeatTracker(m_sampleRate, m_bufferSize, m_channels, 86, 1.3);
     m_beatTracker->setBand(40.0, 400.0);      // bit wider band for now
     m_beatTracker->setFluxSmoothing(0.6);     // less smoothing
     m_beatTracker->setMinBeatInterval(0.20);  // ~300 BPM max
  #else
-    m_beatTracker = new BeatTracking(2);
+    m_beatTracker = new BeatTracking(m_sampleRate, m_channels);
  #endif
 }
 
@@ -93,6 +98,7 @@ AudioCapture::~AudioCapture()
     // stop() has to be called from the implementation class
     Q_ASSERT(!this->isRunning());
 
+    delete m_beatTracker;
     delete[] m_audioBuffer;
     delete[] m_audioMixdown;
     delete[] m_fftInputBuffer;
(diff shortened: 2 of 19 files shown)


mcallegari:
@varghele I just tested your tracker and it seems to work pretty well, so thanks a lot for it!
I have two comments:
- as you anticipated re-locking on BPM change is very slow. This version looks already good but I'm wondering if for a next version there might be any improvement (tradeoff) that can be done
- I would merge the beatDetected and beatBpmChanged signals. I would keep only the first with an argument where 0 means "no BPM detected". In this way the UI knows if BPM is already detected or if it has to do the delta calculation.

I'm thinking also when no audio is processed, the UI should keep emitting the latest BPM detected, otherwise all functions will freeze (might or might not be wanted)

varghele:
Thanks for testing it @mcallegari! In ca63de21a the merge fix you proposed is implemented. `beatDetected(int bpm)` is now the only  signal. bpm carries the active tracker's own estimate (all three implementations  provide one), 0 means "no estimate" and `slotProcessBeat()` falls back to the wall-clock delta. I'll have a think about a quicker BPM recognition and the latest BPM. I think the tracker should be honest if it "hears" silence or not, but maybe that can be transported alongside a last detected BPM.

mcallegari:
I have merged this one in the dedicated branch. Thanks a lot again!
I think I'm gonna do some class renaming before merging into master as the default tracker class must become "BeatTracker"

@Wazzledi are you OK if we work together on @varghele implementation?
Right now we have 3 trackers and I need to do some cleanups

Wazzledi:
@mcallegari Yes, of course. Cleanup the implementations as you like.

varghele:
Thanks to all as well, I'll think about some of this again at some point :)

PR #2065, opened 2026-07-12
color tool (full tab) fixes

edogawa23:
1. fix the scale issue (scaling not applied to mouse click/drag coordinates) in rgb colorfield

2. make W, A and UV sliders visible when available, and not interfering with each other when starting to drag from initial 0 value

(solved with help from Devin Desktop)

### DIFF
--- a/qmlui/qml/ColorTool.qml
+++ b/qmlui/qml/ColorTool.qml
@@ -198,6 +198,8 @@ Rectangle
                     item.currentRGB = Qt.binding(function() { return colorToolBox.currentRGB })
                 if (item.hasOwnProperty("currentWAUV"))
                     item.currentWAUV = Qt.binding(function() { return colorToolBox.currentWAUV })
+                if (item.hasOwnProperty("isPaletteEditing"))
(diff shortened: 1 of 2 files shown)


mcallegari:
Hi, thanks but purely looking at the code I'm not sure this will work when a palette has W/A/UV values.
Can you please test that case?

edogawa23:
Well I saved two color palettes and the palette editor shows the correct color in the bottom field. I can change the RGB part but there are no WAUV faders... if I go to black there it shows what I stored for WAUV. 

This would be so great if it could be extended to more led colors, what do you think?
The bottom preview works quite nice already, but should go towards brighter/white the more channels add up, currently amber and uv tend to darken the mix...

Btw there's still something not quite right with the scaling, It seemed ok this afternoon but now it stretches a bit outside the field again... I'll have a look

edogawa23:
W A and UV faders show now again in Palette editor.
I've tested with fixtures containing all three or only a subset, and it seems to be ok now. in ColorTool popup we see only those supported by the fixture , in Palette Editor all three are always shown so that they can be altered.

mcallegari:
Thanks 👍

PR #2066, opened 2026-07-12
fix yaml syntax in obs workflow file

edogawa23:
duplicate keys aren't possible in yaml, thus qlcplus-git trigger got overwritten by the later entry...


mcallegari:
Oh wow, thanks! :+1: 
How come I didn't realize this before.

PR #2072, opened 2026-07-16
another go at fixing OBS workflow, this time hopefully correct

edogawa23:
I'm sorry, but the previous attempt to fix this still didn't work.
We have to test this live, and I only now found time to go ahead with this.
I'm confident it will work this time, though


mcallegari:
Thanks and no problem.
I know it's tricky with online triggers!

PR #2075, opened 2026-07-17
translations: update German translation for QLC+ 5.3.0

jpue:
Added German translations for the new strings introduced since QLC+ 5.2.1.

For most strings, I simply checked the existing (machine?) translation (from 187b779fdc942644f7fefecec547229a1046afdc) and “finished” it.

@Loewe111 @kripton @FloEdelmann @ClarkLiam

### DIFF
--- a/qmlui/qlcplus_de_DE.ts
+++ b/qmlui/qlcplus_de_DE.ts
@@ -3246,7 +3246,7 @@ Ungespeicherte Änderungen gehen verloren.</translation>
     <message>
         <location filename="qml/popup/PopupCreatePalette.qml" line="49"/>
         <source>Position 3D</source>
-        <translation>Position 3D</translation>
+        <translation>3D-Position</translation>
     </message>
     <message>
         <location filename="qml/popup/PopupCreatePalette.qml" line="50"/>
@@ -4597,7 +4597,7 @@ Bitte einen anderen Namen angeben.</translation>
     <message>
         <location filename="qml/fixturesfunctions/SequenceEditor.qml" line="147"/>
         <source>Preview the channel changes on the output</source>
-        <translation>Vorschau der Kanaländerungen an der Ausgabe</translation>
+        <translation>Vorschau der Kanaländerungen in der Ausgabe</translation>
     </message>
     <message>
         <location filename="qml/fixturesfunctions/SequenceEditor.qml" line="157"/>
@@ -6839,7 +6839,7 @@ fallen lassen um die Liste zu füllen</translation>
         <location filename="qml/virtualconsole/VCSpeedDialProperties.qml" line="210"/>
         <source>Control Properties</source>
         <oldsource>Dial Properties</oldsource>
-        <translation>Drehrad-Eigenschaften</translation>
+        <translation>Eigenschaften der Steuerelemente</translation>
     </message>
     <message>
         <location filename="qml/virtualconsole/VCSpeedDialProperties.qml" line="232"/>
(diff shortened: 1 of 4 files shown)


mcallegari:
Thanks, and yes, this is kind of an experiment. I asked Claude to fill in translations.
I checked the Italian one and it was 90% accurate so I'd say to cope with disappeared translators this could be a viable way at least to have some translations rather than a UI in two languages.

Loewe111 (approved):

qmlui/qlcplus_de_DE.ts
```diff
@@ -232,65 +232,65 @@ Ungesicherte Änderungen gehen verloren, wenn du sie nicht speicherst.</translat
...
         <location filename="fixtureeditor/aliasedit.cpp" line="220"/>
         <source>Source and target channels are the same</source>
-        <translation type="unfinished">Quell- und Zielkanal sind identisch</translation>
+        <translation>Quell- und Zielkanal sind identisch</translation>
     </message>
     <message>
         <location filename="fixtureeditor/aliasedit.cpp" line="224"/>
         <source>Mode &apos;%1&apos; does not exist</source>
-        <translation type="unfinished">Modus &apos;%1&apos; existiert nicht</translation>
+        <translation>Modus &apos;%1&apos; existiert nicht</translation>
```
FloEdelmann: `&apos;` (`'`) is an apostrophe character, which should not really be used as a quotation mark. In German, either use `&bdquo;foo&ldquo;` (`„foo“`) or `&sbquo;foo&lsquo;` (`‚foo‘`), but single quotes should only be used inside double quotes.

See https://de.wikipedia.org/wiki/Anf%C3%BChrungszeichen.

qmlui/qlcplus_de_DE.ts
```diff
@@ -3246,7 +3246,7 @@ Ungespeicherte Änderungen gehen verloren.</translation>
     <message>
         <location filename="qml/popup/PopupCreatePalette.qml" line="49"/>
         <source>Position 3D</source>
-        <translation type="unfinished">Position 3D</translation>
+        <translation>Position 3D</translation>
```
FloEdelmann: Maybe


```suggestion
        <translation>3D-Position</translation>
```

qmlui/qlcplus_de_DE.ts
```diff
@@ -4592,12 +4592,12 @@ Bitte einen anderen Namen angeben.</translation>
...
         <location filename="qml/fixturesfunctions/SequenceEditor.qml" line="109"/>
         <source>Add a fixture/group</source>
-        <translation type="unfinished">Neues Gerät/Gruppe hinzufügen</translation>
+        <translation>Neues Gerät/Gruppe hinzufügen</translation>
     </message>
     <message>
         <location filename="qml/fixturesfunctions/SequenceEditor.qml" line="147"/>
         <source>Preview the channel changes on the output</source>
-        <translation type="unfinished">Kanaländerungen an der Ausgabe vorschauen</translation>
+        <translation>Vorschau der Kanaländerungen an der Ausgabe</translation>
```
FloEdelmann: Maybe

```suggestion
        <translation>Vorschau der Kanaländerungen in der Ausgabe</translation>
```

(Not sure what "Ausgabe" refers to here.)

qmlui/qlcplus_de_DE.ts
```diff
@@ -6839,12 +6839,12 @@ fallen lassen um die Liste zu füllen</translation>
         <location filename="qml/virtualconsole/VCSpeedDialProperties.qml" line="210"/>
         <source>Control Properties</source>
         <oldsource>Dial Properties</oldsource>
-        <translation type="unfinished">Rad Eigenschaften</translation>
+        <translation>Drehrad-Eigenschaften</translation>
```
FloEdelmann: The source only says "Control", not "Dial", so it should rather be a more generic word than "Drehknopf".

FloEdelmann (changes requested):

FloEdelmann (approved):

mcallegari:
Thanks as always guys! :+1:

PR #2076, opened 2026-07-17
Revised qmlui, ui and webaccess catalan translation

sbenejam:
Revised qmlui, ui and webaccess catalan translation
Revised qmlui, ui and webaccess  spanish translation


mcallegari:
Thanks @sbenejam :+1:

PR #2081, opened 2026-07-21
Improved Audio Beat Tracker

mcallegari:
Improved audio beat tracker contributed by @varghele

An onset front end computes, per 512-sample hop (~86/s), the positive log-magnitude spectral difference over a ~93 ms window, split into three bands (<200 Hz / 200-4000 Hz / >=4 kHz) with per-band peak saturation and rising-edge detection, so kicks and snares contribute near-equally (no backbeat half-lock) while hats stay weaker (no subdivision double-tempo). Every 2 s an 8 s onset window is autocorrelated and every candidate on a fractional 50-240 BPM grid is scored by a harmonic comb (ACF at 1-4x the period); a temporal belief filter picks a stable base, an octave-raise step resolves the half-tempo trap, and the reported BPM is the median of the last 3 analyses, gated on confidence. Beats are emitted from the predicted tempo+phase grid ("steady beats"), not per audible onset.


mcallegari:
@varghele  @Wazzledi I cleanup up the branch and left a single, hopefully final tracker.
If you are both OK with this I would proceed to merge.
Thanks

PR #2089, opened 2026-08-06
Rename qlcplus-qml -> qlcplus5 on all platforms

mcallegari:
Ensure no collisions on Linux. Only shared resources are fixtures, rgb scripts, gobos, etc.


jpue:
I think the changes in this PR (keeping QLC+ 4 and 5 installations more separate) will pay off in the long run.

In the short term, however, this will probably generate a lot of activity and issue reports on the forums.
Especially on Windows, users are going to download the new version and simply install it over their existing installation, since this is how updates are carried out for most software (which do not have a built-in updater).
By doing so, the old `qlcplus-qml.exe` will likely remain in-place, whilst unrenamed `.dll` library files will be updated. If there now exists a link/shortcut to `qlcplus-qml.exe` used to start QLC+, it will either no longer work at all or crash unexpectedly/in weird places since parameters of calls to library functions have been changed in number and/or type.

Therefore, maybe a note should be added to the next release notes advising users to completely uninstall QLC+ and then reinstall it if they experience such problems.

In addition, whilst working on the build and deployment tooling, perhaps these changes (fad1e0f7e7a708a446ef35adb81e1ff9a12a9072) could also be integrated, so that in QLC+ 5, the plugin configuration dialogues and the `webaccess` configuration page can also access the existing translations.

mcallegari:
@jpue yes, I'll clearly state that from 5.2.x to 5.3.0 on Windows you need to uninstall first.

@edogawa23 I am worried about OBS builds. This change will basically split QLC+ into 3 packages on Debian: qlcplus (v4), qlcplus5 (v5) and qlcplus-data (common). We'll have to adjust the rules files, unless we want to keep building as a single package but the v4/v5 clash will stay.

edogawa23:
I'm not familiar with how the Windows installers are being built. I briefly looked at the Actions logs for this PR v4 and v5 Windows builds, but I can't take the time to study more carefully, currently.
So my question may seem silly, but can't we delete the previous installation's files while installing over or updating? AFAIK vlc, Ardour/Mixbus and many other software do it that way...

As for the debian package changes, I'd be more than happy to split common files into a -data subpackage, and would do the same for rpm distros and Arch. It will take a bit of work and time that I'm willing to spend, and would be able to do so in the next few weeks.

Generally I find it less than ideal that both the package from upstream debian and our OBS debian/ubuntu builds are generated from different dsc/rules/control etc. files than those in the QLC+ sources debian/ dir. I think there is an alternative way to build deb in OBS, and get rid of our debian.rules etc. duplicates in the OBS package and use the ones in debian/ directly, and again I am happy to take the time and find out how that's done properly.

edogawa23:
Can we have a look on debian OBS building? It seems that a1e6b7a940c05c8cf0ceac02e719c8c85e36de90 broke all the debian builds for v5, as at build time debtransform replaces our fake debian changelog in the OBS package sources by debian/changelog from upstream, and this now is a symlink to changelog-v4. I tried to patch this and point the symlink to changelog-v5 but dpkg-source cannot handle replacing symlinks, it seems. I tried all kinds of things with help from Devin AI but haven't found a way to work this out so far. Right now my daily quota exceeded but I somehow doubt I'll be able to fix this via AI, nor have the time or inclination to dive that deep into debian packaging, especially with that debtransform wizardry on top. If any of you guys know better, please help, otherwise I'll continue trying with that AI agent.

PR #2093, opened 2026-08-16
qmlui: support concurrent native client sessions

infrafast:
## Summary

This change fixes concurrent QLC+ native TCP clients and makes authorization session-safe.

Native clients are currently keyed by peer IP address. A second client connecting from the same address, including multiple clients on `127.0.0.1`, can replace the first socket. Manual authorization is also ambiguous because requests and decisions use the advertised client name while the GUI has only one reusable popup.

## Changes

- assign a UUID to every accepted native TCP connection;
- store and address clients by session ID rather than IP address or client name;
- retain client name, peer IP, and peer port as display-only metadata;
- broadcast updates independently to every authenticated session;
- remove only the session associated with the disconnected socket;
- reject native actions from unauthenticated sessions;
- enforce the authorized session's access mask for inbound actions;
- make project transfer and authorization decisions target an immutable session ID;
- serialize manual authorization requests through a FIFO queue;
- keep immutable name/peer metadata and a guarded socket reference in each queued request;
- clean up active and queued requests when their socket disconnects;
- show client name, peer endpoint, and session ID in the authorization popup;
- add the runtime-only `--allow-all-native` option for unattended deployments;
- log every automatic authorization and print a prominent startup warning;
- document that native protocol encryption is not a strong security boundary and TCP 9998 should remain firewalled or restricted to trusted networks.

## Compatibility and defaults

- The native wire protocol is unchanged.
- Existing native clients remain compatible.
- Manual authorization remains the default.
- `--allow-all-native` is not persisted in workspaces or settings.

## Validation

- Applied cleanly on current `upstream/master` (`e4c85a5c`).
- Full ARM64 `qlcplus5` target built successfully.
- `git diff --check` passes.

This is opened as a draft to confirm that the session model, authorization queue, and access-mask mapping fit the maintainers' preferred direction before marking it ready for merge.


infrafast:
Unique UUID per TCP session; IP/name are metadata only.
Per-session authentication and access-mask enforcement.
Authenticated broadcasts reach every session.
FIFO manual authorization with immutable identity and guarded socket references.
Popup displays name, peer IP/port, and session ID.
Disconnect cleanup handles active and queued requests safely.
Added runtime-only --allow-all-native, full-access logging, startup warning, and documentation.
Wire protocol remains unchanged.

mcallegari:
Something is wrong with this PR. It says 2910 files modified
I guess you need to rebase/merge your fork/branch first

infrafast:
Yes, and sorry for mistake. A packaging commit was accidentally pushed to the previous PR branch. I closed #2093 and recreated a clean PR directly from the current master. #2094 now contains exactly one commit and eight source/documentation files => https://github.com/mcallegari/qlcplus/pull/2094

Issue #1775, opened 2025-07-03
Custom fixture crashes main program while trying to add an instance of it


fg-uulm:
**Describe the bug**
See attached fixture definiton. When trying to add an instance of this fixture QLC+ crashes instantly. Definition file renamed to .txt for github limitations.

[fgtech-BlinderBowlv2.txt](https://github.com/user-attachments/files/21045521/fgtech-BlinderBowlv2.txt)

**To Reproduce**
0. Install fixture to fixture folder
1. Start QLC+
2. Click green cross to add device
3. Select fixture in list

**Desktop (please complete the following information):**
 - OS: macOS 15.4.1
 - QLC+ Version 4.14.1-1

mcallegari:
4.14.1-1 is an obsolete release.
Please test with the latest release or I'll close this report.

mcallegari:
Also, the provided fixture definition shows 23 errors in the [fixture definition validator](https://www.qlcplus.org/fixture_validator.php)
The issue is most likely the missing mode.
This is definitely not the way to report an issue on GitHub.
You haven't read the [relevant documentation](https://docs.qlcplus.org/v4/fixture-definition-editor)
You haven't searched the forums for similar issues.
You haven't validated your definition first.
You haven't read the [fixture sharing guidelines](https://www.qlcplus.org/forum/viewtopic.php?t=6667).

fg-uulm:
Dear @mcallegari, thank for your answer despite my lacking issue! Your answer also resolved my issue, the missing mode was the answer.

However I report such things mainly in order to also make other peoples life easier, and I am also willing to contribute PRs for code or documentation if this is welcome, so let me reply shortly to your message:

- I actually read the documentation on the definition editor, however fell for the misunderstanding that if a device doesn't have different modes I would not need to add modes at all. This requirement could be added in the docs, or even in the software (display a warning like with the manufacturer name and fixture name etc.). 
- The fixture definition validator does not seem to be mentioned in the docs, at least I wasn't aware of it. This could be added to the docs as well, or one could even add the validation logic to the fixture definition editor itself?
- I didn't read the sharing guidelines, that's correct. I didn't plan on sharing the fixture, so it didn't seem relevant
- I searched github for similar issues, as I reckoned that a full blown crash would be more of a bug than a user problem. Maybe a small hint in the docs that incorrect fixtures can fully crash the program would be good?
[...]

PR #2096, opened 2026-08-18
fixtureeditor: reject definitions without modes

### LINKED ISSUE #2095: [v4.14.4] Fixture Editor saves definitions that it can no longer reopen
## Regression

QLC+ 4.14.4 rejects a fixture definition that QLC+ 4.14.3 accepts when the definition contains no `<Mode>` element.

This was introduced by commit 8fa21c7eb149fcdf233a9b13d112ea3727c3c08c (`engine: handle fixture definitions without modes (fix #1775)`), which adds:

```cpp
if (m_modes.isEmpty())
    return false;
```

The loader-side check prevents the crash in #1775, but the Qt Widgets Fixture Editor still permits a user to create and save a fixture definition with a manufacturer/model but no modes. The resulting `.qxf` then cannot be reopened by the Fixture Editor or resolved when loading a workspace, with no actionable validation message.

## Steps to reproduce

1. In QLC+ 4.14.4 Fixture Editor, create a new definition.
2. Enter a manufacturer and model.
3. Save it without adding a mode.
4. Close the editor and attempt to open the saved `.qxf`, or use the definition in a workspace and reopen that workspace.

## Expected behavior

The editor should prevent saving a definition with no modes and explain that a fixture definition requires at least one mode. The loader must continue rejecting such definitions so #1775 is not reintroduced.

## Proposed fix

Add fixture-definition validation to the v4 Fixture Editor save path:

- reject saving when `m_fixtureDef->modes().isEmpty()`;
- display a clear error explaining that at least one mode is required;
- add a regression test covering the editor/model validation and rejected load behavior.

I confirmed the behavior against the attached-style minimal output generated by the v4 Fixture Editor. A separate complete 43-channel fixture definition with a valid mode does not trigger this condition.

gmint:
## Summary

Prevent the Fixture Editor from saving a fixture definition that has no modes.

## Root cause

Since v4.14.4, `QLCFixtureDef::loadXML()` correctly rejects definitions without modes to prevent the crash fixed by #1775. The Fixture Editor could still create and save exactly that invalid definition, leaving it impossible to reopen.

## Changes

- validate that a definition contains at least one mode before Save or Save As
- show the existing missing-information warning, select the Modes tab by widget reference, and focus the add-mode control
- rename the validation helper to reflect its broader responsibility

This preserves the engine-side load guard while preventing the editor from producing files it cannot reload.

Fixes #2095.

## Validation

- `git diff --check`
- inspected against the v4.14.4 source and the reported no-mode QXF reproduction
- the upstream CI build is awaiting maintainer approval because this PR originates from a fork

The local workspace does not include CMake or Qt development dependencies, so the project's `make check` suite could not be run there.

### DIFF
--- a/fixtureeditor/fixtureeditor.cpp
+++ b/fixtureeditor/fixtureeditor.cpp
@@ -256,7 +256,7 @@ bool QLCFixtureEditor::checkFixtureDefinition()
                              tr("Missing important information"),
                              tr("Missing fixture mode.\n"
                                 "Unable to save fixture."));
-        m_tab->setCurrentIndex(3);
+        m_tab->setCurrentWidget(Modes);
         m_addModeButton->setFocus();
         return false;
     }
--- a/fixtureeditor/fixtureeditor.cpp
+++ b/fixtureeditor/fixtureeditor.cpp
@@ -226,10 +226,10 @@ void QLCFixtureEditor::closeEvent(QCloseEvent *e)
     }
 }
 
-bool QLCFixtureEditor::checkManufacturerModel()
+bool QLCFixtureEditor::checkFixtureDefinition()
 {
-    /* Check that the fixture has a manufacturer and a model for
-       unique identification */
+    /* Check that the fixture has a manufacturer, a model and at least one
+       mode for unique identification and loading */
     if (m_fixtureDef->manufacturer().length() == 0)
     {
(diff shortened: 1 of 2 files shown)


mcallegari:
Superseded by 4db9faa579bc18012d8f0cb2ba9c624450f55dd7

PR #2114, opened 2026-09-01
qmlui: persist 3D View "Show FPS" setting across view switches

mjcmatrix:
### What this fixes

The 3D view's **Rendering -> Show FPS** option is switched off again every time the user leaves the 3D view. Turn it on, go to Virtual Console and come back, and it is off. The other Rendering settings — Quality, Ambient light, Smoke amount — are plain members that nothing resets, so only this one is affected.

### Why it happens

`MainView3D::resetItems()` unconditionally calls `setFrameCountEnabled(false)`, and `resetItems()` runs from `enableContext(false)`, i.e. on every view switch.

On top of that, the enabled state was derived from `m_frameAction != nullptr`. The `QFrameAction` is reparented onto the scene root when it is added as a component, so it is destroyed with the 3D scene — meaning the object being used to remember the user's choice does not outlive the scene the choice applies to.

### The change

The preference is tracked in a dedicated `m_frameCountEnabled` bool, independent of the transient `QFrameAction`. `resetItems()` now only detaches the action, and `initialize3DProperties()` recreates and re-attaches it whenever the scene is rebuilt, if the option is enabled.

`SettingsView3D.qml` also gains the `checked` binding the "Show FPS" checkbox was missing, so it reflects the current state like the other controls in that panel.

Three files, all in `qmlui`. **No engine changes.**

**Related Issues:**

There are a series of 5 related PRs + 1 doc PR to improve the 3D View and fix a couple of other issues. This is the first PR.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Test Cases:**
1. Enter the 3D View. Check "Show FPS". Switch to Virtual Console. Switch back to 3D view. Verify that the "Show FPS" setting is still on.

### Testing

Builds clean and `make check` passes, with the same single pre-existing
`RGBText_Test::load()` font failure seen before the change.

Smoke tested.

Linux (Ubuntu and Debian) / Qt 6.11 only. **Not tested on Windows or macOS.**

**Test Results:**
Pass

## Additional Notes

The setting is still runtime-only here — it is not written to the project. Persisting it, along with the rest of the Rendering group, is the follow-up PR, which builds on this one because it touches the same code.

### DIFF
--- a/qmlui/mainview3d.cpp
+++ b/qmlui/mainview3d.cpp
@@ -84,6 +84,7 @@ MainView3D::MainView3D(QQuickView *view, Doc *doc, QObject *parent)
     , m_createItemCount(0)
     , m_sceneGeneration(0)
     , m_frameAction(nullptr)
+    , m_frameCountEnabled(false)
     , m_frameCount(0)
     , m_minFrameCount(0)
     , m_maxFrameCount(0)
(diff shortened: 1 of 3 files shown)


mcallegari:
Thanks

PR #2115, opened 2026-09-01
qmlui: render light beams for LED Bar (Beams) fixtures in the 3D view

mjcmatrix:
## Description

**Summary of Changes:**

Fixtures of type "LED Bar (Beams)" were drawn as a bare cuboid mesh in the 3D view with no light output, so users had to re-type them as Color Changers to see anything. The multi-head code path they use (MultiBeams3DItem plus LightEntity) was an unfinished work in progress.

This PR implements light emission from the heads/emitters of this fixture type (and resolves the past issue mentioned in a comment that would crash the 3D engine).

**Related Issues:**

A further PR will follow to implement a new 3D View Rendering setting: "Fixture Light (%)" to work in tandem with "Smoke amount" to help address the issue where many fixtures casting light together at the stage space cause flat white or full-intensity color, due to fixture light being additive without an exposure or gain setting. This will allow the lighting scene in the 3D view to be made more realistic. 

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Test Cases:**
1. Use fixture(s) of type "LED Bar (Beams)" with 1 x 8 (or more) layout that declare no heads/emitters. Activate the light DMX Channels (e.g. in Simple Desk). Open the 3D view. Observe that the fixture(s) cast light.
2. Test with fixtures of type "LED Bar (Beams)" that declare heads/emitters.

**Test Results:**

image

**Other**

`make check` passes (one pre-existing `RGBText_Test::load()` font failure, identical before and after). Tested on Linux / Qt 6.11 against a rig with four 3-emitter and six single-emitter bars.

**Not tested on Windows or macOS.** Given the root cause was a uniform type mismatch, the Windows backend is worth a look before merging.

**One code path is untested:** the tilt binding. No fixture in my rig has a Tilt channel; it needs a motorised bar such as an Ayrton MagicBlade.

### What this adds

LED Bar (Beams) fixtures currently render as a bare cuboid in the 3D view with no light output, so the only way to see anything is to re-type them as Color Changers. This completes the multi-head code path they already use (`MultiBeams3DItem` + `LightEntity`) so each cell emits a beam.

### Why it did not work

`LightEntity` never declared the pan/tilt angles or the light view/projection matrices that `SpotlightConeEntity` reads as uniforms, so every cone transform collapsed and nothing was drawn.

It also declared `raymarchSteps` as a QML `real`, while `spotlight_scattering.frag` declares the matching uniform as an `int`. Qt3D stores a `real` as a float and passes the raw bytes to `glUniform1iv` without converting, so the shader read the *bit pattern* of the float — 40.0 becomes 1109393408 — and ray marched over a billion times per fragment. That hangs the GPU and resets the display driver, which I believe is the reason the scattering cone carried the `// this hangs your PC` comment. The cone was never the problem; the uniform type was. `Fixture3DItem` has always declared it as an `int`.

### Changes to existing code

**No engine changes.** Everything is in `qmlui`.

`mainview3d.cpp` — one change. `initializeFixture()` only bound a tilt transform when the loaded scene had a base item under the head, which an item drawn without a mesh never has, so motorised beam bars animated a tilt angle nothing applied. The condition now also accepts the no-loader path. `MultiBeams3DItem` is currently the only item type that reaches this code without a scene loader (`PixelBar3DItem`, `Strobe3DItem` and `Generic3DItem` declare no `headEntity`, so `m_headItem` stays null and the whole block is skipped), so nothing else changes behaviour.

`3DView.qml` — `getHead()` returned `undefined` for any gap between `headsNumber` and the real emitter count. The resulting exception aborted `updateFrameGraph()` *after* it had destroyed the previous frame graph, leaving no G-buffer fill, no lighting and no blit, i.e. an unusable 3D view. The head lookups are now guarded, and a shadow map is only requested from emitters that cast shadows.

`LightEntity.qml` / `MultiBeams3DItem.qml` — the rest, described in the commit message. The notable correctness fixes beyond the uniform type: emitters are built once for a valid item ID rather than on every `itemID` change (including the `-1` that `resetItems()` writes on teardown, which rebuilt every emitter into a scene being destroyed); `cleanupScattering()` runs before an emitter is destroyed, since `setupScattering()` re-parents its three cones to the scene root and they were being orphaned; and the emitter grid honours the physical layout only when it matches the mode's head count, because a Pulse LED BAR 320 declares an 8x1 layout while its 4-channel mode declares no head at all, which put its single synthesised emitter at one end of the bar.

### Performance

Cost scales with lit cells, not total cells: `RenderShadowMapFilter` matches no render pass while an emitter is dark, and `SpotlightConeEntity.enabled` does the same for the shading and scattering passes. Ray march steps are additionally scaled by emitter count so a bar costs roughly a fixed budget rather than one spotlight per cell.

### DIFF
--- a/qmlui/mainview3d.cpp
+++ b/qmlui/mainview3d.cpp
@@ -1241,17 +1241,17 @@ void MainView3D::initializeFixture(quint32 itemID, QEntity *fxEntity, const QSce
 
         Qt3DCore::QTransform *transform = getTransform(meshRef->m_headItem);
 
-        if (baseItem != nullptr)
+        // A mesh based fixture only tilts if the loaded scene has a base item
+        // under the head, i.e. it is a moving head or a scanner. An item drawn
+        // without a mesh has no base to look for: its head entity IS the movable
+        // part, so a Tilt channel is enough. Without the second case, a motorized
+        // beam bar would animate a tilt angle that nothing ever applies.
+        if ((baseItem != nullptr || loader == nullptr) && transform != nullptr &&
+            fixture->channelNumber(QLCChannel::Tilt, QLCChannel::MSB) != QLCChannel::invalid())
         {
-            if (fixture->channelNumber(QLCChannel::Tilt, QLCChannel::MSB) != QLCChannel::invalid())
-            {
-                // If there is a base item and a tilt channel,
-                // this is either a moving head or a scanner
-                if (transform != nullptr)
-                    QMetaObject::invokeMethod(meshRef->m_rootItem, "bindTiltTransform",
-                            Q_ARG(QVariant, QVariant::fromValue(transform)),
-                            Q_ARG(QVariant, tiltDeg));
-            }
+            QMetaObject::invokeMethod(meshRef->m_rootItem, "bindTiltTransform",
+                    Q_ARG(QVariant, QVariant::fromValue(transform)),
+                    Q_ARG(QVariant, tiltDeg));
         }
 
         meshRef->m_rootItem->setProperty("focusMinDegrees", focusMin);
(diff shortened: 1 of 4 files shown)


mcallegari:
Thanks for this! :+1: 
Let's hope it doesn't create any subtle regression somewhere else.

PR #2116, opened 2026-09-01
engine/qmlui: persist the 3D View "Rendering" and "Scale" lock settings in the project file

mjcmatrix:
## Description

**Summary of Changes:**

### What this adds

Two 3D view settings are now stored in the project file, so they survive a view switch and a restart instead of reverting to their defaults every time:

- the **Rendering** settings — Quality, Ambient light, Smoke amount and Show FPS;
- the **Scale** X/Y/Z **lock** of the settings panel (the padlock button next to the scale fields of a generic item).

### Why

The Rendering settings were plain runtime members of `MainView3D`, never written to the workspace. A user who dials in ambient light and smoke for a show has to redo it on every launch.

The Scale lock was worse: pure QML state (`IconButton { checked: true }`), so it reverted to "locked" every time the user left the 3D view, because the 3D context — and therefore `SettingsView3D.qml` — is destroyed on a view switch. A user working on a non-uniformly scaled item had to unlock it again after every trip to the Virtual Console or the 2D view.

### Changes to `engine` — to be reviewed and approved

_To be reviewed and approved in the context of the linked set of PRs submitted to improve the 3D view plus other fixes._

The values move into `MonitorProperties`, alongside the stage type and the environment grid. The rendering settings are saved and loaded under a new `<Rendering>` element of the `<Monitor>` node, and the scale lock as a `<ScaleLock>` element next to it — both guarded by `QMLUI` exactly like `<StageItem>`.

A project without those elements keeps the previous defaults — High quality, 0.6 ambient, 0.8 smoke, FPS off, scale axes locked — so **existing projects load and render exactly as they do today**. `monitorproperties_test` is extended with a save/load round trip for the new elements and with checks in the defaults and reset paths.

That round trip test is guarded with `QMLUI`, since the new elements are only written by the QML UI build and the test would otherwise fail in a QLC+ 4 build; it is named `view3DSettingsXML()` as it now covers more than the rendering settings.

### Changes in `qmlui`

`MainView3D`'s getters and setters proxy to `MonitorProperties`, the same way `stageIndex()` already does, and mark the document modified so that changing one of these settings on its own still prompts a save.

On project load, and when the 3D view becomes visible, `MainView3D::applyRenderSettings()` re-applies the persisted values to the running scene and notifies QML without touching the modified flag. `setFrameCountEnabled()` is split so that the load path can apply the FPS state without persisting it or marking the project modified.

The scale lock only affects the settings panel, so `slotRefreshView()` simply notifies the QML side. The lock button binds to the new `scaleLocked` property; since clicking a checkable `Button` assigns `checked` and therefore breaks that binding, a `Connections` handler restores it whenever the value changes on the C++ side.

**Related Issues:**
- Builds on the "Show FPS" fix; that touches the same code and should land first. https://github.com/mcallegari/qlcplus/pull/2114
- This is the prerequisite for the **Fixture light** Rendering setting (a PR to follow) which adds one more attribute to the `<Rendering>` element introduced here.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Test Cases:**
1. Load/create a new project. Go to 3D View. Change the Rendering settings. Save project. Close QLC+ V5. Reopen. Open the same project. Verify the settings have persisted.
2. Add a generic item to the 3D view and select it. Unlock the Scale axes and give X, Y and Z different values. Switch to the Virtual Console and back to the 3D View: the lock is still open and the values are unchanged. Save, restart, reload the project: the lock is still open.

**Test Results:**
Test case 1: Pass
Test case 2: Pass

Smoke test for regressions/bugs: Passed.

**Other**

Builds clean and `make check` passes, with the same single pre-existing `RGBText_Test::load()` font failure seen before the change, plus the new `monitorproperties_test` coverage described above. `monitorproperties_test` was also run against a non-`qmlui` (QLC+ 4) build, where the round trip test is skipped.

Linux / Qt 6.11 only. **Not tested on Windows or macOS.**

## Additional Notes

None.

### DIFF
--- a/engine/src/monitorproperties.cpp
+++ b/engine/src/monitorproperties.cpp
@@ -50,6 +50,12 @@
 #define KXMLQLCMonitorLightEmitter  QStringLiteral("LightEmitter")
 #define KXMLQLCMonitorStageItem     QStringLiteral("StageItem")
 #define KXMLQLCMonitorMeshItem      QStringLiteral("MeshItem")
+
+#define KXMLQLCMonitorRendering         QStringLiteral("Rendering")
+#define KXMLQLCMonitorRenderQuality     QStringLiteral("Quality")
+#define KXMLQLCMonitorRenderAmbient     QStringLiteral("Ambient")
+#define KXMLQLCMonitorRenderSmoke       QStringLiteral("Smoke")
+#define KXMLQLCMonitorRenderShowFPS     QStringLiteral("ShowFPS")
 #define KXMLQLCMonitorItemName      QStringLiteral("Name")
 #define KXMLQLCMonitorItemRes       QStringLiteral("Res")
 
@@ -79,6 +85,13 @@
 #define GRID_DEFAULT_HEIGHT 3
 #define GRID_DEFAULT_DEPTH  5
 
+/* 3D view rendering defaults. RENDER_DEFAULT_QUALITY matches
+   MainView3D::HighQuality and the other two match the MainView3D members
+   they replace, so a project without a <Rendering> node looks unchanged. */
+#define RENDER_DEFAULT_QUALITY  2
+#define RENDER_DEFAULT_AMBIENT  0.6
+#define RENDER_DEFAULT_SMOKE    0.8
+
 MonitorProperties::MonitorProperties()
     : m_font(QFont("Arial", 12))
     , m_displayMode(DMX)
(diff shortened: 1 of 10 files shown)


mjcmatrix:
Added a second commit persisting the Scale lock; title/description updated

PR #2117, opened 2026-09-01
engine/qmlui: add a "Fixture light" intensity setting to the 3D view

mjcmatrix:
## Description

**Summary of Changes:**

### What this adds

A **Fixture light** percentage in the 3D view "Rendering" settings: a global multiplier on the light fixtures cast onto surfaces.

### Why

The deferred renderer accumulates emitters additively (`SpotlightShadingFilter` blends `One/One`) and `gamma_correct.frag` tone maps the sum with `1 - exp(-hdr)`. That means:

| fixtures at full on the same area | accumulated | on screen |
|---|---|---|
| 1 | 1.0 | 0.63 |
| 8 | 8.0 | 0.9997 |
| 18 | 18.0 | 1.0 |

A rig with enough fixtures aimed at the stage saturates it to flat white and there is no way to pull it back. **Ambient light** sets how bright the set is on its own and **Smoke amount** scales the volumetric beams, but nothing scaled the light landing on surfaces.

Old3DOversaturation

I considered a camera-style exposure control on the tone map first and rejected it: because the fixtures are already saturated while the set is still in the linear region, lowering exposure darkens the set faster than the fixtures and makes the imbalance worse, not better.

### Changes to `engine` — for review and approval

`MonitorProperties` gains one `qreal` member with an inline getter/setter, one attribute (`FixtureLight`) on the existing `<Rendering>` element, and default and reset handling. It follows exactly the pattern of the neighbouring render settings; no existing member, signature or XML element changes.

The default is 1.0 and the loader leaves it at 1.0 when the attribute is absent, so **every existing project renders byte-identically**. `monitorproperties_test` is extended to cover the new value in `defaults()`, `reset()` and the XML round-trip.

This sits in `engine` only because that is where the other 3D view rendering settings are stored so they persist in the workspace. If you would rather it were held elsewhere, or not persisted at all, say so and I will move it.

### Changes in `qmlui`

`MainView3D` gains a proxying `Q_PROPERTY` alongside `ambientIntensity` and `smokeAmount`, emitted from `applyRenderSettings()`. The value reaches the shader as a single `Parameter` on `SpotlightShadingFilter`, which already supplies `albedoTex`, `shadowTex` and `useShadows` the same way, and one `uniform float` in `spotlight_shading.frag`.

`PixelBar3DItem` and `Strobe3DItem` scale their own `lightIntensity` instead.
Fixtures that cast cones contribute only through the spotlight passes — their meshes get `bloom` 0 in `inspectEntity()` — while these two cast nothing and contribute only their emissive surface. So the multiplier reaches every fixture's light contribution exactly once.

The volumetric beams are deliberately left alone: `Smoke amount` already scales them in `spotlight_scattering.frag`.

**Related Issues:**
- Depends on the `<Rendering>` persistence work; that needs to land first.  https://github.com/mcallegari/qlcplus/pull/2116
- Documentation PR: https://github.com/mcallegari/qlcplus-docs/pull/133

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

- Tested on Linux / Qt 6.11 only. **Not tested on Windows or macOS.**

**Test Cases:**
1. Place multiple fixtures including "LED Bar (Beams)" fixture types (used for stage color wash), e.g. 320 bright LED fixtures, around a stage, to mirror a real venue, all illuminating the stage area, and turn on all the lights. Observe flat white effect or maximum-intensity color saturation. The lights are all additive. This does not reflect the true lighting on the stage being modelled without the new "Fixture light" setting working in tandem with "Smoke amount". Reduce both the "Fixture light" and "Smoke amount" settings. Observe that the "Fixture light" settings controls the level of light cast by all fixtures onto surfaces.

**Test Results:**
1. Passed

Smoke/regression testing passed.

Result:
New3DFeatures

## Additional Notes

- New option so a parallel `qlcplus-docs` PR has been prepared: https://github.com/mcallegari/qlcplus-docs/pull/133

### DIFF
--- a/engine/src/monitorproperties.cpp
+++ b/engine/src/monitorproperties.cpp
@@ -55,6 +55,7 @@
 #define KXMLQLCMonitorRenderQuality     QStringLiteral("Quality")
 #define KXMLQLCMonitorRenderAmbient     QStringLiteral("Ambient")
 #define KXMLQLCMonitorRenderSmoke       QStringLiteral("Smoke")
+#define KXMLQLCMonitorRenderFxLight     QStringLiteral("FixtureLight")
 #define KXMLQLCMonitorRenderShowFPS     QStringLiteral("ShowFPS")
 #define KXMLQLCMonitorItemName      QStringLiteral("Name")
 #define KXMLQLCMonitorItemRes       QStringLiteral("Res")
@@ -91,6 +92,8 @@
 #define RENDER_DEFAULT_QUALITY  2
 #define RENDER_DEFAULT_AMBIENT  0.6
 #define RENDER_DEFAULT_SMOKE    0.8
+/* Unscaled: a project that has never touched this renders exactly as before */
+#define RENDER_DEFAULT_FXLIGHT  1.0
 
 MonitorProperties::MonitorProperties()
     : m_font(QFont("Arial", 12))
@@ -104,6 +107,7 @@ MonitorProperties::MonitorProperties()
     , m_renderQuality(RENDER_DEFAULT_QUALITY)
     , m_ambientLightIntensity(RENDER_DEFAULT_AMBIENT)
     , m_smokeAmount(RENDER_DEFAULT_SMOKE)
+    , m_fixtureLightIntensity(RENDER_DEFAULT_FXLIGHT)
     , m_showFPS(false)
     , m_showLabels(false)
 {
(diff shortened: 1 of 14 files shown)


mjcmatrix:
I believe the LED Bar (Beams) merge has affected the base of this. I will fix this tomorrow

PR #2123, opened 2026-09-03
resources: improve the American DJ Jolt 300 definition

mjcmatrix:
## Description

**Summary of Changes:**

Expands the Jolt 300's colour macro channel into the 64 macros of the manual's chart, sets the fixture type to LED Bar (Pixels), adds the missing `Lumens` and white LED colour temperature, and normalises `LED's` to `LEDs` in the shutter capability labels.

**Related Issues:**

None.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Test Cases:**

| # | Case | Result |
|---|------|--------|
| 2 | Colour macro DMX ranges tile 0–255 with no gap or overlap | Pass — asserted while generating from the chart |
| 3 | Every mode channel reference resolves after the macro channel was rewritten | Pass — no unresolved references |
| 4 | Macro channel appears in exactly the modes the manual's chart marks | Pass — 8 Channel (ch5), A10 Channel (ch5), A18 Channel (ch13) |
| 5 | Every mode's channel order re-checked against the manual's DMX chart | Pass — including B10 placing White on ch7, and B18 splitting the whites to ch13–15 |
| 6 | Definition loads in QLC+ 5 | Pass — no cache or capability warnings |
| 7 | 2D view head geometry | Pass — three heads, matching the three RGB sections |
| 8 | Tested on real fixture hardware - core capabilities only | Works |

**Test Results:**

All pass. Tested on Linux, Qt 6.11.2.

## Additional Notes

**Fixture type.** The Jolt 300 has three independently controllable RGB sections, which the definition already declared as three heads with a 3×1 layout, but its type was `Other`, which renders nothing meaningful in the 3D view. It is typed here as LED Bar (Pixels), the three sections being pixels of the bar; this follows the maintainer's review of the Pulse LED BAR 320, where a bar of this kind was judged a pixel bar rather than a set of beam emitters. An earlier revision of this PR used LED Bar (Beams).

**Colour macros.** The channel was a bare `Preset="ColorMacro"`, i.e. one undifferentiated 0–255 range with no capability data. The manual's chart gives an RGB **and white** intensity per macro, but a `Res1` swatch has no white channel. Dropping white would render macro 64 — white LEDs only, RGB all zero — as black. The swatch therefore folds white in additively with a clamp, `min(255, R + W)` per component, the way the emitters mix. Macro 64 comes out white and macro 32 stays white; the cost is that the macro 1–31 pastels read slightly less saturated in the swatch than their raw RGB triples. This affects only the swatch and the Stage Wizard's colour picking — the rendered beam colour comes from the RGBW intensity channels, not from `Res1`. DMX 0 is left as a plain capability rather than a `ColorMacro` of black.

Macros are named as the manual names them (`Color 1` … `Color 64`) rather than given invented descriptive names; the swatches carry the colour.

**Lumens.** 12900, from ADJ's published 4113 lux at 1 m over the 120° beam: `flux = lux * distance^2 * 2*pi*(1 - cos(beam / 2))`. The figure is for the whole fixture; the 3D view divides it between the heads it draws. Colour temperature 6000 K is ADJ's figure for the white LEDs.

**Left alone deliberately.** The manual gives Program 21 as 156–161 and Program 22 as 162–169, while the definition has 156–162 / 163–169. Every other program in that table is exactly 7 values wide and the manual's own numbers break the pattern only here, so this reads as a typo in the manual and the existing values were kept.

## Source(s) of Information

- ADJ Jolt 300 User Manual — DMX chart pages 11–14, colour macro chart pages 15–16
- ADJ Jolt 300 cut sheet — light source, sections, electrical, dimensions
- Product page: www.adj.com/jolt-300

### DIFF
--- a/resources/fixtures/American_DJ/American-DJ-Jolt-300.qxf
+++ b/resources/fixtures/American_DJ/American-DJ-Jolt-300.qxf
@@ -8,7 +8,7 @@
  </Creator>
  <Manufacturer>American DJ</Manufacturer>
  <Model>Jolt 300</Model>
- <Type>LED Bar (Beams)</Type>
+ <Type>LED Bar (Pixels)</Type>
  <Channel Name="Red" Preset="IntensityRed"/>
  <Channel Name="Green" Preset="IntensityGreen"/>
  <Channel Name="Blue" Preset="IntensityBlue"/>
--- a/resources/fixtures/American_DJ/American-DJ-Jolt-300.qxf
+++ b/resources/fixtures/American_DJ/American-DJ-Jolt-300.qxf
@@ -4,11 +4,11 @@
  <Creator>
   <Name>Q Light Controller Plus</Name>
   <Version>4.12.7 GIT</Version>
-  <Author>David Gouronc</Author>
+  <Author>David Gouronc + Matt Carter</Author>
  </Creator>
  <Manufacturer>American DJ</Manufacturer>
  <Model>Jolt 300</Model>
- <Type>Other</Type>
+ <Type>LED Bar (Beams)</Type>
  <Channel Name="Red" Preset="IntensityRed"/>
  <Channel Name="Green" Preset="IntensityGreen"/>
  <Channel Name="Blue" Preset="IntensityBlue"/>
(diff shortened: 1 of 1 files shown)


resources/fixtures/American_DJ/American-DJ-Jolt-300.qxf
```diff
@@ -4,11 +4,11 @@
  <Creator>
   <Name>Q Light Controller Plus</Name>
   <Version>4.12.7 GIT</Version>
-  <Author>David Gouronc</Author>
+  <Author>David Gouronc + Matt Carter</Author>
  </Creator>
  <Manufacturer>American DJ</Manufacturer>
  <Model>Jolt 300</Model>
- <Type>Other</Type>
+ <Type>LED Bar (Beams)</Type>
```
mcallegari: This is not a beam bar. Change it to pixel.

mjcmatrix: I should not have submitted this instead of addressing the root cause. I thought it was the best option for now visually, to be revised later. Same as the last bar fixture PR, the glow in the 3D view is currently unrealistic, in reality it is a bright stage wash more like wide beams. I will use this as a local workaround until we address the dim glow of Pixel bars. Taken in context with the whole set of 3D view PRs (submitted and still in development pending submission) you might see why I did this - but it doesn't make it the right choice.

I may at some point begin work on an "LED Bar (Pixels)" light improvement for beam-like/wash-style high brightness fixtures but will have to consult you closely on that one.

(I will amend this PR to pixel.)

mjcmatrix: Corrected type to LED Bar (Pixels)

mcallegari:
Thanks

PR #2125, opened 2026-09-03
resources: add the Equinox Fusion 140 and 260ZR

mjcmatrix:
## Fixture Information

**Manufacturer:** Equinox

**Model:** Fusion 140 and Fusion 260ZR — two moving head washes, one definition each

**Mode(s) included:**

- Fusion 140 — 2 Channel, 10 Channel, 16 Channel
- Fusion 260ZR — 2 Channel, 8 Channel, 17 Channel, 25 Channel

**Channels used:** as per the mode names. Pan and tilt are 16-bit in the Fusion 140's 16 channel mode and in the Fusion 260ZR's 17 and 25 channel modes (coarse plus fine pairs); all other channels are 8-bit.

## Testing

**Test cases and results**

| # | Case | Result |
|---|------|--------|
| 1 | `fixtures-tool.py --validate` on both definitions | Pass — 0 errors |
| 2 | Every mode's channel order checked against each manual's DMX chart | Pass |
| 3 | Fusion 140 colour macro ranges match the manual's 37-entry table | Pass |
| 4 | Fusion 140 physical block against the manual | Pass — 130 W, 310x262x155 mm, 4.5 kg, 35 degrees, pan 540, tilt 180, 3-pin |
| 5 | Fusion 260ZR physical block against the manual | Pass — 290 W, 414x324x221 mm, 9.7 kg, zoom 6-50 degrees, pan 540, tilt 270, 3-pin |
| 6 | Both definitions load in QLC+ 5 | Pass — no cache or capability warnings |
| 7 | 2D view head geometry | Pass — Fusion 140 single head, Fusion 260ZR three heads, one per LED circle |
| 8 | `FixturesMap.xml` regeneration is idempotent after both additions | Pass — no drift |
| 9 | Tested on real fixture hardware - core capabilities only | Works |

Software tests on Linux, Qt 6.11.2.

## Source(s) of Information

- Fusion 140 User Manual (order code EQLED074) — DMX charts, specifications
  www.prolight.co.uk/product_manual/Fusion%20140_Manual.pdf
- Fusion 260ZR User Manual (order code EQLED024) — DMX charts, specifications
  www.prolight.co.uk/product_manual/Fusion%20260ZR_Manual.pdf

## Notes

Two fixtures in one PR because each adds a line to the `Equinox` block of `FixturesMap.xml` at the same position, so as separate PRs the second to be merged would conflict. They are otherwise independent: one commit per fixture, and the definitions do not interact.

### Fusion 140

A 7 x 18 W RGBWAUV moving head wash with a fixed 35 degree beam.

**Lumens.** 3400, from the manual's 2930 lux at 2 m over the 35 degree beam: `flux = lux * distance^2 * 2*pi*(1 - cos(beam / 2))`. The figure is for the whole fixture; the 3D view divides it between the heads it draws.

**Dimmer and strobe are separate channels.** The manual puts a plain 0–100% master dimmer on CH6 (CH3 in the 10 channel mode) and strobe on CH7 (CH4), with 0–9 inactive, and the definition follows it.

**Pan/tilt speed direction.** The manual gives no direction for the pan/tilt speed channel, so `SpeedPanTiltSlowFast` is assumed. The 260ZR manual documents its equivalent channel as fast-slow, so this may want inverting if anyone can confirm against hardware.

**Colour macro names** are the manual's own labels, verbatim — which means "Blue" and "White" each appear twice, at different DMX ranges. Disambiguating them would mean inventing names the manual does not use.

### Fusion 260ZR

A 19 x 12 W RGBW moving head wash with a 6–50 degree motorised zoom.

**Heads.** The LEDs sit in three concentric circles and the 25 channel mode addresses each circle's RGBW separately, so that mode declares a head per circle, each holding only its four colour channels. The 8 and 17 channel modes drive all the LEDs together and declare no head.

**Lumens.** 14900, from the manual's 6316 lux at 2 m at the 50 degree end of the zoom, by the same formula. The wide end is used deliberately — the manual's 6° figure of 40098 lux at 2 m works out to roughly 1380 lm, an order of magnitude lower, because a narrow-beam lux reading measures the hot spot rather than total output.

**Channel details taken from the manual rather than assumed.** Sound sensitivity has DMX 0 as blackout and 1–255 as the sensitivity ramp. Strobe is inactive at 0–15 and again at 255, so it is a custom channel rather than the `ShutterStrobeSlowFast` preset. Pan/tilt speed is documented as fast-slow. Reset is 0–127 inactive, 128–255 reset on a five second hold.

### DIFF
--- a/resources/fixtures/Equinox/Equinox-Fusion-140.qxf
+++ b/resources/fixtures/Equinox/Equinox-Fusion-140.qxf
@@ -1,0 +1,134 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE FixtureDefinition>
+<FixtureDefinition xmlns="http://www.qlcplus.org/FixtureDefinition">
+ <Creator>
+  <Name>Q Light Controller Plus</Name>
+  <Version>4.14.4</Version>
+  <Author>Matt Carter</Author>
+ </Creator>
+ <Manufacturer>Equinox</Manufacturer>
+ <Model>Fusion 140</Model>
+ <Type>Moving Head</Type>
+ <Channel Name="Pan" Default="127" Preset="PositionPan"/>
+ <Channel Name="Pan fine" Preset="PositionPanFine"/>
+ <Channel Name="Tilt" Default="127" Preset="PositionTilt"/>
+ <Channel Name="Tilt fine" Preset="PositionTiltFine"/>
+ <Channel Name="Pan/tilt speed" Preset="SpeedPanTiltSlowFast"/>
+ <Channel Name="Master dimmer" Preset="IntensityMasterDimmer"/>
+ <Channel Name="Strobe">
+  <Group Byte="0">Shutter</Group>
+  <Capability Min="0" Max="9" Preset="ShutterOpen">No function</Capability>
+  <Capability Min="10" Max="255" Preset="StrobeSlowToFast">Strobe (slow-fast)</Capability>
+ </Channel>
+ <Channel Name="Red" Preset="IntensityRed"/>
+ <Channel Name="Green" Preset="IntensityGreen"/>
+ <Channel Name="Blue" Preset="IntensityBlue"/>
+ <Channel Name="White" Preset="IntensityWhite"/>
+ <Channel Name="Amber" Preset="IntensityAmber"/>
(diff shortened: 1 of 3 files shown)


mcallegari:
Thanks

PR #2126, opened 2026-09-03
resources: add the Ledj Stage Par CZ200 3200K

mjcmatrix:
## Fixture Information

**Manufacturer:** Ledj

**Model:** Stage Par CZ200 3200K

**Mode(s) included:** 2 channel, 3 channel, 5 channel

**Channels used:** 2, 3 and 5 respectively. All channels are 8-bit.

## Testing

**Test cases and results**

| # | Case | Result |
|---|------|--------|
| 2 | Every mode's channel order checked against the manual's DMX chart | Pass |
| 3 | Strobe and dimmer curve capability bands match the manual | Pass |
| 4 | Physical block against the manual's specifications | Pass — 205 W, 370x275x390 mm, 4.3 kg, zoom 13-50 degrees, 5-pin |
| 5 | Definition loads in QLC+ 5 | Pass — no cache or capability warnings |
| 6 | Tested on real fixture hardware - core capabilities only | Works |

Tested on Linux, Qt 6.11.2.

## Source(s) of Information

- Stage Par CZ200 3200K User Manual (order code LEDJ193) — DMX charts, specifications

## Notes

A 200 W warm white COB par with a 13–50 degree motorised zoom. There is no colour mixing: the only intensity control is the master dimmer, so the definition has no colour channels.

**Lumens.** 13800, from the manual's 5877 lux at 2 m at the 50 degree end of the zoom: `flux = lux * distance^2 * 2*pi*(1 - cos(beam / 2))`. As with other zoom fixtures the wide end is used, since the manual's 13° figure of 39362 lux at 2 m measures the hot spot rather than total output. Colour temperature is set to 3200 K, which is what distinguishes this model from the Dual White variant and is part of the model name.

**Fixture type.** `Color Changer`, matching how the library types other single-colour pars. `Dimmer` would arguably be more literal for a fixture with no colour mixing, but both render the same PAR mesh and `Color Changer` is the established convention here. Happy to switch it if you prefer.

### DIFF
--- a/resources/fixtures/FixturesMap.xml
+++ b/resources/fixtures/FixturesMap.xml
@@ -1236,6 +1236,7 @@
     <F n="Ledj-Slimline-9Q8" m="Slimline 9Q8"/>
     <F n="Ledj-Stage-Color-24" m="Stage Color 24"/>
     <F n="Ledj-Stage-Color-Quad" m="Stage Color Quad"/>
+    <F n="Ledj-Stage-Par-CZ200-3200K" m="Stage Par CZ200 3200K"/>
     <F n="Ledj-Tri-LED-back-drop-controller" m="Pro Tri LED back drop controller"/>
   </M>
   <M n="LedProLight">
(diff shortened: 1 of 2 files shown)


mcallegari:
Thanks

PR #2129, opened 2026-09-04
engine/qmlui: render the 3D view photometrically, behind an experimental "Lumens" setting

mjcmatrix:
## Description

**Summary of Changes:**

Every fixture in the 3D view emits the same amount of light regardless of what it is, and that light lands just as brightly however far it has travelled. A large stage wash and a small LED PAR lay identical pools, and a fixture hung at 2 m looks the same as the same fixture hung at 4 m. This adds an opt-in **Lumens (experimental)** setting to the 3D view "Rendering" settings that renders the rig photometrically instead, so a mixed rig shows the balance it would really have.

Three things change when it is ticked:

* **Each fixture is scaled by its real output.** The `Lumens` value already present in fixture definitions, spread over the solid angle of the beam at the widest the lens opens, so what reaches the shader is luminous intensity in candela. That, not flux, is what the renderer's light intensity behaves like, because nothing in the shading divides by the area the cone covers. Two fixtures of equal lumens therefore differ by their beam angle, as they do in the room. The reference is the brightest emitter in the project by intensity, so that fixture keeps the brightness it had.
* **Zoom concentrates the light.** A moving beam closing down now brightens the pool it lays rather than only shrinking it, by the ratio of the cone's solid angle at its widest to its solid angle now.
* **Light falls off with the distance it travels**, by the inverse square, as illuminance does in the room. The distance at which light lands unscaled is *derived* from the project rather than asked for: the mean height above the floor of the fixtures placed in it, recomputed whenever a fixture is added, removed or moved.

Code:

* `MonitorProperties` gains a `useFixtureLumens` flag, saved as the `Lumens` attribute of the existing `<Rendering>` node, so the choice persists in the workspace.
* `MainView3D` computes each fixture's per-emitter lumens and candela, the project's reference candela (the brightest emitter) and its reference throw (the rig's mean hang height), and exposes them to QML.
* `Fixture3DItem`, `MultiBeams3DItem` / `LightEntity`, `PixelBar3DItem` and `Strobe3DItem` fold a `lumensScale` into their `lightIntensity`, so the scale reaches every fixture type exactly once.
* `spotlight_shading.frag` applies the inverse square term against the derived reference, gated on the same opt-in.
* A checkbox in the 3D view "Rendering" settings, labelled "Lumens (experimental)".

The setting is **off by default and existing workspaces are untouched**: a project with no `Lumens` attribute loads with it off and renders exactly as it does today.

This PR is part of a series of related changes to improve the realism of the 3D View.

**Related Issues:**

This sits on top of two of my open PRs, both in the same 3D view series:

* Depends on #2116 — *engine/qmlui: persist the 3D View "Rendering" and "Scale" lock settings in the project file*. That PR adds the `<Rendering>` node this feature stores its `Lumens` attribute in.
* Depends on #2117 — *engine/qmlui: add a "Fixture light" intensity setting to the 3D view*. That PR adds the global gain this feature is designed to compose with: **Lumens** sets the balance between fixtures, **Fixture light** sets the overall level. #2117 is itself based on #2116.
* Builds on #2115 — *qmlui: render light beams for LED Bar (Beams) fixtures in the 3D view* (merged), whose per-head beams this feature scales individually.

Because #2116 and #2117 are not merged yet, this PR is based on the #2117 branch and its diff therefore includes both of theirs. The commits new to this PR are the last four:

| commit | |
|---|---|
| `baa21851e` | scale 3D view fixture light by the fixture's Lumens |
| `635892115` | scale 3D view fixture light by intensity, not by flux |
| `c5e201cc5` | fall off 3D view fixture light over the distance it travels |
| `b90b1a8e7` | mark the 3D view "Lumens" setting experimental |

Everything else in the diff belongs to #2116 and #2117 and should be reviewed there. Reviewing or merging those two first will shrink this one to the four commits above.

Plus a merge of `master`. #2115 reached `master` by a different route than this branch's history took, so `LightEntity.qml` and `MultiBeams3DItem.qml` conflicted even though both sides carry the same code; they are resolved to `master`'s content plus the properties this PR adds, and the diff for those two files is now only those additions.

Documentation is a companion change to docs.qlcplus.org: mcallegari/qlcplus-docs#135.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Test Cases:**

Brightness figures are measured from screenshots by inverting the tone map in `gamma_correct.frag` and subtracting the ambient level, so they are proportional to the light actually cast. Render quality High, ambient 5%, smoke off.

*Relative output.* Three PARs hung at the same height so the falloff term is equal across them: A at 10000 lm / 25°, B at 10000 lm / 50°, C at 20000 lm / 50°. Predicted intensities against A are 1.000, 0.253 and 0.506.

1. All three with the setting off — equally bright, i.e. unchanged.
2. All three with the setting on — each scaled to its candela against the brightest emitter.
3. A fixture whose definition leaves Lumens unset renders at the reference level, never dark.
4. Effect of the flux-to-candela correction across fixture types, one of each, against the previous revision of this branch.

*Distance falloff.* Two identical PARs, one at 2 m and one at 4 m, so a 3 m derived reference.

5. Setting off — the two lay equally bright pools, as every release so far has rendered.
6. Setting on — the lower fixture lays the brighter pool, by the inverse square of the distance.
7. The same rig doubled in size, at 4 m and 8 m in a taller room, to confirm the derived reference makes the look independent of the venue's scale.
8. Moving a fixture up or down in the 3D view rebalances the rig live.

*Persistence and compatibility.*

9. Setting persists to the workspace on save and is restored on load.
10. Project with no `Lumens` attribute loads with the setting off and renders unchanged.
11. `MonitorProperties` unit tests cover defaults, XML round trip and reset.

*On a real rig.*

12. A full rig with the setting off and on, judged by eye.

**Test Results:**

| # | Result | Notes |
|---|--------|-------|
| 1 | Pass | Equal within sampling noise |
| 2 | Pass | Measured 1.000 / 0.260 / 0.506 against predicted 1.000 / 0.253 / 0.506. Note A and B carry identical lumens and differ four-fold on beam angle alone |
| 3 | Pass | Renders at the reference level, the same as with the setting off |
| 4 | Pass | Against the flux-scaled revision: a 120° blinder falls to 0.20, a 35° wash rises to 2.03, a 40° bar to 1.54, and two fixtures sharing a 50° widest lens are unchanged at 1.00 and 1.01 |
| 5 | Pass | Ratio 1.000 |
| 6 | Pass | Ratio 1.000 → 4.14, against 4.0 predicted by 1/d²; the excess is the emitter sitting a little below its stated hang height |
| 7 | Pass | Ratio 4.10, and the doubled rig renders at 0.96 and 0.98 of the smaller rig's levels — the look does not change with the size of the venue |
| 8 | Pass | Dragging a fixture down makes its light more intense, dragging it up less intense |
| 9 | Pass | `<Rendering ... Lumens="1"/>` written on save; reopened project reproduced the render without touching the checkbox |
| 10 | Pass | Attribute stripped by hand; matches test 1 |
| 11 | Pass | `monitorproperties_test`: 9 passed, 0 failed on this branch, 10 passed, 0 failed on a build merged with the beam softness work |
| 12 | Pass | See screenshots below |

Tested on Linux, Qt 6.11.2, Mesa llvmpipe.

## Screenshots

Screenshots, with the setting off and on. The flat magenta of the "off" renders is every fixture emitting the same amount of light, at the same intensity however far it has travelled.

**With the LED Bar (Beams) stage wash fixtures at uniform height**

`WithoutLumens.png` - current state
WithoutLumens
`WithLumens.png` - shows how this feature can help make a scene more realistic
WithLumens

**With the LED Bar (Beams) stage wash fixtures at different heights**

`WithoutLumens-Demo1.png` - current state - intensity is the same regardless of distance
WithoutLumens-Demo1
`WithLumens-Demo1.png` - shows how distance now affects intensity
WithLumens-Demo1

## Additional Notes

**Why the feature is marked experimental.** The look it produces may still change, and two limitations are worth stating plainly. It depends on fixture definitions carrying sensible `Lumens` and lens angle figures, and many do not (see coverage below). And the derived reference averages every placed fixture together, so a rig that mixes hung fixtures with floor standing ones pulls the reference below the height the hung fixtures throw from, and an uplighter close to what it lights renders very hot. Weighting or excluding near-floor fixtures is left for a later change; the **Fixture light** gain from #2117 pulls the frame back in the meantime.

**Why intensity rather than flux.** The first revision of this work fed each fixture's luminous flux straight into the renderer's light intensity. Nothing in `spotlight_shading.frag` divides by the area the cone covers — the gobo mask is sampled in normalised cone coordinates, so a wide beam lights a larger patch at the same brightness per pixel as a narrow one. The uniform therefore behaves as luminous intensity, and putting flux in it made two fixtures of equal output equally bright however tightly either one focused. Dividing by the solid angle of the beam, `2π(1 - cos(angle / 2))`, fixes that and makes zoom a live multiplier on top. It can push the intensity above 1.0, which is fine: the frame buffer is RGBA32F and `gamma_correct.frag` tone maps with `1 - exp(-x)`, so it rolls off rather than clipping.

**Why the falloff reference is derived and not a setting.** A prototype exposed it as a "Light falloff" distance in metres. That was wrong twice over. The 3D view is already a to-scale model of the venue — the grid is given in metres and fixtures sit at the heights they are really hung at — so the difference between a club and an arena is in the project already, and any number typed in can disagree with the rig actually built. More concretely, a reference distance reaches the shader only as a constant multiplying every fragment, which is exactly what the **Fixture light** gain already does; it would have been a second global exposure control wearing a physical name. Deriving it from the rig's own scale leaves the frame's overall exposure where it was and changes only the balance between near and far throws, which test 7 measures.

The reference depends only on where fixtures hang, never on pan and tilt. A reference that tracked where the heads point would make the whole frame breathe as they move.

**Why relative rather than absolute.** Each fixture is measured against the brightest emitter in the project, so the brightest fixture renders close to as it always has and everything else falls in below it. An absolute reference was rejected because a rig made entirely of small LED PARs would render uniformly dark and need the gain wound up to compensate. The trade-off is that adding a brighter fixture rescales everything below it.

**Per emitter, not per fixture.** `QLCPhysical` hangs off `QLCFixtureMode` (falling back to the definition's global `<Physical>`) and there is no per-head physical, so the value describes the whole fixture. It is divided between the emitters the 3D view draws — one lamp per channel for a Dimmer, one per head otherwise. Without that an 8 cell bar would cast eight times the light of a moving head carrying the same figure. That division is unchanged by the candela and falloff commits and was measured on an earlier revision: a 3-head bar's cells at 0.738 of a 2-head bar's, against 0.737 expected.

**A fixture with no lens angle** falls back to the same 30° the cone geometry already falls back to, so the photometry describes the cone actually drawn. That constant is shared by every such fixture and so does not disturb their balance against each other.

**Backwards compatibility was the governing constraint.** Nothing about an existing project may look different after this change, and there are three independent guards:

1. The setting defaults to off, and a workspace saved before this change has no `Lumens` attribute, so it loads off. Existing projects render exactly as they do today (tests 5 and 10).
2. With the setting on, a fixture whose definition has no lumens data renders at the reference level — the brightness it has always had. It never goes dark (test 3).
3. With the setting on in a project where no fixture has the data, the reference is 0, every scale collapses to 1.0 and the falloff term is skipped, so the whole feature is a no-op rather than a surprise.

**Library coverage.** The data exists but is far from universal, which is why "unknown" had to be a first-class case rather than an edge case:

| | count | share |
|---|---|---|
| Fixture definitions in the library | 1735 | |
| ...with a `<Physical>` block | 1735 | 100% |
| ...giving a non-zero `Lumens` | 612 | 35% |
| Fixture modes | 5040 | |
| ...resolving to a non-zero `Lumens` | 2020 | 40% |

Of the 612 definitions that do carry the value, 286 are Moving Heads and 190 Colour Changers — the cone-casting types this setting affects most — plus 41 Scanners, 45 LED bars and 17 Strobes. As definitions gain the value the setting simply gets more accurate.

**Scope of the falloff.** Only the surface shading pass is affected. The volumetric beam in `spotlight_scattering.frag` would need the term applied per raymarch sample and is left alone for now, so a beam in smoke still reads at full length.

### DIFF
--- a/engine/src/monitorproperties.cpp
+++ b/engine/src/monitorproperties.cpp
@@ -55,6 +55,7 @@
 #define KXMLQLCMonitorRenderQuality     QStringLiteral("Quality")
 #define KXMLQLCMonitorRenderAmbient     QStringLiteral("Ambient")
 #define KXMLQLCMonitorRenderSmoke       QStringLiteral("Smoke")
+#define KXMLQLCMonitorRenderFxLight     QStringLiteral("FixtureLight")
 #define KXMLQLCMonitorRenderShowFPS     QStringLiteral("ShowFPS")
 #define KXMLQLCMonitorItemName      QStringLiteral("Name")
 #define KXMLQLCMonitorItemRes       QStringLiteral("Res")
@@ -91,6 +92,8 @@
 #define RENDER_DEFAULT_QUALITY  2
 #define RENDER_DEFAULT_AMBIENT  0.6
 #define RENDER_DEFAULT_SMOKE    0.8
+/* Unscaled: a project that has never touched this renders exactly as before */
+#define RENDER_DEFAULT_FXLIGHT  1.0
 
 MonitorProperties::MonitorProperties()
     : m_font(QFont("Arial", 12))
@@ -104,6 +107,7 @@ MonitorProperties::MonitorProperties()
     , m_renderQuality(RENDER_DEFAULT_QUALITY)
     , m_ambientLightIntensity(RENDER_DEFAULT_AMBIENT)
     , m_smokeAmount(RENDER_DEFAULT_SMOKE)
+    , m_fixtureLightIntensity(RENDER_DEFAULT_FXLIGHT)
     , m_showFPS(false)
     , m_showLabels(false)
 {
@@ -119,6 +123,7 @@ void MonitorProperties::reset()
     m_renderQuality = RENDER_DEFAULT_QUALITY;
     m_ambientLightIntensity = RENDER_DEFAULT_AMBIENT;
     m_smokeAmount = RENDER_DEFAULT_SMOKE;
+    m_fixtureLightIntensity = RENDER_DEFAULT_FXLIGHT;
     m_showFPS = false;
     m_fixtureItems.clear();
     m_lightItems.clear();
@@ -729,6 +734,8 @@ bool MonitorProperties::loadXML(QXmlStreamReader &root, const Doc *mainDocument)
                 setAmbientLightIntensity(tAttrs.value(KXMLQLCMonitorRenderAmbient).toString().toDouble());
             if (tAttrs.hasAttribute(KXMLQLCMonitorRenderSmoke))
                 setSmokeAmount(tAttrs.value(KXMLQLCMonitorRenderSmoke).toString().toDouble());
+            if (tAttrs.hasAttribute(KXMLQLCMonitorRenderFxLight))
+                setFixtureLightIntensity(tAttrs.value(KXMLQLCMonitorRenderFxLight).toString().toDouble());
             if (tAttrs.hasAttribute(KXMLQLCMonitorRenderShowFPS))
                 setShowFPS(tAttrs.value(KXMLQLCMonitorRenderShowFPS).toString().toInt() != 0);
             root.skipCurrentElement();
@@ -954,6 +961,7 @@ bool MonitorProperties::saveXML(QXmlStreamWriter *doc, const Doc *mainDocument) 
     doc->writeAttribute(KXMLQLCMonitorRenderQuality, QString::number(renderQuality()));
     doc->writeAttribute(KXMLQLCMonitorRenderAmbient, QString::number(ambientLightIntensity()));
     doc->writeAttribute(KXMLQLCMonitorRenderSmoke, QString::number(smokeAmount()));
+    doc->writeAttribute(KXMLQLCMonitorRenderFxLight, QString::number(fixtureLightIntensity()));
     doc->writeAttribute(KXMLQLCMonitorRenderShowFPS, QString::number(showFPS() ? 1 : 0));
     doc->writeEndElement();
 #endif
--- a/engine/src/monitorproperties.cpp
+++ b/engine/src/monitorproperties.cpp
@@ -50,6 +50,12 @@
 #define KXMLQLCMonitorLightEmitter  QStringLiteral("LightEmitter")
 #define KXMLQLCMonitorStageItem     QStringLiteral("StageItem")
 #define KXMLQLCMonitorMeshItem      QStringLiteral("MeshItem")
+
+#define KXMLQLCMonitorRendering         QStringLiteral("Rendering")
+#define KXMLQLCMonitorRenderQuality     QStringLiteral("Quality")
+#define KXMLQLCMonitorRenderAmbient     QStringLiteral("Ambient")
+#define KXMLQLCMonitorRenderSmoke       QStringLiteral("Smoke")
+#define KXMLQLCMonitorRenderShowFPS     QStringLiteral("ShowFPS")
 #define KXMLQLCMonitorItemName      QStringLiteral("Name")
 #define KXMLQLCMonitorItemRes       QStringLiteral("Res")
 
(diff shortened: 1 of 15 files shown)


mjcmatrix:
I will close this for now and keep it in the fork until such a time that direction has been agreed with you Massimo and due to the deprecated 3D library.

PR #2130, opened 2026-09-04
3D view, bugfix: show the selected item's values, and follow changes made under the panel

mjcmatrix:
## Description

**Summary of Changes:**

The Position, Rotation and Scale fields of the 3D view settings panel showed stale or zeroed values in two cases: when a selection of several items shrank back to one, and when an undo or redo moved the selected item. Both are fixed.

**Related Issues:**

None.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Steps to reproduce**

Selection shrinking back to one item, in a project with at least one fixture and one environment custom item:

1. Open the 3D view and its settings panel.
2. Select a fixture. The Position and Rotation fields show its values.
3. Add a custom item to the selection. The fields reset to zero, as they now hold an offset.
4. Deselect the custom item, leaving the single fixture selected.
5. The fields stay at zero instead of showing the fixture's position and rotation again. The mirror case, deselecting the fixture and leaving the custom item, behaves the same way.

Undo:

1. Select a single fixture or a single custom item.
2. Change its position or rotation in the settings panel, or drag it in the view.
3. Press Ctrl+Z. The item moves back, but the fields keep showing the undone values.

**Test Cases and Results:**

| Test | Result |
| --- | --- |
| Fixture and custom item selected, custom item deselected: fields show the fixture's values | Pass |
| Fixture and custom item selected, fixture deselected: fields show the custom item's values | Pass |
| Two items selected: fields hold a relative offset, as before | Pass |
| Undo of a fixture position change: fields follow the restored value | Pass |
| Undo of a custom item position, rotation and scale change: fields follow | Pass |
| Single fixture and single custom item edits still apply an absolute value | Pass |
| Dragging an item in the view still updates the fields | Pass |

## Additional Notes

The panel refreshed its fields from two independent handlers, one per item type, and each assumed its own type owned the values. Deselecting one of two selected items therefore ran the handler for the type that no longer had a selection, which wrote that empty selection's zeroed position and rotation over the remaining item's values. Both handlers now call a single function that picks the source from what is actually selected: the neutral base for a multiple selection, the generic item's values for a single custom item, and the fixture's values otherwise.

The undo case is separate. Tardis applies these actions through `ContextManager::setFixturePosition` / `setFixtureRotation` and `MainView3D::updateGenericItemPosition` / `Rotation` / `Scale`, and none of them emitted the change signal of the property the panel reads, so the panel was never told. Those five functions now notify their property, which is where the stored value actually changes; the emissions that already existed in their callers only covered the callers. The panel follows the signals and refreshes only when exactly one item is selected, because with a multiple selection the fields hold the relative offset entered so far and re-reading them would discard what the user typed.


mjcmatrix:
I somehow merged various other commits onto this branch. I will close this PR as I cannot fix it in-place.

I will not resubmit, I will keep this fix only on my fork because I was asked not to commit any more 3D view work, until I am told otherwise.

PR #2132, opened 2026-09-05
qmlui, bugfix: rebuild the 3D view frame graph when the render quality changes

mjcmatrix:
## Description

**Summary of Changes:**

Changing Rendering > Quality in the 3D view only took effect once the view was left and entered again. Most visibly, raising the quality away from Low left the scene with no fixture light at all. `MainView3D::setRenderQuality()` now rebuilds the frame graph, the same way adding or removing a fixture item already does.

**Related Issues:**

None.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

**Test Cases:**

Open a project with a spotlight fixture aimed at the stage, run a scene that lights it, and open the 3D view. The render quality starts at High.

1. Set Rendering > Quality to Low, staying in the 3D view.
2. Leave the 3D view and return to it.
3. Set the quality back to High, staying in the 3D view.
4. Leave the 3D view and return to it.
5. Repeat step 3 for Medium and for Ultra.

**Test Results:**

| # | Expected | Before | After |
|---|---|---|---|
| 1 | Fixtures stop lighting the scene | **Fail** - the beams in the air go, but surfaces stay lit | Pass |
| 2 | Scene dark | Pass - this is where Low finally took effect | Pass |
| 3 | Fixtures light the scene again | **Fail** - stays dark | Pass |
| 4 | Fixtures lit | Pass - this was the workaround | Pass |
| 5 | Fixtures light the scene | **Fail** - stays dark | Pass |

Tested on Linux (Qt 6, Intel iGPU, hardware OpenGL). No crash, GPU error or rendering corruption across the quality transitions.

## Additional Notes

`3DView.qml`'s `updateFrameGraph()` builds the frame graph once, and for each fixture it reads that item's `useScattering` and `useShadows` flags to decide whether to create a shadow map pass, a spotlight shading pass and a scattering pass. Those flags are bound to `View3D.renderQuality`, and at Low quality they are all false, so no spotlight pass is created for any fixture.

`setRenderQuality()` previously only stored the value and emitted the notify signal. The flags flipped, but nothing rebuilt the frame graph, so the passes stayed absent. The frame graph was otherwise only rebuilt on entering the view and on adding or removing a fixture item, which is why leaving the view and coming back appeared to fix it.

`raymarchSteps` is a plain shader uniform and did track the quality live, which is why the dial looked half-applied rather than inert: dropping to Low removed the volumetric beams immediately, but the surface lighting stayed until the frame graph was next built.

The call is guarded on `m_scene3D` so that changing the quality while the 3D view has never been opened stays a no-op, and it uses the same `updateFrameGraph(true)` entry point as `removeFixtureItem()`, which destroys the existing nodes before recreating them.

### DIFF
--- a/qmlui/mainview3d.cpp
+++ b/qmlui/mainview3d.cpp
@@ -2616,6 +2616,15 @@ void MainView3D::setRenderQuality(MainView3D::RenderQuality renderQuality)
 
     m_renderQuality = renderQuality;
     emit renderQualityChanged(m_renderQuality);
+
+    // The frame graph is built once, and it reads the flags on each fixture item
+    // that say which passes that fixture needs - flags that follow the render
+    // quality. Changing the quality therefore only flips those flags: the passes
+    // they select are added or dropped when the frame graph is built again, so do
+    // that here. Without it, raising the quality from Low leaves the scene with no
+    // spotlight pass at all until the 3D view is left and entered again.
+    if (m_scene3D)
+        QMetaObject::invokeMethod(m_scene3D, "updateFrameGraph", Q_ARG(QVariant, true));
 }
 
 QStringList MainView3D::stagesList() const


mcallegari:
While I appreciate the contributions and the interesting in the project, please be aware that:
- you are burying me with notifications and code to review. I don't really have the time to process all of that
- the Qt3d module is marked for deprecation, so all your efforts are concentrated on something that will eventually die, so I won't approve all your PRs on the 3d view. I am working on another pipeline much more evolved
- instead of throwing code at me, some of these changes could have been discussed first, either in the forums or as issues on GitHub. Maybe ask my opinion first? (especially lumen stuff)

So I kindly ask you to stop sending PRs unless they are valuable fixes not just for your own usage case. (e.g curtain stuff)
Thank you

mjcmatrix:
@mcallegari Sorry about the flood of PRs. Thank you for the timely feedback. It was not my intention to generate work for you. I became motivated to help you progress V5 rapidly. I have been on annual leave for a week and have an upcoming show, so have invested many hours lately.

I only have 2 more (big) branches pending, both intended to increase realism of the 3D view - but these require deeper design decisions so definitely require discussion and agreement with you on the best approach.

I had not discovered yet that Qt3d was deprecated and I was unaware that you were working on a new more evolved pipeline.

I prefer to be more collaborative. There are reasons why I had not been yet. Firstly I didn't want to burden you with my own objectives because the product has many users. When I began to need to consult you directly, though, when I got deeper into the product design, I wasn't sure where to go. Initially I looked around the forum, GitHub and docs, but didn't find any places where there was discussion of the roadmap or a place for contributors to chat; just a Feature Requests forum section. I started to submit PRs as I went along, as I felt all were generally useful improvements for others, this seems a direct contact route, and I figured we might be able to discuss any contentious changes and I would happily amend the approach where you object, hence providing detailed rationale.

I would like to find out more from you about the future of QLC+ and ensure my future effort is compatible with your plans. 

If you are willing and have time I would like to coordinate. Do you have a preferred forum? Shall I raise GitHub issues on some broad topics or post on the forum (in the Feature Requests section?)

Kind regards - Matt

mcallegari:
In the forums there is a place to discuss developments
https://www.qlcplus.org/forum/viewforum.php?f=12

And when you install Qt it is very clear Qt3d is going to be deprecated. It's impossible you haven't seen it

image

mjcmatrix:
I used the Qt offline installer on the command-line - to avoid having to sign up for a Qt account.

PR #2135, opened 2026-09-08
Guard optional Qt 6 FFmpeg plugin installation on macOS

liaruxxx-lgtm:
## Why

Some Qt 6 macOS installations do not include the optional FFmpeg multimedia plugin. A mandatory install rule makes the build fail even though the plugin is not required.

## What changed

Install the optional Qt 6 FFmpeg plugin only when the file exists.

## Expected behavior

macOS packaging succeeds with Qt installations that provide the plugin and also with installations that omit it.

## Validation

CMake-only change; the full upstream CI matrix should verify the packaging step.


mcallegari:
Rejected.
ffmpeg is not optional and is required for audio

PR #2137, opened 2026-09-08
Fix high-resolution trackpad zoom in the 2D view

liaruxxx-lgtm:
## Why

On high-resolution trackpads, Qt can report vertical motion through pixelDelta while angleDelta is zero. The previous condition treated that event as a zoom-out request, which made trackpad zoom unreliable.

## What changed

Use the vertical wheel delta that is actually available and ignore neutral or horizontal-only events. The change is limited to the 2D view zoom handler.

## Expected behavior

Pinch or scroll zoom follows the trackpad direction and magnitude without unexpected zoom-out steps.

## Validation

The change is platform-specific event handling and will be covered by the upstream CI matrix.

### DIFF
--- a/qmlui/qml/fixturesfunctions/2DView.qml
+++ b/qmlui/qml/fixturesfunctions/2DView.qml
@@ -329,10 +329,16 @@ Rectangle
 
                 onWheel: (wheel)=>
                 {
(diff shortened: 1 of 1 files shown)


mcallegari:
Thanks

PR #2140, opened 2026-09-08
Make Actions menu submenus fit and receive clicks reliably

liaruxxx-lgtm:
## Why

Recent files, network actions, and language choices are nested inside the main menu. On narrow windows or when a submenu extends outside the parent bounds, labels can be clipped and entries can become difficult to click.

## What changed

Move submenu content into dedicated popup components, calculate safe submenu geometry against the window, constrain long recent-file labels, and add focused component tests.

## Expected behavior

Submenus stay within the available window area, long paths remain identifiable, and entries remain clickable even when the submenu lies outside the main menu bounds.

## Validation

qmltestrunner: 10 passed, 0 failed.

### DIFF
--- a/qmlui/qml/ActionsMenu.qml
+++ b/qmlui/qml/ActionsMenu.qml
@@ -29,11 +29,52 @@ Popup
 {
     id: menuRoot
     padding: 0
+    contentWidth: requiredMenuWidth()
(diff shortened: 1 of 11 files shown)


mcallegari:
I think this is superseded by #2118

PR #2144, opened 2026-09-10
resources: add Warmdance ZQ06199 wall washer fixture

sandinak:
## Fixture Information

**Manufacturer:** Warmdance

**Model:** ZQ06199

**Mode(s) included:** 4ch, 11ch, 165ch (matches the manufacturer's console-mode menu exactly)

**Channels used:** 165 (largest mode)

## Testing

- [x] Physical data is included (Width/Height/Depth measured directly off the unit; weight from the listed retail spec, 3.1 lb — see Notes)
- [ ] Fixture has been tested using the online [fixture validator](https://www.qlcplus.org/fixture_validator.php) — **this link 404s**, same as noted in #2142/#2143; validated locally instead via `resources/fixtures/scripts/check`: 1739/1739 fixtures schema-valid, 0 errors, FixturesMap.xml schema-valid. Capability ranges on the Mode/Strobe/White-Auto-Mode channels also cross-checked programmatically for full 0-255 coverage with no gaps or overlaps.
- [x] Capabilities are labeled and ordered clearly (see Notes for one deliberate exception)

## Source(s) of Information

Manufacturer's manual for this exact model (data plate confirms Model: ZQ06199): https://manuals.plus/warmdance/zq06199-180w-strobe-rgbw-led-wall-washer-light-manual — used for the DMX channel tables (4ch/11ch/165ch). Weight (3.1 lb) from a retail listing (https://www.amazon.com/dp/B0FQJ28RN5); Width/Height/Depth measured directly off the physical unit, since the manual doesn't list them.

## Notes

The 11ch mode's channel 7 ("Mode") picks a coarse band (manual / colour-mix / auto / mixed-effect / sound), and channel 8's actual meaning depends on which band is active (20 colour combos, or 52 auto effects, or 4 mixed effects, or 4 sound modes — the manual gives four overlapping descriptions of the same 0-255 range on this channel, not four disjoint sub-ranges). Channel 10's white auto-effect similarly covers "36 effects" with no per-effect DMX boundary given. Rather than invent boundaries the source doesn't provide, both are represented as single broad capabilities that name the dependency/effect count and point back to the manual. Happy to tighten these up if someone has a more detailed source.

### DIFF
--- a/resources/fixtures/FixturesMap.xml
+++ b/resources/fixtures/FixturesMap.xml
@@ -2022,6 +2022,9 @@
     <F n="Venue-THINPAR-64" m="ThinPar 64"/>
     <F n="Venue-TriStrip3Z" m="TriStrip3Z"/>
   </M>
+  <M n="Warmdance">
+    <F n="Warmdance-ZQ06199" m="ZQ06199"/>
+  </M>
   <M n="XStatic">
     <F n="XStatic-X-240Bar-RGB" m="X-240Bar RGB"/>
   </M>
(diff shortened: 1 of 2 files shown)


mcallegari:
Rejected. No-website-brand Chinese stuff.

PR #2157, opened 2026-09-19
qmlui: fix a crash clicking off the Actions menu with a sub-menu open

mjcmatrix:
## Description

**Summary of Changes:**

### What this fixes

On current master, QLC+ 5 segfaults when you open the Actions menu, hover "Open file" so the recent projects list shows, and then click anywhere outside the menu (the toolbar, the title bar, ...). The crash is in `QQuickPopup::overlayEvent()`. It happens on any screen size, not only in compact mode.

### Why it happens

Two recent changes combine to cause it:

- #2118 (`712c5ae59`, mine) turned the Actions menu sub-menus into `Popup`s so they receive the mouse on Qt 6.11. They are anchored to the menu entries and were left at the default `z`, which puts them below the Actions menu itself (`z: visible ? 99 : 0` in `MainView.qml`).
- `6949be1f0` (compact UI) added the "File" sub-menu, `fileMenu`, as another `SubMenu`. Its visibility follows the recent projects list as well:

  ```qml
  visible: submenuItem === fileMenu || submenuItem === recentMenu
  ```

  so it is opened, empty, whenever the recent list is, on every screen size. Its `parent: menuRoot` also fails to assign (a `Popup` is not an `Item`), which is the `ActionsMenu.qml:262:9: Unable to assign ActionsMenu_QMLTYPE_… to QQuickItem` warning at startup.

On a press, `QQuickOverlayPrivate::handlePress()` takes a copy of the open popups in stacking order, topmost first, and calls `overlayEvent()` on each:

```cpp
const auto popups = stackingOrderPopups();
for (QQuickPopup *popup : popups) {
    if (popup->overlayEvent(source, event)) {
```

The Actions menu is reached first and closes, because the press is outside it. Its exit transition unparents its popup item, so the menu entries leave the scene, and every sub-menu anchored to one of them gets its window set to `nullptr`. The loop then reaches the sub-menus stacked below the menu, and `overlayEvent()` starts with:

```cpp
auto *overlay = QQuickOverlay::overlay(d->window, d->parentItem);
Q_ASSERT(overlay);
```

With no window, `overlay` is null, and it is dereferenced straight away in release builds.

A temporary log on the sub-menus showed their `window` changing to `null` before the menu's `closed` signal, while they were still in that list. Hiding the sub-menus earlier (in `onAboutToHide`, or with `Qt.callLater`) does not help: a sub-menu that is already closed stays in the copied list and loses its window the same way. What matters is that no popup inside the menu is stacked below it.

### The change

- `SubMenu.qml`: sub-menus get `z: 1000`, so they always stack above the menu that holds them. Qt then handles them before the menu closes. They are `NoAutoClose`, so they just return `false` and the loop carries on to the menu.
- `ActionsMenu.qml`: `fileMenu` only shows in compact mode, and it is parented to `menuRoot.contentItem`, which is what `parent: menuRoot` was meant to do and removes the startup warning. In compact mode the File → Open file → recent chain still nests one sub-menu popup inside another, and the `z` change is what fixes that case.

**Related Issues:**
Follow-up to #2118.

## Checklist

  - [x] Linux
  - [ ] Windows
  - [ ] macOS

## Testing

Built and run as the real application on Linux/X11 with Qt 6.11.2, with the clicks scripted.

**Test Cases:**
1. Open Actions, hover "Open file" so the recent projects list opens, click on the toolbar.
2. As 1, with the Network sub-menu open.
3. As 1, with the Language sub-menu open.
4. Compact mode (forced on for the test): Actions → File → Open file → recent list, click on the toolbar.
5. Click a recent project entry: it still loads.
6. Check the startup log for the `Unable to assign` warning.

**Test Results:**

| Build | 1 | 2 | 3 | 4 | 5 | 6 |
|---|---|---|---|---|---|---|
| master (`c28e303c1`) | crash | OK | OK | crash¹ | OK | warning |
| master + this PR | OK | OK | OK | OK | OK | no warning |
| `6949be1f0^` (#2118, before compact UI) | OK | OK | OK | n/a | OK | n/a |
| `6949be1f0` | crash | — | OK | — | — | — |

¹ Tested on the fork, which has the same ActionsMenu/SubMenu code as master.

Only tested with Qt 6.11.2, where #2118 is needed.


mcallegari:
Thanks :+1:

